PR #118, opened 2017-06-11
ceph-perf: adds Ceph baseliner

michaelsevilla:
Right now it works multiple node (tested on our in-house cluster) and on a single node. For single node, it launches one OSD, MDS, MON, and Graphite on the same node. RADOS bench sucks in this scenario (going from about 50MB/s to 30MB/s). 

TODO: 
- [ ] Run Popper check
- [ ] Fill out `validate.sh`

I had two questions:
1. Should `validate.sh` work for both single node and multiple node?
2. Do we want the results and Python notebook to be part of the template?
  - it makes the commit kinda beefy but I think the markdown in the notebook is helpful

Let me know what you want changed.


michaelsevilla:
It appears that Popper templates should not have submodules:

```bash
msevilla@pl3:~$ cd progFS-popper/
msevilla@pl3:~/progFS-popper$ popper experiment add ceph-perf nn
msevilla@pl3:~/progFS-popper$ ls experiments/nn/
ansible  docker  hosts  README.md  results-multinode  results-singlenode  run.sh  setup.sh  teardown.sh  validate.sh  vars.yml  visualize
msevilla@pl3:~/progFS-popper/experiments$ git add -A
fatal: Not a git repository: experiments/nn/ansible/roles/srl/../../../../../../.git/modules/templates/experiments/ceph-perf/ansible/roles/srl
```

ivotron:
@michaelsevilla once we merge this, `popper update` or `popper init` will clone this repo recursively (see [here](https://github.com/systemslab/popper/blob/master/popper/main.go#L38)) . So this should work after merging.

What I've been doing with some of my experiments that generate a lot of files is, as you mentioned, to just commit the notebook but not the logs/csv/text files. For this scenario, I'm working on #119 .

thanks a lot for this man, this will make life much easier for people wanting to start experimenting with Ceph

ivotron:
regarding your first question on the OP: `validate.sh` does _not_ have to work for single-node if your experiment is a multi-node one. Also, we can merge this and open a separate issue for adding a `validate.sh` for this template.

Issue #120, opened 2017-06-14
cli: experiment add should not copy submodules


michaelsevilla:
Experiment templates that use submodules are not properly added to the experiment repository's`.gitmodules`. For example, the `ceph-perf` templates uses modules but instantiating it prevents the user from committing the new experiment:

```
msevilla@pl3:~$ popper update
HEAD is now at b1db612 Bumps version to v0.3
Already up-to-date.
Using default tag: latest
latest: Pulling from ivotron/popperci-experimenter
Digest: sha256:4f8cbd7782572054e0cca6d723181b437ea91cbf00117f6c3b30588ad8369457
Status: Image is up to date for ivotron/popperci-experimenter:latest
Updated Popper repository successfully.
msevilla@pl3:~$ mkdir scratch
msevilla@pl3:~$ cd scratch/
msevilla@pl3:~/scratch$ popper init
Initialized Popper repository.
msevilla@pl3:~/scratch$ git init
Initialized empty Git repository in /home/msevilla/scratch/.git/
msevilla@pl3:~/scratch$ popper experiment add ceph-perf new-experiment
Added new-experiment to experiments/ folder.
msevilla@pl3:~/scratch$ git status
On branch master

Initial commit

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.popper.yml
	experiments/

nothing added to commit but untracked files present (use "git add" to track)
msevilla@pl3:~/scratch$ git add experiments/
fatal: Not a git repository: experiments/new-experiment/ansible/roles/srl/../../../../../../.git/modules/templates/experiments/ceph-perf/ansible/roles/srl
```

michaelsevilla:
The problem is [here](https://github.com/systemslab/popper/blob/master/popper/experiment.go#L75). Using `cp -r` means that the submodules do not get added properly to the new experiment repository.

To fix this, we can do 2 things:
- add a special metadata file that in the template that tells Popper to do `git submodule add`
- copy all files to the `.git/modules` folder

michaelsevilla:
We can't use the default `.gitmodules` and `.git/` formats because we don't want to be copying this git-specific metadata, so we will go with option 1.

ivotron:
We could check if there's some git foo that can be used here. An alternative is to use this opportunity to rethink how we deal with templates. Maybe we could just have a YAML file with a list of known popperized repos? Something like:

```yaml
repos:
- github.com/user/foo
- bitbucket.org/user/bar
```

[...]

Issue #70, opened 2017-02-28
cli: search pipelines from github/popperized


ivotron:
Right now, if users want to reuse an existing pipeline (implemented by others), they need to go the corresponding repository (e.g. one of the repos in https://github.com/popperized), download or clone a repository and copy-paste the contents of a folder into their own repository. As part of this issue, we will add a `search` command that allows users to list the available pipelines in a github organization. The naming convention of a pipeline will be `org/repo/pipeline`

ivotron:
This is going to be simpler. The global config will by default point to the `github/popperized` and the `search` dataset will come from any pipeline in any repo of that organization.

UjjwalAyyangar:
I will take a look

ivotron:
@UjjwalAyyangar @ankan17 sorry for the confusion, I hadn't realized that #144 was a dup of this. I've added more detail.

UjjwalAyyangar:
@ivotron I am almost done with the search command. Just a small doubt - 
What should we return when we find a pipeline? Like, when we find a repository, we return it's metadata along with the repository name. Say we find the pipeline, will we be returning the metadata of the repository it is in?

ivotron:
Something like:

```bash
popper search quiho
```

would result in:

```
popperized/quiho-popper
```

If there are multiple matches, then we can display a list

UjjwalAyyangar:
Got it :+1:

ankan17:
@UjjwalAyyangar Are you still working on this one??

ivotron:
closed by #315

PR #128, opened 2017-07-04
cli: add submodules using git interface

michaelsevilla:
- specialized gitmodules file tells popper to 'git submodule add'
- experiment add does not copy submodules with cp -r
- for clarity we avoid using YAML and git-specific formats

Signed-off-by: Michael Sevilla <mikesevilla3@gmail.com>

### DIFF
--- a/popper/experiment.go
+++ b/popper/experiment.go
@@ -6,6 +6,8 @@ import (
 	"log"
 	"path"
 	"strings"
+	"os"
+	"bufio"
 
 	"github.com/codeskyblue/go-sh"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


michaelsevilla:
Addresses #120. @ivotron PTAL 

Need some feedback.

popper/experiment.go
```diff
@@ -63,6 +65,64 @@ func listTemplates(templateType string) {
...
+	s := bufio.NewScanner(f)
+	s.Split(bufio.ScanLines)
+
+	// gs (gitmodule settings): nested map that maps keys of type string to maps
+	var gs map[string]map[string]string = make(map[string]map[string]string)
+
+	// g (gitmodule): the current module we are parsing
+	var g string
+
+	// parse gitmodule settings (this is a hack)
```
michaelsevilla: This parsing is hideous. @ivotron: what do you think? Can we make this cleaner?

popper/experiment.go
```diff
@@ -63,6 +65,64 @@ func listTemplates(templateType string) {
...
+			gs[g] = make(map[string]string)
+		} else {
+			// we are parsing a gitmodule attribute
+			words := strings.Split(line, "=")
+			key := strings.TrimSpace(words[0])
+			val := strings.TrimSpace(words[1])
+
+			// save attribute for current gitmodule
+			switch key {
+			case "path":
```
michaelsevilla: Potential to overwrite attribute if the user creates a malformed `.gitmodules` file for the template.

popper/experiment.go
```diff
@@ -63,6 +65,64 @@ func listTemplates(templateType string) {
...
+			case "url":
+				gs[g]["url"] = val
+			case "hash":
+				gs[g]["hash"] = val
+			}
+		}
+	}
+
+	// use git to add submodule(s)
+	for _, g := range gs {
```
michaelsevilla: Could fail in the middle, in which case we have checked out master and we aren't on the correct hash/branch.

popper/experiment.go
```diff
@@ -63,6 +65,64 @@ func listTemplates(templateType string) {
...
+		path:= folder + "/" + g["path"]
+		if err := sh.Command("git", "submodule", "add", g["url"], path).Run(); err != nil {
+			log.Fatalln(err)
+		}
+		if err := sh.Command("git", "checkout", "-b", g["hash"], g["hash"], sh.Dir(path)).Run(); err != nil {
```
michaelsevilla: Could fail in the middle, in which case we have checked out master and we aren't on the correct hash/branch.

ivotron:
@michaelsevilla I think I finally found a practical solution: popper repos shouldn't have submodules. If you need a submodule, it probably is an indication that whatever you're submoduling needs its own package (e.g. docker image, pip package, cli interface, etc..). In the case of the ceph baseliner, I feel like it's sufficiently complex that you can put it on its own docker image. what do you think?

ivotron:
subsumed by #70

PR #230, opened 2018-02-25
#221-Single-source version of codebase

### LINKED ISSUE #221: cli: single-source version of codebase
Currently, there are two sources of the version of the popper CLI, namely the [`setup.py` file](https://github.com/systemslab/popper/blob/master/cli/setup.py) and the [`cmd_version.py`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_version.py). In order to avoid having them be out of sync, we need to single-source them. More info can be found [here](https://stackoverflow.com/q/17583443), [here](https://packaging.python.org/guides/single-sourcing-package-version/) and [here](https://stackoverflow.com/q/458550)

UjjwalAyyangar:
Currently, there are two sources of the version of the popper CLI, namely the setup.py file and the cmd_version.py. In order to avoid having them be out of sync, they have been single-sourced.

Closes #221

### DIFF
--- a/cli/popper/__init__.py
+++ b/cli/popper/__init__.py
@@ -1,0 +1,1 @@
+__version__ = 'Popper 0.6-dev.'
--- a/cli/popper/__init__.py
+++ b/cli/popper/__init__.py
@@ -1,1 +1,1 @@
-__version__ = 'Popper 0.6-dev.'
+__version__ = '0.6-dev.'
--- a/cli/popper/__init__.py
+++ b/cli/popper/__init__.py
@@ -1,1 +1,1 @@
-__version__ = '0.6-dev.'
+__version__ = '0.6-dev'
--- a/cli/popper/cli.py
+++ b/cli/popper/cli.py
@@ -1,7 +1,7 @@
 import os
 import sys
 import click
-import popper
+from . import __version__ as popper_version
 
 class Context(object):
 
@@ -24,7 +24,7 @@ class Context(object):
 pass_context = click.make_pass_decorator(Context, ensure=True)
 cmd_folder = os.path.abspath(os.path.join(os.path.dirname(__file__),
                                           'commands'))
-popper_version = popper.__version__
+popper_version = popper_version
 
 
 class PopperCLI(click.MultiCommand):
--- a/cli/popper/cli.py
+++ b/cli/popper/cli.py
@@ -1,7 +1,7 @@
 import os
 import sys
 import click
-
+import popper
 
 class Context(object):
 
@@ -24,7 +24,7 @@ class Context(object):
 pass_context = click.make_pass_decorator(Context, ensure=True)
 cmd_folder = os.path.abspath(os.path.join(os.path.dirname(__file__),
                                           'commands'))
-popper_version = '0.6-dev'
+popper_version = popper.__version__
 
 
 class PopperCLI(click.MultiCommand):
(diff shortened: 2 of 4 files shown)


cli/popper/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = 'Popper 0.6-dev.'
```
ivotron: This should be just the id, e.g. `0.6-dev`.

cli/setup.py
```diff
@@ -1,8 +1,9 @@
 from setuptools import setup
+import popper
 
 setup(
     name='popper',
-    version='0.6-dev0',
+    version='popper.__version__',
```
ivotron: This is not going to work with setuptools. One way you can try is by doing the following:

```
cd cli/
python setup.py sdist
```

you should see error and warning messages. The referenced links in issue #221 give descriptions of what can be done for single-sourcing the version number, for example, alternative number 3 of the [official packaging documentation on this](https://packaging.python.org/guides/single-sourcing-package-version/) describes a way of using `execfile` to load the `__init.py__` file from `setup.py`.

cli/popper/cli.py
```diff
@@ -1,7 +1,7 @@
 import os
 import sys
 import click
-
+import popper
```
ivotron: Since this file is already in the `popper` package, importing it a package from itself it's kind of confusing. The "pythonic" way is to use [relative imports](https://docs.python.org/3/tutorial/modules.html#intra-package-references). In this case, this could work:

```python
from . import __version__ as popper_version
```

cli/popper/commands/cmd_version.py
```diff
@@ -1,11 +1,11 @@
 import click
 import popper.utils as pu
-
+import popper
```
ivotron: same as above

ivotron:
Oh, one more thing that I forgot to add. It'd be really cool if you could add a test for this, for example, a  bash script that checks that `popper version` and `python setup.py sdist` both show the same version. If you feel that this is not really something you could work on as part of this, that's cool, just let me know and I'll open another issue for this (or, if you have time, could you open such issue yourself?) Thanks!

UjjwalAyyangar:
No. I am cool working on this. I will add a test and I don't think a new issue for this would be required. Thanks for the detailed review, I will start working on it.

cli/popper/commands/cmd_version.py
```diff
@@ -1,11 +1,11 @@
...
+from .. import __version__ as popper_version
+from ..cli import pass_context
 
 
 @click.command('version', short_help='Show version of Popper and exit.')
 @pass_context
 def cli(ctx):
     """Displays version of Popper and exit."""
-    pu.info('Popper 0.6-dev.')
+    pu.info(popper_version)
```
ivotron: Minor thing: info message should be `pu.info('popper version ' + popper_version)`, this is typically what other CLI tools do

cli/setup.py
```diff
@@ -1,8 +1,13 @@
 from setuptools import setup
+import popper
+
+version={}
+with open('popper/__init__.py') as f:
+    exec(f.read(),version)
```
ivotron: space between comma and `version` argument as per PEP8

cli/popper/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.6-dev.'
```
ivotron: Remove last period, i.e. `0.6-dev`

Issue #231, opened 2018-02-26
cli: improve message for undefined commands


ivotron:
Currently, when a command that doesn't exist is passed to the CLI, the following message is displayed:

```python
$ popper status
Traceback (most recent call last):
  File "/Users/ivo/tmp/popper/cli/bin/popper", line 16, in <module>
    popper.cli.cli()
  File "/Users/ivo/Downloads/popper/cli/bin/vendor/click/click/core.py", line 731, in __call__
    return self.main(*args, **kwargs)
  File "/Users/ivo/Downloads/popper/cli/bin/vendor/click/click/core.py", line 706, in main
    rv = self.invoke(ctx)
  File "/Users/ivo/Downloads/popper/cli/bin/vendor/click/click/core.py", line 1075, in invoke
    cmd_name, cmd, args = self.resolve_command(ctx, args)
  File "/Users/ivo/Downloads/popper/cli/bin/vendor/click/click/core.py", line 1114, in resolve_command
    cmd = self.get_command(ctx, cmd_name)
  File "/Users/ivo/Downloads/popper/cli/popper/cli.py", line 47, in get_command
    raise(e)
ImportError: No module named cmd_status
```

For this issue, we will change this message to something more friendly, e.g. `Command 'status' doesn't exist. Type 'popper --help' for more`.

PR #234, opened 2018-02-27
#231 - Improved the message for undefined commands

UjjwalAyyangar:
A command `Command 'xyz' doesn't exist. Type 'popper --help' for more.`
for an undefined command 'xyz' has been added.


ivotron:
Thanks a lot!

PR #237, opened 2018-03-02
Fixes #235 Add check for .popper.yml integrity

### LINKED ISSUE #235: cli: add check for .popper.yml integrity
Add code to check that `pipelines` and `metadata` dictionaries are defined in the `.popper.yml` file. This can be added to the [`read_config`](https://github.com/systemslab/popper/blob/master/cli/popper/utils.py#L44) function, otherwise all other commands will fail. For this, we should warn the user that `.popper.yml` doesn't contain expected entries, that they should remove it and reinitialize the repository. For example:

```
ERROR: cannot read .popper.yml file. Consider deleting it and reinitializing the repo. See popper init --help for more.
```

ankan17:
The read_config() function in utils.py is modified to check if the
.popper.yml file is not empty and if the "metadata" and "pipelines"
dictionaries are defined in it and fail with a proper message if
either criterion is not met.

### DIFF
--- a/cli/popper/utils.py
+++ b/cli/popper/utils.py
@@ -54,6 +54,14 @@ def read_config():
 
     with open(config_filename, 'r') as f:
         config = yaml.load(f.read())
+        if not config:
+            fail(".popper.yml is empty. Consider deleting it and "
+                 "reinitializing the repo. See popper init --help for more.")
+        for key in ["metadata", "pipelines"]:
+            if key not in config:
+                fail(".popper.yml doesn't contain expected entries. "
+                     "Consider deleting it and reinitializing the repo. "
+                     "See popper init --help for more.")
 
     return config
 


cli/popper/utils.py
```diff
@@ -54,6 +54,14 @@ def read_config():
 
     with open(config_filename, 'r') as f:
         config = yaml.load(f.read())
+        if not config:
+            fail(".popper.yml is empty. Consider deleting it and "
+                 "reinitializing the repo. See popper init --help for more.")
+        for key in ["metadata", "pipelines"]:
```
ivotron: this could be simplified by doing (no need for a loop):

```python
if 'metadata' not in config or 'pipelines' not in config:
```

ankan17:
There are a few benefits of doing it this way:
1. Easily extendable if more checks were to be included down the road (3-4 or's will look clumsy!)
2. With 'or', even if the first check fails the second one runs but in looping we are checking two cases anyway and if the first one fails second one is not checked.

But still if you prefer the 'or' approach, I'll make the changes :)

ivotron:
ok, let's leave it like it is. `or` doesn't work like that though, if LHS is true, right one is not checked. Thanks a lot for your contribution!

ankan17:
Thanks for the clarification @ivotron I can modify the PR if you feel this not the best way to solve the problem. Just let me know :)

ivotron:
no worries, it’s OK as it is. Thanks!

PR #238, opened 2018-03-03
Fixes #228 Add support for providing string timeout values

### LINKED ISSUE #228: cli: add support for providing timeout values in string format
Currently, the values for `--timeout` flag of the [`popper run`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_run.py) command are given in seconds. As part of this issue, we will add support for human-friendly time format such as `180s`, `3m`, `3h`, and so on..

ankan17:
Support for human-friendly timeout values in string format has been added. A function has been added in utils.py which parses the timeout value which can be given as 20s or 10m etc. --timeout option now
takes integer or string as a string and it is converted to seconds accordingly.

### DIFF
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -16,10 +16,13 @@ from subprocess import check_output
 @click.argument('pipeline', required=False)
 @click.option(
     '--timeout',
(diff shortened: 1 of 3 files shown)


ivotron:
thanks a lot!

PR #240, opened 2018-03-03
Issue-236 - Initialize paper command fixed

### LINKED ISSUE #236: cli: initializing a paper makes a pipeline
According to the documentation, we should be able to create a folder for our paper:

```
$ popper init --help
Without any arguments, this command initializes a popper repository. If
an argument is given, an pipeline or paper folder is initialized. If the given
name is 'paper', then a 'paper' folder is created. Otherwise, an pipeline named
'name' is created.
[... snip ...]
```

But when I create a paper, it makes a pipeline instead:

```bash
$ popper init paper
Initialized paper pipeline.
$ ls 
pipelines
$ ls pipelines/
baseline	paper
```

I think there should be an `else if name == 'paper':` statement on [Line 59](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_init.py#L59) of `cmd_init.py`.

UjjwalAyyangar:
This fixes the problems faced while initializing a paper pipeline. 
Closes #236

### DIFF
--- a/cli/popper/commands/cmd_init.py
+++ b/cli/popper/commands/cmd_init.py
@@ -57,6 +57,7 @@ def cli(ctx, name, stages, envs, existing):
         relative_path = name
         initialize_existing_pipeline(abs_path, stages, envs)
     elif name == 'paper':
+        # create a paper pipeline
         abs_path = os.path.join(project_root, 'paper')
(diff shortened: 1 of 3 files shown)


ivotron:
thanks a lot!

PR #244, opened 2018-03-04
Fixes #225 Add --ignore-errors flag to 'popper run'

### LINKED ISSUE #225: cli: add --ignore-errors flag to 'popper run'
Currently, when the [`popper run`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_run.py) command executes multiple pipelines (i.e. when no `PIPELINE` argument is given), the command exits as soon as a pipeline fails. For this issue, we will add a boolean `--ignore-errors` flag so that, when given, all pipelines are executed regardless of whether there is a failure.

ankan17:
A --ignore-errors flag has been added to the popper run command which executes all pipelines even when there is an error in one (when no pipeline argument is provided) and shows a warning when used while providing a pipeline argument.

### DIFF
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -54,7 +54,7 @@ def cli(ctx, pipeline, timeout, skip, ignore_errors):
 
     if pipeline:
         if ignore_errors:
-            pu.info("W: ignore-errors flag is ignored when pipeline argument is provided")
+            pu.warn("ignore-errors flag is ignored when pipeline argument is provided")
         if pipeline not in pipes:
             pu.fail("Cannot find pipeline {} in .popper.yml".format(pipeline))
         status = run_pipeline(project_root, pipes[pipeline], time_out, skip)
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -29,8 +29,15 @@ from subprocess import check_output
     help='Comma-separated list of stages to skip.',
     required=False,
 )
+@click.option(
+    '--ignore-errors',
+    is_flag=True,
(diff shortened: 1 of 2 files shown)


cli/popper/commands/cmd_run.py
```diff
@@ -42,6 +49,8 @@ def cli(ctx, pipeline, timeout, skip):
         sys.exit(0)
 
     if pipeline:
+        if ignore_errors:
+            pu.info("W: ignore-errors flag is ignored when pipeline argument is provided")
```
ivotron: does the `W` stand for WARNING? If that's the case, then we could add a new utility function named `warn`

cli/popper/commands/cmd_run.py
```diff
@@ -56,7 +65,8 @@ def cli(ctx, pipeline, timeout, skip):
                 status = run_pipeline(project_root, pipes[pipe], timeout, skip)
 
                 if status == 'FAIL':
-                    break
+                    if not ignore_errors:
+                        break
```
ivotron: this could be simplified to `if status == 'FAIL' and not ignore_errors`

ankan17:
How do I write tests for this one?

cli/popper/utils.py
```diff
@@ -103,3 +103,7 @@ def fail(msg):
 
 def info(msg):
     click.echo(msg)
+
+
+def warn(msg):
+    click.echo('W: ' + msg)
```
ivotron: `WARNING:` instead of `W:`

ivotron:
Also, can you please rebase `master`?

ankan17:
Done @ivotron

ivotron:
Thanks!

PR #246, opened 2018-03-05
Fixes #227 Implement convention of always running 'teardown.sh'

### LINKED ISSUE #227: cli: implement convention of always running 'teardown.sh'
Currently, when a pipeline stage fails as part of the [`popper run`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_run.py) command, the entire pipeline is interrupted and no subsequent stage is executed. As part of this issue, we will implement the convention of executing (if defined) a `teardown.sh` stage, regardless of whether previous stages have failed.

ankan17:
- A get_filename method has been added to utils.py for getting the filename of a stage.
- When a stage fails, teardown.sh runs if present and if not skipped.

### DIFF
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -129,7 +129,7 @@ def run_pipeline(project_root, pipeline, timeout, skip):
                     stage != 'teardown' and
                     'teardown' in pipeline['stages'] and
                     'teardown' not in skipped):
-                    execute(pu.get_filename(abs_path, 'teardown'), timeout)
+                        execute(pu.get_filename(abs_path, 'teardown'), timeout)
 
                 break
 
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -98,16 +98,17 @@ def run_pipeline(project_root, pipeline, timeout, skip):
                            label="Current stage: ", item_show_func=str,
                            bar_template='[%(bar)s] %(label)s %(info)s',
                            show_percent=False) as stages:
+
         for stage in stages:
 
-            if os.path.isfile(stage):
-                stage_file = stage
-            elif os.path.isfile(stage + '.sh'):
-                stage_file = stage + '.sh'
-            else:
-                continue
+            stage_file = pu.get_filename(abs_path, stage)
 
-            if skip and stage in skip.split(','):
+            if skip:
+                skipped = skip.split(',')
+            else:
+                skipped = []
+
+            if not stage_file or stage in skipped:
                 continue
 
             ecode = execute(stage_file, timeout, stages)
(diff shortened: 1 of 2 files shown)


cli/popper/commands/cmd_run.py
```diff
@@ -97,6 +99,12 @@ def run_pipeline(project_root, pipeline, timeout, skip):
             for t in ['.err', '.out']:
                 with open('popper_logs/{}{}'.format(stage, t), 'r') as f:
                     pu.info(f.read())
+
+            if pu.get_filename(abs_path, 'teardown') and \
+               (not skip or 'teardown' not in skip.split(',')) and \
+               'teardown' not in is_executed:
+                execute(pu.get_filename(abs_path, 'teardown'), timeout)
+
```
ivotron: this could be simplified (`is_executed` list wouldn't be needed):

```python
if stage != 'teardown' and 'teardown' in pipeline['stages']:
    execute(pu.get_filename(abs_path, 'teardown'), timeout)
```

ankan17: Sorry, but I don't get how does it solve the problem

ivotron: do you want to keep track of what's has been executed with the goal of not running `teardown` when `teardown` fails? or is there another reason?

ankan17: I mean if by any chance `teardown` is ahead of a stage which fails, then `teardown` has already been run and after the stage fails we'll again run `teardown`. That is why I want to keep track of what has been executed.

UjjwalAyyangar: @ankan17 instead of keeping a whole list, maybe you could just use a flag variable to check if teardown has been executed or not. 

For example something like :- 
```
is_executed = 0 #outside the for-loop
if stage != 'teardown' and 'teardown' in pipeline['stages'] and is_executed == 0 :
    is_executed = 1
    execute(pu.get_filename(abs_path, 'teardown'), timeout)
```

ivotron: thanks @ankan17 and @UjjwalAyyangar . The idea is that `teardown`, if used, is supposed to be the last stage of a pipeline. While it is possible to have it before than the last one, that wouldn't be the way we intend users to make use of this functionality when defining pipelines.

So maybe what we could do is check on `popper init` if a user is defining a pipeline with `teardown` _not_ being the last stage. E.g.:

```bash
popper init --stages=one,teardown,three mypipe
ERROR: teardown should be the last stage. Consider renaming it or putting it at the end.
```

what do you guys think?

UjjwalAyyangar: @ivotron Yes yes, that makes a lot of sense. I will add in a check for this

ankan17:
Changes made @ivotron

cli/popper/commands/cmd_run.py
```diff
@@ -91,10 +91,8 @@ def run_pipeline(project_root, pipeline, timeout, skip):
 
     for stage in pipeline['stages']:
 
-        if os.path.isfile(stage):
-            stage_file = stage
-        elif os.path.isfile(stage + '.sh'):
-            stage_file = stage + '.sh'
+        if pu.get_filename(abs_path, stage):
+            stage_file = pu.get_filename(abs_path, stage)
```
ivotron: this can avoid invoking `get_filename` twice by doing:

```python
stage_file = pu.get_filename(abs_path, stage)
skipped = skip.split(',')

if not stage_file or stage in skipped:
    continue
```

cli/popper/commands/cmd_run.py
```diff
@@ -110,6 +108,12 @@ def run_pipeline(project_root, pipeline, timeout, skip):
             for t in ['.err', '.out']:
                 with open('popper_logs/{}{}'.format(stage_file, t), 'r') as f:
                     pu.info(f.read())
+
+            if stage != 'teardown' and 'teardown' in pipeline['stages'] and \
+               pu.get_filename(abs_path, 'teardown') and \
+               (not skip or 'teardown' not in skip.split(',')):
+                execute(pu.get_filename(abs_path, 'teardown'), timeout)
+
```
ivotron: The new utility function is already checking if the file exist, so you could do:

```python
teardown_file = pu.get_filename(abs_pat, 'teardown')

if (teardown_file and
      stage != 'teardown' and
      'teardown' in pipeline['stages'] and
      'teardown' not in skipped):
    execute(teardown_file, timeout)
```

where `skipped` would be obtained in the way mentioned on the other comment

ankan17:
Please review @ivotron

ivotron:
Sorry, this slipped through the cracks. Can you please rebase master?

ankan17:
Rebased master @ivotron

ivotron:
thanks a lot!

Issue #144, opened 2017-08-09
cli: add 'search' command


ivotron:
add `search` command to search experiments

ankan17:
So is this about creating another file called cmd_search.py? And where from are we searching really?

ivotron:
Sorry @ankan17 this is actually a duplicate issue for #70 . I'll close this and move the discussion to that issue

PR #247, opened 2018-03-05
issue-144 search command added for experiments

UjjwalAyyangar:
Added a search command for searching experiments 
as asked in #144


ivotron:
So sorry, this is actually a dup of #70 . The idea is to use the github API to get repo metadata information for a git repository or github org and allow a user to search pipelines by name or content. Issue #70 has more details

Issue #250, opened 2018-03-06
cli: check if 'teardown' is last stage for popper init --stages


UjjwalAyyangar:
Check on the popper init command to see if a user is using the --stages argument while defining
 pipeline and is keeping the teardown stage at last.

 Return a relevant error message if the convention is not being followed.

Example :- 
```
popper init --stages=setup,teardown,validate experiment1
ERROR: teardown should be the last stage. Consider renaming it or putting it at the end.
```

pcwilcox:
Should this error cause an exit, or should it just give a warning and continue?

PR #251, opened 2018-03-06
Issue-250 - teardown check on popper init.

UjjwalAyyangar:
Addresses #250

### DIFF
--- a/docs/ci/demo
+++ b/docs/ci/demo
@@ -38,6 +38,13 @@ popper run --timeout "1m 10s"
 # arbitrary stages and envs
 init_test
 popper init --stages=one,two,three --envs=host,ubuntu-xenial pipetwo
+popper init --stages=one,two,teardown sample_exp
+
+set +e
+popper init --stages=one,teardown,three sample_exp2
+test $? -ne 0
+set -e
+
 cat .popper.yml | grep '\- host'
 cat .popper.yml | grep '\- ubuntu-xenial'
 cat .popper.yml | grep '\- one'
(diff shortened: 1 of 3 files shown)


docs/ci/demo
```diff
@@ -38,6 +38,8 @@ popper run --timeout "1m 10s"
 # arbitrary stages and envs
 init_test
 popper init --stages=one,two,three --envs=host,ubuntu-xenial pipetwo
+popper init --stages=one,two,teardown sample_exp
+
```
ivotron: we can also check for when `teardown` is not the last step and check that an error was thrown.

UjjwalAyyangar: @ivotron Apologies. I was slightly confused on how to handle the stderr output.
I have added it now. Let me know if that is not the correct way to do it.

UjjwalAyyangar: Apologies. I was confused on how to handle the stderr output. 
I have added a test. Please let me know if it requires any changes.

docs/ci/demo
```diff
@@ -38,6 +38,9 @@ popper run --timeout "1m 10s"
 # arbitrary stages and envs
 init_test
 popper init --stages=one,two,three --envs=host,ubuntu-xenial pipetwo
+popper init --stages=one,two,teardown sample_exp
+popper init --stages=one,teardown,three sample_exp2 2>/dev/null || echo 'failed' || grep 'failed'
```
ivotron: this will always succeed regardless of the output of the `popper` command.

This can be an alternative:

```bash
set +e
popper init --stages=one,teardown,three sample_exp2
test $? -ne 0
set -e
```

ivotron:
thanks!

Issue #259, opened 2018-03-06
cli: add info command


ivotron:
As part of this issue we will add an `info` command that will display the information related to a pipeline. The idea is that the flow that one follows when searching for a pipeline is the following:

```bash
popper search quiho
Matching pipelines:
popperized/quiho-popper/single-node   popperized/quiho-popper/multinode
```

So multiple matches are available for a given keyword. So providing an `info` command allows users to obtain information about a pipeline:

```bash
popper info popperized/quiho-popper/single-node
```

the above will show the the github url, homepage (if specified on github), pipeline name, version (latest SHA1 for the repo) and the contents of the README for the pipeline.

UjjwalAyyangar:
Should multiple options be added for this?
for example :- 
* `Popper info  popperized/a/b --README` shows the content of the README
* `popper info popperized/a/b -- version` shows the  version
and etc

ivotron:
I've never seen that type of functionality on other tools (e.g. package managers). I think it'd fine without these options

PR #260, opened 2018-03-07
Issue 259 - Add a popper info command for pipelines

UjjwalAyyangar:
This addresses the issue #259 
 I am waiting for the popper add repository command to be finished, after that I will add the tests.

### DIFF
--- a/cli/popper/commands/cmd_add.py
+++ b/cli/popper/commands/cmd_add.py
@@ -48,9 +48,16 @@ def cli(ctx, pipeline):
 
     pu.info("Downloading pipeline {} ... ".format(pipeline_name))
     r = requests.get(download_url)
-    z = zipfile.ZipFile(BytesIO(r.content))
-    z.extractall()
-    z.close()
+    if r.status_code == 200:
+        try:
+            z = zipfile.ZipFile(BytesIO(r.content))
+            z.extractall()
+        except BadZipFile as e:
+            z.close()
+    else:
+        pu.fail("Unable to fetch the pipeline. Please check if the name" +
+                " of the pipeline is correct and the internet is connected")
+
     pu.info("Updating the configuration ... ")
     repo_config = get_config(owner, repo)
     update_config(owner, repo, pipeline_name, pipeline_path, repo_config)
--- a/cli/popper/commands/cmd_add.py
+++ b/cli/popper/commands/cmd_add.py
@@ -50,7 +50,6 @@ def cli(ctx, pipeline):
     r = requests.get(download_url)
     z = zipfile.ZipFile(BytesIO(r.content))
     z.extractall()
-    z.close()
     pu.info("Updating the configuration ... ")
     repo_config = get_config(owner, repo)
     update_config(owner, repo, pipeline_name, pipeline_path, repo_config)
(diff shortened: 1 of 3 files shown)


cli/setup.py
```diff
@@ -16,7 +16,8 @@
     include_package_data=True,
     install_requires=[
         'click',
-        'pyyaml'
+        'pyyaml',
+        'requests'
```
ivotron: Also, for external dependencies, we are also vendoring them in [`cli/bin/vendor`](https://github.com/systemslab/popper/tree/master/cli/bin/vendor). For this PR to be complete, we need to add [`requests`](https://github.com/requests/requests) as a submodule here.

UjjwalAyyangar:
I have already finished most of  the work with #70, just waiting for the add command to be finished so that I don't break anything. I will notify when those things are done. Thankyou.

ivotron:
Thanks @UjjwalAyyangar! Can you please rebase master and add tests to `docs/ci/demo`?

cli/popper/commands/cmd_info.py
```diff
@@ -0,0 +1,60 @@
+import click
+import popper.utils as pu
+import requests
+import os
+from popper.cli import pass_context
+
+try:
+    FileNotFoundError
+except NameError:
+    FileNotFoundError = IOError
```
ivotron: is this what you intended to have on this file? If so, can you please add a comment to explain what this is doing?

docs/ci/demo
```diff
@@ -249,9 +252,10 @@ set +e
 init_test
 popper search quiho
 popper search linux --skip-update
-test $? - eq 0
+test $? -eq 0
```
ivotron: when `set -e` is specified, then any non-zero exit code will cause the script to exit. So, in this case, if we remove line 251 (where `set +e` is specified) this line is not needed, as well as the following one (256)

cli/popper/commands/cmd_add.py
```diff
@@ -48,9 +48,12 @@ def cli(ctx, pipeline):
...
     pu.info("Downloading pipeline {} ... ".format(pipeline_name))
     r = requests.get(download_url)
-    z = zipfile.ZipFile(BytesIO(r.content))
-    z.extractall()
-    z.close()
+    try:
+        z = zipfile.ZipFile(BytesIO(r.content))
+        z.extractall()
+    except BadZipFile as e:
+        z.close()
```
ivotron: in the case of an exception, this will close the file and keep going. Should we instead throw an error when `BadZipFile` is raised?

Also, we might want to check the code we get in line 50 so that we are certain that the request succeeded

ivotron:
Oh, I also forgot to mention that there are some PEP8 issues as well

cli/popper/commands/cmd_add.py
UjjwalAyyangar: > in the case of an exception, this will close the file and keep going. Should we instead throw an error when BadZipFile is raised?

I am not sure if we would need to do this since it is not interfering with the downloads. What do you think?

cli/popper/commands/cmd_info.py
UjjwalAyyangar: Yes. I have been trying to write (and test) code which is compatible with both python3.x and python2.x versions.
FileNotFoundError does not exists in python2.x which makes use of "IOError" for such things.

ivotron:
awesome, thanks a lot for working on this

PR #261, opened 2018-03-07
CLI: bash auto-completion

### LINKED ISSUE #255: cli: add bash-completion
For this issue, we will add [bash-completion](https://debian-administration.org/article/316/An_introduction_to_bash_completion_part_1) functionality to popper so command names and flags can be auto-completed.

UjjwalAyyangar:
A bash auto-completion feature has been added.
closes #255

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -34,3 +34,23 @@ help:
 ```bash
 popper --help
 ```
+
+## Bash-completion
+To enable auto-completion support, you will need to activate the bash-completion.sh script
+inside the `cli/extras` folder. This can be done by copying this script to 
+`/etc/bash_completion.d/` or `/usr/local/etc/bash_completion.d/`. Before this, you need to 
+have bash-completion enabled/installed. Refer to [this guide](https://github.com/bobthecow/git-flow-completion/wiki/Install-Bash-git-completion)
+for the instructions. 
+
+When done with the above,
+go to the extras folder and copy and paste the following commands in your terminal :- 
+
+```bash
+ cp ./bash-completion.sh /etc/bash_completion.d/popper.sh
+```
+ 
+or
+
+```bash
+cp ./bash-completion.sh /usr/local/etc/bash_completion.d/popper.sh
+```
--- a/cli/extras/bash-completion.sh
+++ b/cli/extras/bash-completion.sh
@@ -1,0 +1,8 @@
+_popper_completion() {
+    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
+                   COMP_CWORD=$COMP_CWORD \
+                   _POPPER_COMPLETE=complete $1 ) )
+    return 0
+}
+
+complete -F _popper_completion -o default popper;
--- a/cli/popper/cli.py
+++ b/cli/popper/cli.py
@@ -4,6 +4,7 @@ import click
 from . import __version__ as popper_version
 from .exceptions import UsageError
 
+
 class Context(object):
 
     def __init__(self):
@@ -45,9 +46,9 @@ class PopperCLI(click.MultiCommand):
             mod = __import__('popper.commands.cmd_' + name,
                              None, None, ['cli'])
         except ImportError as e:
-             raise UsageError(
-                   "Command '"+name+"' doesn't exist. " + 
-                   "Type 'popper --help' for more."
+            raise UsageError(
+                "Command '" + name + "' doesn't exist. " +
+                "Type 'popper --help' for more."
             )
         return mod.cli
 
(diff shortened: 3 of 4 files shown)


UjjwalAyyangar:
I have added the instructions, will add the tests.

cli/README.md
```diff
@@ -34,3 +34,9 @@ help:
 ```bash
 popper --help
 ```
+
+To enable auto-completion support, you will need to activate the popper-complete.sh script
+inside the popper folder. This can be done by adding this into your bashrc :-  
+```bash
+ . /path/to/popper-complete.sh 
+```
```
ivotron: While this works, bash-completion is usually enabled by copying the script to `/etc/bash_completion.d/` or `/usr/local/etc/bash_completion.d/`. In our case, `/usr/local/bash_completion.d/` should work ok. Also the script has to be renamed `popper` when placed in these folders.

cli/README.md
```diff
@@ -34,3 +34,9 @@ help:
 ```bash
 popper --help
 ```
+
```
ivotron: We can add a `## bash-completion` header, before the below paragraph

ivotron:
@UjjwalAyyangar we can add support for installing the bash-completion via pip, e.g. see [here](https://stackoverflow.com/q/9944320/389188). Maybe we could open a new issue for that?

cli/popper/popper-complete.sh
```diff
@@ -0,0 +1,8 @@
+_popper_completion() {
+    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
+                   COMP_CWORD=$COMP_CWORD \
+                   _POPPER_COMPLETE=complete $1 ) )
+    return 0
+}
+
+complete -F _popper_completion -o default popper;
```
ivotron: @UjjwalAyyangar can we place this file in `cli/extras/bash-completion.sh` instead?

UjjwalAyyangar:
@ivotron I was working on installing it via pip only and just realized that I was making the mistake of using pip install in the editable mode. Thanks a lot!

cli/README.md
UjjwalAyyangar: Yes. I will make the changes. The same thing was specified on  [this answer](https://stackoverflow.com/a/18898614).

UjjwalAyyangar:
@ivotron - All the requested changes have been made. 
* Manual instructions for auto-complete added
* Support for installing auto-complete via pip added.
* Completion script added moved to a /extras folder

cli/extras/popper.sh
```diff
@@ -0,0 +1,8 @@
+_popper_completion() {
+    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
+                   COMP_CWORD=$COMP_CWORD \
+                   _POPPER_COMPLETE=complete $1 ) )
+    return 0
+}
+
+complete -F _popper_completion -o default popper;
```
ivotron: I think it would be better if this script is renamed to `bash-completion.sh`, to avoid getting it confused with the other `bin/popper` bash script, and also to make it explicitly that this script is implementing bash completion.

cli/README.md
```diff
@@ -34,3 +34,19 @@ help:
...
+## Bash-completion
+To enable auto-completion support, you will need to activate the popper.sh script
+inside the `cli/extras` folder. This can be done by copying this script to 
+`/etc/bash_completion.d/` or `/usr/local/etc/bash_completion.d/`.To do this, 
+go to the extras folder and copy and paste the following commands in your terminal :-
```
ivotron: bash-completion is not available by default on bash, so we should maybe explicitly mention that it has to be installed/enabled separately. [this](https://github.com/bobthecow/git-flow-completion/wiki/Install-Bash-git-completion) is a good guide that we could link to, since it has instructions for many distinct OSs.

ivotron:
I don't think we need to add tests for this, since that is already tested by click.

cli/README.md
UjjwalAyyangar: Yes, I will a proper guide to this.

UjjwalAyyangar:
@ivotron - I have made a separate markdown file for bash completion instructions. I was thinking of opening an issue titled "Add instructions for enabling custom bash auto-completion", since right now I don't have the tools to test the script in different Operating Systems. It can be labelled as a good-first-issue and people can keep adding the commands for their Operating System.

ivotron:
What you had was OK, no need to write instructions ourselves. We just need to explicitly say that bash-completion needs to be enabled/installed, and we can add **a link** to [this guide](https://github.com/bobthecow/git-flow-completion/wiki/Install-Bash-git-completion) (i.e. no need to write the guide ourselves).

UjjwalAyyangar:
@ivotron Yes I understood that. But I got misled by [this comment](https://superuser.com/questions/289539/custom-bash-tab-completion#comment303165_289604) and formed the idea of making a self-made guide for the users. 

I have made the requested changes. Everything should be in place now.

ivotron:
Cool, thanks a lot for working on this!

PR #263, opened 2018-03-09
Issue-254 : infer-stages flag added to pipeline init

### LINKED ISSUE #254: cli: add '--infer-stages' flag to init
An existing pipeline is added to popper (from an existing folder) by passing the `--existing` flag. For this issue, we will add an `--infer-stages` flag that infers the name of the stages from any existing bash script on the folder. For example:

```bash
cd myproject/
ls ci/
a.sh  b.sh  c.sh
```

If we do `popper init --existing --infer-stages ci/` a pipeline will be created with stages `a, b, c`. The order comes from sorting lexicographically

UjjwalAyyangar:
* `--infer-stages` flag added for the `popper init --existing` command
* tests for the new feature added
Closes #254

### DIFF
--- a/cli/popper/commands/cmd_init.py
+++ b/cli/popper/commands/cmd_init.py
@@ -1,5 +1,6 @@
 import click
 import os
+import glob
 import popper.utils as pu
 from popper.exceptions import BadArgumentUsage
 from popper.cli import pass_context
@@ -26,8 +27,15 @@ from os.path import isfile, isdir, basename
           'creating an entry in .popper.yml for this existing folder.'),
     is_flag=True
 )
+@click.option(
+    '--infer-stages',
+    help=('Infers the name of the stages from the existing bash scripts'
+          'on the folder while creating a new pipeline from the existing'
+          'folder.'),
+    is_flag=True
+)
 @pass_context
-def cli(ctx, name, stages, envs, existing):
+def cli(ctx, name, stages, envs, existing, infer_stages):
     """Initializes a repository or a pipeline. Without an argument, this
     command initializes a popper repository. If an argument is given, a
     pipeline or paper folder is initialized. If the given name is 'paper',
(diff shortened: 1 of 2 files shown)


ivotron:
@pep8speaks resume

docs/ci/demo
```diff
@@ -54,6 +54,13 @@ echo '' > ci/one.sh
 echo '' > ci/two
 echo '' > ci/three
 popper init --existing --stages=one,two,three ci
+
+mkdir ci2
+echo '' > ci/a.sh
+echo '' > ci/b.txt
+echo '' > ci/c.sh
+popper init --existing --infer-stages ci2
+
```
ivotron: Can we add a check that tests that only `a` and `c` are the stages?

cli/popper/commands/cmd_init.py
```diff
@@ -55,7 +63,10 @@ def cli(ctx, name, stages, envs, existing):
         # existing pipeline
         abs_path = os.path.join(project_root, name)
         relative_path = name
-        initialize_existing_pipeline(abs_path, stages, envs)
+        if infer_stages:
+            stages = ",".join(map(lambda x: x[:-3], sorted(glob.glob1(abs_path,'*.sh'))))
```
ivotron: this is not PEP8 compliant

docs/ci/demo
ivotron: This is the only missing piece in order to merge this PR

UjjwalAyyangar: Yes yes, I am working on it. I'll notify when completed.

ivotron: no, what I mean is that, after `popper init --existing --infer-stages ci2` runs, the entry in the `.popper.yml` file for pipeline `c2` will only contain two stages (`a` and `c`). So you can `cat` the file and grep for those, and also `cat` to see that `b` **is not** part of the pipeline.

ivotron:
thanks a lot!

Issue #266, opened 2018-03-10
doc: update documentation on `.popper.yml`


ivotron:
with the move to python, the section documenting the `.popper.yml` file on [our docs](https://github.com/systemslab/popper/blob/master/docs/ci/popperyml.md) is now outdated. We need to update it so it reflects the new structure: `metadata` and `pipelines` dictionaries, and a `popperized` list of repos to use to look for new pipelines (see #70).

PR #268, opened 2018-03-11
Doc: added info on metadata, pipelines and popperized

UjjwalAyyangar:
Addresses #266

### DIFF
--- a/docs/ci/popperyml.md
+++ b/docs/ci/popperyml.md
@@ -30,6 +30,32 @@ envs:
   arbitrary_stages: centos-7.2
   paper: host
 
+metadata: 
+  authors: Dennis Ritchie
+  book: The C Programming Language
+
+pipelines:
+  paper:
+    envs:
+    - host
+    path: paper
+    stages:
+    - build
+  sample_pipeline:
+    envs:
+    - host
+    path: pipelines/sample_pipeline
+    stages:
+    - setup
+    - run
+    - post-run
+    - validate
+    - teardown
+
+popperized:
+  - github/popperized
+  - github/ivotron/quiho-popper
+
 stages:
   myexp: [setup.sh, run.sh, post-run.sh, validate.sh, teardown.sh]
   myanalysis: [setup.sh, run.sh, validate.sh]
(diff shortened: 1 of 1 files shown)


docs/ci/popperyml.md
```diff
@@ -67,15 +93,44 @@ popper init mypipe --env=alpine-3.4
...
 ## Stages
 
 The `stages` YAML entry specifies the sequence of stages that are 
 executed by the `popper check` command. By default, the `popper init` 
 command generates scaffold scripts for `setup.sh`, `run.sh`, 
-`post-run.sh`, `validate.sh`, `teardown.sh`. If any of those are not 
-present when the pipeline is executed using `popper check`, they are 
-just skipped (without throwing an error). At least one stage needs to 
-be executed, otherwise `popper check` throws an error.
+`post-run.sh`, `validate.sh`, `teardown.sh`.
```
ivotron: This is still the case:

> If any of those are not present when the pipeline is executed using `popper check`, they are just skipped (without throwing an error).

but should read `popper run` instead of `popper check`.

UjjwalAyyangar: @ivotron Changes made

ivotron: thanks. Sorry for the confusion. Besides changing `check` for `run`, I also meant that this phrase:

> If any of those are not present when the pipeline is executed using `popper check`, they are just skipped (without throwing an error).

still applies and should be part of the document.

docs/ci/popperyml.md
```diff
@@ -67,15 +93,44 @@ popper init mypipe --env=alpine-3.4
 The above specifies that the pipeline named `mypipe` will be executed 
 inside a docker container using the `alpine-3.4` popper check image.
 
+## Metadata
+
+The `metadata` YAML entry specifies the set of data that gives information
+about the user's project. 
+
+It can be added using the `popper metadata --add command`
```
ivotron: this fragment should be part of the previous sentence, and followed by "For example" (all in one single paragraph):

> The `metadata` YAML entry specifies the set of data that gives information about the user's project.  It can be added using the `popper metadata --add command`. For example:

UjjwalAyyangar:
@ivotron Sorry for the trouble, I hope it is fine now.

ivotron:
thanks!

Issue #271, opened 2018-03-13
cli: test python 3.x support


ankan17:
Currently, popper is tested on 2.7. For this issue, we need to update the travis tests so that we also test on python3. This can be done by creating a new container based on python three (we currently just build a [2.7 one.](https://github.com/systemslab/popper/blob/master/cli/Dockerfile))

vdevgan:
It does support python 3 versions. What issues did you face?

ankan17:
Requires thorough testing @vdevgan

ankan17:
Can this be closed now that we are creating virtualenv with python3? @ivotron

ivotron:
yes, thanks!

PR #272, opened 2018-03-13
Check python version to load pyyaml appropriately

ankan17:
pyyaml has two separate libs for python2 and python3 and we need
to specifically load the particular lib for popper to work
correctly with both python2 and python3.

First step towards #271.


ivotron:
thanks!

PR #273, opened 2018-03-13
Change color of warning from orange to magenta

ankan17:
An error had crept up in PR #244 because orange color used for warnings is no ANSI color and hence the warn utility method won't function correctly. Changing 'orange' to 'magenta' does the job.

### DIFF
--- a/cli/popper/utils.py
+++ b/cli/popper/utils.py
@@ -118,7 +118,7 @@ def fail(msg):
 
 
 def warn(msg):
-    click.secho('WARNING: ' + msg, fg='orange', bold=True)
+    click.secho('WARNING: ' + msg, fg='magenta', bold=True)
 
 
 def info(msg, **styles):


UjjwalAyyangar:
@ankan17 Good spot! Maybe change it to "yellow" instead of "magenta" ?
I did some research on this and found out that the usual color code for warnings is something in between 
yellow and orange. Also, according to the [Click Documentation](http://click.pocoo.org/5/api/#click.secho), sometimes "yellow" can behave as an "orange" i.e. it will be consistent with our ideas both ways.

cli/popper/utils.py
```diff
@@ -118,7 +118,7 @@ def fail(msg):
 
 
 def warn(msg):
-    click.secho('WARNING: ' + msg, fg='orange', bold=True)
+    click.secho('WARNING: ' + msg, fg='magenta', bold=True)
```
UjjwalAyyangar: @ankan17 Good spot! Maybe change it to "yellow" instead of "magenta" ?
I did some research on this and found out that the usual color code for warnings is something in between yellow and orange. Also, according to the [Click Documentation](http://click.pocoo.org/5/api/#click.secho), sometimes "yellow" can behave as an "orange" i.e. it will be consistent with our ideas both ways.

ankan17:
yellow is being used at other places so I went with magenta.. magenta looks kinda cool too :stuck_out_tongue:

UjjwalAyyangar:
Ok  :stuck_out_tongue: 
Sure, whatever feels best to you. :+1: 
(I will keep it yellow for me though, magenta is not visible on my terminal)

ivotron:
thanks!

Issue #222, opened 2018-02-24
cli: add --list flag to 'popper env'


ivotron:
The [`popper env`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_env.py) command allows assigning multiple environments to a pipeline. These environments are used to execute a pipeline when `popper run` is issued (one execution for each environment). For this issue, a new `--list` will be added to the `env` command so that it shows the available environments, which will be retrieved by querying our [`DockerHub`](https://hub.docker.com/r/falsifiable/poppercheck/) repository. Each different `tag` of the `poppercheck` image is a new environment. Currently, the [available tags are](https://hub.docker.com/r/falsifiable/poppercheck/tags/) `debian-9`, `centos-7.2` and `alpine-3.4` tags, so executing the following:

```bash
popper env --list
```

should show the following output:

```python
['debian-9', 'centos-7.2', 'alpine-3.4']
```

shamedian:
What about `stretch`, which is [also shown on Docker Hub](https://hub.docker.com/r/falsifiable/poppercheck/tags/)?

ivotron:
I thought I had deleted the `stretch` tag (should have been `debian-stretch` or `debian-9`). But, in general, any tag on Dockerhub should be shown as the output.

shamedian:
Sounds good, thanks for the clarification!

PR #274, opened 2018-03-14
 add --ls flag to popper env command

shamedian:
Added a new `--ls` flag to `popper env` that queries and shows available tags on the Docker Hub repository.

This closes systemslab/popper#222.


cli/popper/commands/cmd_env.py
```diff
@@ -48,4 +55,13 @@ def cli(ctx, pipeline, add, rm):
...
             config['pipelines'][pipeline]['envs'].remove(e)
 
+    if ls:
+        web_results = urllib2.urlopen("https://hub.docker.com/v2/repositories/falsifiable/poppercheck/tags").read()
+        results = json.loads(web_results)['results']
+        environments = []
+        for result in results:
+            environments.append(result['name'])
+        list_of_environments = ", ".join(environments)
+        print list_of_environments
```
ivotron: there's a [utility function](https://github.com/systemslab/popper/blob/master/cli/popper/utils.py#L129) to pretty-print a YAML list, maybe we could use that by creating a YAML-formatted string and passing it to this function?

cli/popper/commands/cmd_env.py
```diff
@@ -1,12 +1,14 @@
 import click
+import json
+import urllib2
```
ivotron: We need to add these new dependencies to the `setup.py` file, as well as [vendor them](https://github.com/systemslab/popper/blob/master/cli/bin/popper) (add the corresponding projects as submodules and load their librarly paths to python's syspath).

PR #280, opened 2018-03-22
Fixes #278 Fix cmd_run.py to gracefully exit when a stage fails

### LINKED ISSUE #278: cli: error when executing pipeline when a certain stage fails
If a certain stage fails and teardown is run, the following error is encountered

```bash
TypeError: execute() missing 1 required positional argument: 'bar'
```

ankan17:
Fixed the error that occurred due to no progress bar object passed to the execute method when teardown stage was run after any stage failed which led popper to crash. Now after any stage fails,
popper will gracefully exit out after running teardown, if present.


ivotron:
thanks!

PR #281, opened 2018-03-22
Suggest command names

### LINKED ISSUE #253: cli: suggest command names
Add the ability of providing suggestions when the name of a command does not match to any of the available ones. Example from what Git does:

```bash
$ git adr
git: 'adr' is not a git command. See 'git --help'.

The most similar command is
    add
```

we will implement something similar.

UjjwalAyyangar:
Closes #253

### DIFF
--- a/cli/popper/cli.py
+++ b/cli/popper/cli.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import click
+import difflib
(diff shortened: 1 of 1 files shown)


ivotron:
cool! thanks a lot for working on this

Issue #282, opened 2018-03-26
Replace PyYaml with Ruamel.Yaml


UjjwalAyyangar:
The popper CLI tool uses pyyaml for parsing YAML. However, pyyaml only supports the old 1.1 YAML format. Since the work on pyyaml module has been abandoned we can use [ruamel.yaml](https://pypi.python.org/pypi/ruamel.yaml) module instead, which follows the new 1.2 YAML format. 

More information about the differences between the two modules can be found [here].(http://yaml.readthedocs.io/en/latest/pyyaml.html#)

For this issue, we need to replace pyyaml by ruamel.yaml

ankan17:
Think this can be closed now?

PR #283, opened 2018-03-26
Replaced pyyaml with ruamel.yaml

UjjwalAyyangar:
* Created a git [mirror](https://github.com/UjjwalAyyangar/ruamel-yaml) for ruamel.yaml.
* Added ruamel.yaml as a submodule.
* Added support for installing ruamel.yaml from pip.

Addresses issue #282

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
-[submodule "cli/bin/vendor/pyyaml"]
-	path = cli/bin/vendor/pyyaml
-	url = https://github.com/yaml/pyyaml
 [submodule "cli/bin/vendor/click"]
 	path = cli/bin/vendor/click
 	url = https://github.com/pallets/click/
 [submodule "cli/bin/vendor/requests"]
 	path = cli/bin/vendor/requests
 	url = https://github.com/requests/requests/
+[submodule "cli/bin/vendor/ruamel-yaml"]
+	path = cli/bin/vendor/ruamel-yaml
+	url = https://github.com/UjjwalAyyangar/ruamel-yaml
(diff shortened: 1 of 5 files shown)


UjjwalAyyangar:
Encountered an installation issue with the build. I will fix it and notify when completed.

cli/setup.py
```diff
@@ -19,7 +19,8 @@
     include_package_data=True,
     install_requires=[
         'click',
-        'pyyaml'
+        'pyyaml',
+        'ruamel.yaml<0.15'
```
ivotron: do we need to strictly use a version < 0.15? Also, I was under the impression that ruamel replaces pyyaml, why do we need to leave the two?

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "cli/bin/vendor/click"]
 	path = cli/bin/vendor/click
 	url = https://github.com/pallets/click/
+[submodule "cli/bin/vendor/ruamel-yaml"]
+	path = cli/bin/vendor/ruamel-yaml
+	url = https://github.com/UjjwalAyyangar/ruamel-yaml
```
ivotron: we would also need to remove `pyyaml`

cli/Dockerfile
```diff
@@ -1,9 +1,12 @@
 FROM python:2.7.14-alpine3.7
 
-RUN apk add --no-cache bash curl git sed grep bc coreutils
+RUN apk add --no-cache bash curl git sed grep bc coreutils && \
+    apk add --update g++ && \
+    rm /var/cache/apk/*
```
ivotron: can you please add `--no-cache` to the new `apk` line?

UjjwalAyyangar:
@ivotron Completed :+1:

ivotron:
Thanks a lot!

PR #296, opened 2018-03-29
Fixes #229 Add a stages subcommand to view or update stages of a pipeline

### LINKED ISSUE #229: cli: add a 'stages' command
Similar to the [`popper env`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_env.py) command, for this issue we will add a new `popper stages` command that displays the stages defined in the `.popper.yml` file. This will also require to add a `--set` flag to re-define the stages of an existing pipeline.

ankan17:
The stages sub-command outputs the stages of a pipeline, provided as an argument without the --set option. With the --set option, the comma-separated list of arguments is added to the existing stages if --replace is not provided, replaced otherwise.


ankan17:
I assume we'd require tests for this functionality?

cli/popper/commands/cmd_stages.py
```diff
@@ -0,0 +1,43 @@
...
+
+from popper.cli import pass_context
+
+@click.command(
+    'stages', short_help='See and manipulate stages of a pipeline.'
+)
+@click.argument('pipeline', required=True)
+@click.option(
+    '--set',
+    help='Provide comma separated list of stages to set the stages of a '
```
ivotron: we can simplify and leave it at `Comma separated list of stages in the pipeline`.

cli/popper/commands/cmd_stages.py
```diff
@@ -0,0 +1,43 @@
...
+         'pipeline. Without the --replace flag, the stages are added alongwith '
+         'the existing stages, otherwise they are replaced.',
+    required=False,
+)
+@click.option(
+    '--replace',
+    is_flag=True,
+    help='Use alongwith the --set option to overwrite the curent stages.',
+    required=False,
+)
```
ivotron: I'd vote for removing this flag. I think the most frequent case is to overwrite the stages

ankan17:
Done @ivotron

Issue #294, opened 2018-03-29
cli : popper add subcommand does not updates the .popper.yml file


UjjwalAyyangar:

UjjwalAyyangar:
Working on this

ankan17:
The new pipeline is added to .popper.yml. Where did you face the issue?

PR #300, opened 2018-03-30
Fix add subcommand

UjjwalAyyangar:
Addresses #294 

Changes made :- 
* Updates the .popper.yml file by adding org/repo/pipeline_name to the `popperized` entry
* Gets the stages of the downloaded pipeline and adds it to the `.popper.yml` file.
* Gets the env of the downloaded pipeline and adds it to the `.popper.yml` file.
* Added a progress bar for downloading files to avoid cluttering the terminal window.
* Creates a `'pipelines` folder if it not already present.


UjjwalAyyangar:
Need to make some more changes. Will notify when done.

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: bash
 services: docker
+env:
+    PATH=$PATH:$PWD/cli/bin && docs/ci/demo
```
ivotron: this change is causing the tests not to run. It was OK as it was before.

UjjwalAyyangar: @ivotron Just a small doubt. 

 Earlier, it was ending with exit error 2. :- 
```The command "PATH=$PATH:$PWD/cli/bin && docs/ci/demo" exited with 2.```
which signifies that misuse of shell builtins. 

Aren't we suppose to add PATH as an environment variable in travis?

ivotron: This is the error https://travis-ci.org/systemslab/popper/builds/360053590#L386 . The `PATH` variable is only needed for the `docs/ci/demo` script, not for subsequent lines in the travis file, so there's no need to put it on the `env` section of travis.

> which signifies that misuse of shell builtins.

not sure how you get to this conclusion, but exit codes distinct than 0 means error. For non-zero exit codes, their semantics depend on the app that is throwing them. In this case, this code 2 is from from click, and being thrown [here](https://github.com/systemslab/popper/blob/master/cli/popper/cli.py#L59). This is actually swallowing the actual error message, which should be an issue with python not finding ruamel.

UjjwalAyyangar: > not sure how you get to this conclusion 
@ivotron It was mentioned [here](http://tldp.org/LDP/abs/html/exitcodes.html) .

Thanks for the detailed explanation. I will look into it.

ivotron: I see. Nobody follows that convention though 😄

UjjwalAyyangar: Ok. I realized the mistake I was making with the PATH :joy: .
And, I replaced the ruamel submodule with pyyaml. Getting error from the requests submodule now. 
Will try to fix it :+1:

UjjwalAyyangar:
@ivotron All done :+1:

Issue #305, opened 2018-04-13
Fix get_project_root() function


UjjwalAyyangar:
Due to the current get_project_root() function in `cli/popper/utils.py` , all the popper
subcommands fail when they are executed from folders other than the project root 
folder. 

For example, if I run `popper add popperized/quiho-popper/single-node` from the
pipelines folder itself, I will get a :- 
 `ERROR: Unable to find the root of your project. Initialize repository first.` error.

It would be really great if the users could use commands like `popper metadata` or 
`popper run` from any folder inside their repository.

PR #306, opened 2018-04-13
fixed: get_project_root() utility function

UjjwalAyyangar:
Fixes the bug in get_project_root() function as explained in #305


UjjwalAyyangar:
Came across some unexpected errors with this one. Will fix and notify when done.

ivotron:
cool! thanks a lot

Issue #198, opened 2017-12-14
cli: add 'popper archive zenodo'


ivotron:
This will take a github repo and archive it on [Zenodo](https://zenodo.org/) using its [API](developers.zenodo.org/). More info can be found [here](https://gist.github.com/arfon/9633189)

PR #309, opened 2018-05-13
[WIP for #198] Add archiving support for zenodo

ankan17:


ivotron:
awesome, this looks great! I was thinking on what would be a good strategy for testing this. Maybe creating a fictitious user on zenodo and hard-code the token on the test scripts. What do you think?

cli/popper/commands/cmd_archive.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Upload the file
+        r = requests.post(service_url, data=data, files=files, params=params)
+
+        response = {'status_code': r.status_code}
+        if r.status_code == 201:
+            file_id = r.json()['id']
+            response['message'] = (
+                "Snapshot has been successfully uploaded. Your deposition id"
+                " is {} and the file id is {}.".format(deposition_id, file_id)
```
ivotron: Do you know if the deposition id is the same as the DOI? So when you go to zenodo, you see something like this:

screen shot 2018-05-15 at 06 36 05

I highlighted in red the two things that we are most interested in. In particular the one with the little red square. Do you know if the deposition id is the same as that id in the little red square? I think that's our ultimate goal, to get that URL since that's the persistent link that will live there forever

ivotron: based on an offline discussion, we came up with a plan to deal with the above in a separate PR

PR #318, opened 2018-05-23
Decreasing API requests for popper add

UjjwalAyyangar:
Reduces the number of GitHub API requests required to download a pipeline using the popper add command.


ivotron:
thanks!

PR #317, opened 2018-05-22
Fixes #316 Handle metadata for archiving feature

ankan17:
Required metadata keys are added at the time of the popperizied repository creation and read and parsed at the time of publishing the snapshot.

PR #323, opened 2018-06-02
Issue 190

### LINKED ISSUE #190: cli: generate .dot graph for a popper pipeline
Main goal is to be "as simple as possible, but not simpler".  User can add comments of the form `# [wf] ....` to bash scripts that correspond to popper pipeline stages. We'll implement a `popper graph` command that will scan bash scripts to generate a .dot file representing the call graph, following this convention:

  * every bash script (pipeline stage) is a node in the graph. Every node is represented by rectangles, instead of circular shapes. These are called "top-level" nodes.

  * top-level nodes are connected in the order of their execution (e.g. `setup.sh -> run.sh`, `run.sh->post-run.sh`, etc.). These stage names will be arbitrary once #184 is completed.

  * every `[wf]` comment in a bash script adds text to the node, e.g. given a `setup.sh` script:

    ```
    #!/bin/bash
    # [wf] allocates resources for experiment
    ```

     results in having the`setup.sh` node labeled with `allocates resources for experiment`.

  * the first occurrence of a `[wf]` comment adds the comment to the top-level node, while any subsequent occurrence creates a child of the top-level node.

  * if a `[wf]` comment is followed by an invocation to another script/binary, a child node is created with the name of the script/binary as a `prefix` to the comment. We don't recursively process  scripts/binaries.

  * if a `[wf]` comment is inside an `if` statement, a new `condition` child is created, which becomes the parent of any `[wf]` nodes contained in the `if` statement (with `else` or `elsif` statements as siblings of the `if` node).

  * if a `[wf]` precedes a `for` or `while` loop, the `loop:` prefix is added to the comment.

ivotron:
fixes #190

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,6 +19,6 @@
 [submodule "cli/bin/vendor/urllib3"]
 	path = cli/bin/vendor/urllib3
 	url = https://github.com/urllib3/urllib3
-[submodule "cli/bin/vendor/cryptography"]
-	path = cli/bin/vendor/cryptography
-	url = https://github.com/pyca/cryptography
+[submodule "cli/bin/vendor/lark-parser"]
+	path = cli/bin/vendor/lark-parser
+	url = https://github.com/lark-parser/lark
--- a/cli/popper/cli.py
+++ b/cli/popper/cli.py
@@ -58,7 +58,7 @@ class PopperCLI(click.MultiCommand):
             raise UsageError(
                 "Command '" + name + "' doesn't exist. " +
                 "\nType 'popper --help' for more."
-                + message
+                + message + '\n' + str(e)
             )
         return mod.cli
 
--- a/cli/popper/commands/cmd_add.py
+++ b/cli/popper/commands/cmd_add.py
@@ -54,8 +54,8 @@ def cli(ctx, pipeline, folder):
     if r.status_code != 200:
         pu.fail("Unable to fetch the pipeline. Please check if the name" +
                 " of the pipeline is correct and the internet is connected")
-    z = zipfile.ZipFile(BytesIO(r.content))
-    z.extractall()
+    with zipfile.ZipFile(BytesIO(r.content)) as z:
+        z.extractall()
 
     os.rename('{}-master/pipelines/{}'.format(repo, pipe_name),
               os.path.join(folder, pipe_name))
--- a/cli/popper/commands/cmd_add.py
+++ b/cli/popper/commands/cmd_add.py
@@ -6,7 +6,6 @@ import requests
 import popper.utils as pu
 import yaml
 import zipfile
-from zipfile import BadZipfile
 from io import BytesIO
 from popper.cli import pass_context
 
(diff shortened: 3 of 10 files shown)


ivotron:
This also ended up including some changes to the `popper add` command

cli/popper/cli.py
```diff
@@ -58,7 +58,7 @@ def get_command(self, ctx, name):
             raise UsageError(
                 "Command '" + name + "' doesn't exist. " +
                 "\nType 'popper --help' for more."
-                + message
+                + message + '\n' + str(e)
```
ankan17: The str(e) is not required imo after the error has been determined

cli/setup.py
```diff
@@ -20,8 +20,8 @@
     install_requires=[
         'click',
         'requests',
-        'pyyaml',
-        'cryptography'
+        'lark-parser',
+        'pyyaml'
```
UjjwalAyyangar: The cryptography module is not present by default in the python 3.x versions. So I think we should keep it here, to keep our codebase compatible with both python 2.x and 3.x

.gitmodules
```diff
@@ -19,6 +19,6 @@
 [submodule "cli/bin/vendor/urllib3"]
 	path = cli/bin/vendor/urllib3
 	url = https://github.com/urllib3/urllib3
-[submodule "cli/bin/vendor/cryptography"]
-	path = cli/bin/vendor/cryptography
-	url = https://github.com/pyca/cryptography
+[submodule "cli/bin/vendor/lark-parser"]
```
UjjwalAyyangar: Need to keep the cryptography module as it is not present by default in python3.x, so that our codebase is compatible with both the versions.

docs/ci/demo
```diff
@@ -174,42 +174,12 @@ popper env mypipeone | grep 'host'
...
-test -f pipelines/single-node/vars.yml
+popper add popperized/lesson/co2-emissions
+test -d pipelines/co2-emissions
+test -f pipelines/co2-emissions/README.md
+test -f pipelines/co2-emissions/run.sh
+test -f pipelines/co2-emissions/setup.sh
+test -f pipelines/co2-emissions/validate.sh
 
 # info command
 popper info popperized/quiho-popper/single-node | grep 'Github Url'
```
UjjwalAyyangar: The `info` subcommand searches the local storage to find the pipeline. Since the single-node pipeline is not added,
this test will fail.

this needs to be changed to `popper info popperized/lesson/co2-emissions | grep 'Github Url' `

cli/popper/commands/cmd_add.py
```diff
@@ -16,55 +16,61 @@
...
-    else:
+    gh_url = 'https://github.com/{}/{}/archive/master.zip'.format(owner, repo)
+
+    pu.info("Downloading pipeline {}... ".format(pipe_name))
+
+    r = requests.get(gh_url)
+    if r.status_code != 200:
         pu.fail("Unable to fetch the pipeline. Please check if the name" +
                 " of the pipeline is correct and the internet is connected")
+    z = zipfile.ZipFile(BytesIO(r.content))
```
UjjwalAyyangar: We might need to open the zipfile with a `with` clause to close the file automatically. I mean something like this :- 

```
with zipfile.ZipFile(BytesIO(r.content)) as z :
    z.extractall()
```

this would help us avoid any unwanted errors due the file not being closed properly. 

Also maybe a try and catch block for the BadZipfile errors should be kept since it seems to be a standard practice when using zip files. What do you think?

cli/popper/cli.py
UjjwalAyyangar: @ankan17 - I think it is important that we keep this. Errors like ImportError and NameError both will fall under the UsageError and in those situations a simple `Command add doesn't exist` may not be that helpful. It will save our time while debugging if we know the problem exactly.

.gitmodules
ivotron: we're not using cryptography anymore, as #317 removes it in favor of `pyes`

docs/ci/demo
ivotron: nice catch. I've updated the tests so this is properly referencing `co2-emissions`

cli/popper/commands/cmd_add.py
ivotron: I added the `with`. We might want to open a separate issue for properly handling exceptions all over our codebase (I think we have more places where we could catch known exceptions), so that when an error occurs, more user friendly messages are displayed

PR #329, opened 2018-06-04
CLI : Workflow graph generation modified

UjjwalAyyangar:
This PR makes a few modifications to the workflow subcommand from #323


UjjwalAyyangar:
I'll notify when done

ivotron:
awesome! thanks a lot

PR #336, opened 2018-06-06
[Docs] Fix help command in "getting started" and link to software carpentry lesson

luizirber:
- `popper help` doesn't work, I guess it should be `popper --help` instead
- fix link to carpentry-style lesson


ivotron:
thanks a lot!

PR #346, opened 2018-06-08
Fixes #342 Create new record for separate repository

### LINKED ISSUE #342: cli: archiving on zenodo using the same deposit
looks like the code assumes that there’s only gonna be 1 deposit for each user? Instead, it might be better to query first the available deposits for the user, and then map the project to the corresponding one (from all the available ones), or create one if one doesn’t exist.

ankan17:
When popper archive is run on new repository which has not been archived yet, a new record is created otherwise, a new version of a previously existing record is created. Also enabled test for popper archive with zenodo and fixed the tests.

### DIFF
--- a/cli/popper/archiving.py
+++ b/cli/popper/archiving.py
@@ -10,7 +10,7 @@ import popper.utils as pu
 from datetime import date
 
 
-class BaseService():
+class BaseService(object):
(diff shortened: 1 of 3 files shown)


ivotron:
awesome!

Issue #152, opened 2017-08-09
cli: add --list-repos flag to search command


ivotron:
After #70 is implemented, this issue will add a `--list-repos` to show all the repos available to the `search` command.

Issue #148, opened 2017-08-09
cli: --add and --rm flags to search command


ivotron:
After #70 is implemented, the work for this issue consists of adding an `--add <org/repo>` flag to the `search` command to add orgs/repos to the `popperized` list in `.popper.yml`. Same for `--rm` to delete orgs/repos.

PR #347, opened 2018-06-09
CLI : Adds --rm and --add flags to popper search command

UjjwalAyyangar:
Addresses #148  and #152

### DIFF
--- a/docs/ci/demo
+++ b/docs/ci/demo
@@ -236,11 +236,16 @@ popper rm mypipetwo
 init_test
 popper search quiho
 test -d ./.cache
+popper search --list-repos
 popper search quiho --skip-update
 popper search linux --skip-update
 popper search linux --rm popperized
+set +e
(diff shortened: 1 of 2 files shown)


docs/ci/demo
```diff
@@ -231,9 +231,11 @@ popper rm mypipetwo
 init_test
 popper search quiho
 test -d ./.cache
+popper search --list-repos
 popper search quiho --skip-update
 popper search linux --skip-update
-
+popper search linux --rm popperized
```
ivotron: maybe `cat` the config file so that we make sure that `popperized` is not there after we removed it?

docs/ci/demo
```diff
@@ -231,9 +231,11 @@ popper rm mypipetwo
...
+popper search linux --add popperized
```
ivotron: same as above

ivotron:
thanks a lot!

Issue #311, opened 2018-05-16
doc: switch to the sphinx readthedocs theme


ivotron:
http://sphinx-rtd-theme.readthedocs.io/en/latest/

PR #352, opened 2018-06-13
switch to the sphinx readthedocs theme 

Franko1307:
Addresses https://github.com/systemslab/popper/issues/311


UjjwalAyyangar:
@Franko1307 Take a look at git commit `--amend`.  It's helpful when you want to add something to the same commit message (like "logo" in the above commits).

docs/conf.py
```diff
@@ -1,3 +1,5 @@
+from recommonmark.parser import CommonMarkParser
```
UjjwalAyyangar: I think this module needs to be vendored. (from https://github.com/rtfd/recommonmark).
Also, it needs to be added to the setup.py file.

ivotron:
@UjjwalAyyangar no need to squash/rewrite commits on a PR branch, github does it automatically.

docs/conf.py
ivotron: this PR is only touching the `docs/` folder, so it's separate from the `cli/` folder

ivotron:
@Franko1307 awesome, looks great!

PR #364, opened 2018-06-20
Fixes #332. Add popper matrix subcommand

### LINKED ISSUE #332: cli: add support for environment variables in 'ci' subcommand
Add ability to define environment variables when `popper ci` is executed. This can be handled by having an `--env-var` flag that can be provided multiple times, in `KEY=VALUE` format. When the config files for a CI service get generated, this variables will be defined (e.g. using `env:` item in Travis' YAML syntax), and also passed down to Docker when an execution environment is not `host`.

ankan17:
popper matrix subcommand can be used to define an execution matrix for a pipeline using environment variables. It is used as `popper matrix pipeline --add key1=val1` to add a key and `popper matrix pipeline --rm key1` to remove a key.

### DIFF
--- a/cli/popper/commands/cmd_matrix.py
+++ b/cli/popper/commands/cmd_matrix.py
@@ -1,0 +1,58 @@
+import click
+import popper.utils as pu
+
+from popper.cli import pass_context
+from popper.exceptions import BadArgumentUsage
+
+
+@click.command('matrix', short_help='Define execution matrix for a '
+               'pipeline using environment variables.')
+@click.argument('pipeline', required=True)
+@click.option(
+    '--add',
+    multiple=True,
+    help='Add environment variable in the form key=value',
+    required=False
+)
+@click.option(
+    '--rm',
+    multiple=True,
+    help='Remove environment variable using the key',
+    required=False
+)
+@pass_context
+def cli(ctx, pipeline, add, rm):
+    try:
+        config = pu.read_config()
+        pipeline_config = config['pipelines'][pipeline]
+    except KeyError:
+        raise BadArgumentUsage("Pipeline {} does not exist.".format(pipeline))
+    if not add and not rm:
+        try:
+            pu.print_yaml(pipeline_config['matrix']['vars'])
+        except KeyError:
+            pu.info("No environment variables defined for this pipeline.")
+    if add:
+        try:
+            matrix = pipeline_config['matrix']
+        except KeyError:
+            matrix = {'vars': {}}
+        for var in add:
+            key, val = var.split('=')
+            matrix['vars'][key] = val
+        config['pipelines'][pipeline]['matrix'] = matrix
+        pu.write_config(config)
+    if rm:
+        try:
+            matrix = pipeline_config['matrix']
+        except KeyError:
+            pu.fail("No environment variables defined for this pipeline.")
+
+        for key in rm:
+            try:
+                del matrix['vars'][key]
+            except KeyError:
+                raise BadArgumentUsage("Environment variable {} does not "
+                                       "exist in the matrix".format(key))
+        config['pipelines'][pipeline]['matrix'] = matrix
+        pu.write_config(config)
--- a/docs/ci/demo
+++ b/docs/ci/demo
@@ -210,6 +210,15 @@ popper cleanup
 ! popper stages mypipetwo | grep teardown
 ! popper stages mypipetwo | grep post-run
 
+# popper matrix subcommand
+init_test
+popper init mypipeone
+popper matrix mypipeone --add key1=val1 --add key2=val2
+popper matrix mypipeone | grep -q key1
+popper matrix mypipeone | grep -q key2
+popper matrix mypipeone --rm key1
+! popper matrix mypipeone | grep -q key1
+
 # test popper rm command
 init_test
 popper init mypipetwo


cli/popper/commands/cmd_matrix.py
```diff
@@ -0,0 +1,58 @@
...
+    help='Remove environment variable using the key',
+    required=False
+)
+@pass_context
+def cli(ctx, pipeline, add, rm):
+    try:
+        config = pu.read_config()
+        pipeline_config = config['pipelines'][pipeline]
+    except KeyError:
+        raise BadArgumentUsage("Pipeline {} does not exist.".format(pipeline))
```
ivotron: maybe we should add a `pu.read_pipeline_config` as a convenience for doing all this? There are other few places where we do this

cli/popper/commands/cmd_matrix.py
```diff
@@ -0,0 +1,58 @@
...
+    if not add and not rm:
```
ivotron: we also need to check that both options are _not_ given at the same time

cli/popper/commands/cmd_matrix.py
```diff
@@ -0,0 +1,58 @@
...
+    if add:
+        try:
+            matrix = pipeline_config['matrix']
+        except KeyError:
+            matrix = {'vars': {}}
+        for var in add:
+            key, val = var.split('=')
+            matrix['vars'][key] = val
+        config['pipelines'][pipeline]['matrix'] = matrix
+        pu.write_config(config)
```
ivotron: we need to exit here

cli/popper/commands/cmd_matrix.py
```diff
@@ -0,0 +1,58 @@
...
+    if not add and not rm:
+        try:
+            pu.print_yaml(pipeline_config['matrix']['vars'])
+        except KeyError:
+            pu.info("No environment variables defined for this pipeline.")
+    if add:
+        try:
+            matrix = pipeline_config['matrix']
+        except KeyError:
+            matrix = {'vars': {}}
```
ivotron: since we always expect `vars`, maybe it should be added as part of `popper init`. Also, and more importantly, `vars` is a list of dictionaries. For every new item in the one-dimensional matrix, we add a dictionary, where the dictionary has as many key-value pairs. We can discuss this further on gitter if you want.

cli/popper/commands/cmd_matrix.py
```diff
@@ -0,0 +1,58 @@
...
+    try:
+        config = pu.read_config()
+        pipeline_config = config['pipelines'][pipeline]
+    except KeyError:
+        raise BadArgumentUsage("Pipeline {} does not exist.".format(pipeline))
+    if not add and not rm:
+        try:
+            pu.print_yaml(pipeline_config['matrix']['vars'])
+        except KeyError:
+            pu.info("No environment variables defined for this pipeline.")
```
ivotron: a more pythonic way of this is to avoid generating an exception. Assuming that we always have a `matrix['vars']` entry (see below):

```python
if len(matrix['vars']) == 0:
    pu.info(....
```

docs/ci/demo
```diff
@@ -210,6 +210,15 @@ popper cleanup
...
 ! popper stages mypipetwo | grep post-run
 
+# popper matrix subcommand
+init_test
+popper init mypipeone
+popper matrix mypipeone --add key1=val1 --add key2=val2
+popper matrix mypipeone | grep -q key1
+popper matrix mypipeone | grep -q key2
+popper matrix mypipeone --rm key1
+! popper matrix mypipeone | grep -q key1
```
ivotron: this will not work as intended. For example:

```bash
popper matrix mypipeone --add key1=val1 --add key2=val2
popper matrix mypipeone --add key1=val1 --add key2=other
```

the above defines _two_ elements, so when we remove an entry, we remove it by providing all the key-value pairs in that item. Hmmm, maybe we should rename the `matrix` subcommand to `vars`?

ankan17: Becuase of the way dictionaries work in python, the above will replace key2 with the value other. So, there will only be one instance of key2.

cli/popper/commands/cmd_matrix.py
ankan17: I need to read the whole config anyways since I have to write the changes back. So, there is no requirement of creating an extra utility function for this imo.

cli/popper/utils.py
```diff
@@ -99,18 +99,21 @@ def is_popperized():
...
     config = read_config()
-    pipeline_config = config['pipelines'][name]
-
-    if not stages:
-        stages = ','.join(pipeline_config['stages'])
-    if not envs:
-        envs = ','.join(pipeline_config['envs'])
-    if not relative_path:
-        relative_path = pipeline_config['path']
+    try:
```
ankan17: I am not sure if this will work perfectly. @ivotron Please have a look.

cli/popper/commands/cmd_env-vars.py
```diff
@@ -0,0 +1,84 @@
...
+            key, val = var.split('=')
+            vars_add[key] = val
+        env_vars.append(vars_add)
+        pu.update_config(pipeline, vars=env_vars)
+
+    elif rm:
+        try:  # This try catch block ensures backward compatibility
+            env_vars = config['vars']
+        except KeyError:
+            pu.fail("No environment variables defined for this pipeline.")
```
ivotron: same as above (pythonic version):

```python
if not config.get('vars', None):
    pu.fail("No environment variables defined for this pipeline.")
```

cli/popper/commands/cmd_env-vars.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if add and rm:
+        raise UsageError("Both add and rm cannot be given at the same time. "
+                         "See popper env-vars --help for more information.")
+
+    if add:
+        try:  # This try catch block ensures backward compatibility
+            env_vars = config['vars']
+        except KeyError:
+            env_vars = []
```
ivotron: pythonic way of doing this:

```python
env_vars = config.get('vars', []):
```

cli/popper/utils.py
```diff
@@ -96,21 +96,44 @@ def is_popperized():
...
 
 
+def read_pipeline_config(name):
+    """Returns the configuration for a pipeline."""
+
+    config = read_config()
+    try:
+        return config['pipelines'][name]
+    except KeyError:
+        fail("Pipeline {} does not exist.".format(name))
```
ivotron: This could return a 2-item tuple (`return config, pipeline_config`), so that we don't need to read the config separately.

cli/popper/commands/cmd_env-vars.py
```diff
@@ -0,0 +1,84 @@
...
+        vars_add = {}
+        for var in add:
+            key, val = var.split('=')
+            vars_add[key] = val
+        env_vars.append(vars_add)
+        pu.update_config(pipeline, vars=env_vars)
```
ivotron: If we make `pu.read_pipeline_config(pipeline)` return a tuple (see below), we can use `pu.write_config()` directly

cli/popper/utils.py
```diff
@@ -96,21 +96,44 @@ def is_popperized():
...
-    if name == 'paper':
-        stages = 'build'
+def update_config(name, stages='', envs='', vars=[], relative_path=''):
+    """Updates the configuration for a pipeline."""
 
     config = read_config()
+    try:
+        pipeline_config = config['pipelines'][name]
+
+        if name == 'paper':
```
ivotron: a `paper` pipeline is the same as any other pipeline, there is no need to create a special case for it

cli/popper/utils.py
```diff
@@ -96,21 +96,44 @@ def is_popperized():
...
+            stages = 'build'
+        else:
+            if not stages:
+                stages = ','.join(pipeline_config['stages'])
+            if not envs:
+                envs = ','.join(pipeline_config['envs'])
+            if not relative_path:
+                relative_path = pipeline_config['path']
+    except KeyError:  # Pipeline does not exist yet
+        pass
```
ivotron: pythonic try-catch replacement:

```python
if config['pipelines'].get(name, None):
    if not stages:
        stages = ','.join(config[name]['stages'])
    if not envs:
        envs = ','.join(config[name]['envs'])
    if not relative_path:
        relative_path = config[name]['path']
```

ivotron:
I forgot: can you please rebase master so that tests are added to the popper pipeline

cli/popper/commands/cmd_env-vars.py
```diff
@@ -0,0 +1,80 @@
+import click
+import popper.utils as pu
+
+from popper.cli import pass_context
+from popper.exceptions import UsageError
+
+
+@click.command('matrix', short_help='Define execution matrix for a '
+               'pipeline using environment variables.')
```
ivotron: should we rename to `env-vars`? Another alternative I was thinking was `jobs`, where a "job" is defined as an execution of the pipeline on a set of env vars. Then, we'll have as many jobs as number of variable sets. Another one would be `executions`, since there's one execution of the pipeline for each variable set.

ci/test-env-vars.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
ankan17: Sorry. I didn't understand. Can you please explain? @ivotron

cli/popper/commands/cmd_env-vars.py
ankan17: env-vars sounds the most appropriate.

ci/test-env-vars.sh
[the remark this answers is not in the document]
ivotron: We're now testing popper using popper, so adding a new stage means that we need to modify .popper.yml to add this stage.

cli/popper/commands/cmd_env-vars.py
ivotron: Sounds good

cli/popper/utils.py
```diff
@@ -98,16 +111,28 @@ def is_popperized():
...
     config = read_config()
+    if config['pipelines'].get(name, None):
+        if not stages:
+            stages = ','.join(config[name]['stages'])
+        if not envs:
+            envs = ','.join(config[name]['envs'])
+        if not relative_path:
+            relative_path = config[name]['path']
+
+    if name == 'paper':
```
UjjwalAyyangar: Can this control statement be removed? 
Since it's already present in lines 117-118

Issue #355, opened 2018-06-17
doc: re-org documentation structure


ivotron:
Re-organize documentation. New proposed structure is below. We can use the checkmark boxes to keep track of progress. All work is being done on branch `docs-reorg`.

  * [ ] Getting Started (**same as current**)

  * [ ] Concepts

      * [ ] Popper pipelines
      * [ ] Pipeline portability
      * [ ] Popper and CI systems
      * [ ] Popper vs. other software

  * [ ] CLI Features

      * [ ] New pipeline initialization
      * [ ] Searching and importing existing pipelines
      * [ ] Executing a pipeline
      * [ ] Continuously validating pipelines
      * [ ] Visualizing a pipeline
      * [ ] Adding metadata to a project
      * [ ] Archiving a project and obtaining DOIs
      * [ ] The `.popper.yml` configuration

  * [ ] Examples

      * [ ] Pipeline portability

          * [x] Using Virtualenv (Python)
          * [ ] Using Packrat (R)
          * [ ] Using Spack
          * [x] Using Docker
          * [ ] Using Vagrant

      * [ ] Dataset management

          * [x] Using datapackages
          * [x] Using data.world

      * [ ] Infrastructure automation

          * [x] On CloudLab using geni-lib
          * [ ] On Chameleon using enos
          * [ ] On EC2/GCE using terraform

      * [ ] Domain-specific pipelines

          * [ ] Atmospheric science
          * [ ] Machine learning
          * [ ] Linux kernel development
          * [ ] Relational databases
          * [ ] Genomics
          * [ ] High-energy physics

      * [ ] Results validation

          * [ ] Codifying validation of results
          * [ ] Parameterization and parameter sweeps
          * [ ] Continuous validation (CI setup)
          * [ ] Artifact evaluation criteria

  * [ ] Contributing

      * [ ] Code of Conduct
      * [ ] Contributing New CLI Features
      * [ ] Contributing New Example Pipelines

PR #365, opened 2018-06-21
Adds virtualenv documentation and examples

Franko1307:
addresses https://github.com/systemslab/popper/issues/355

### DIFF
--- a/docs/sections/examples.md
+++ b/docs/sections/examples.md
@@ -20,16 +20,15 @@ that need to be followed in order to create these isolated environments.
 
 Examples:
 
-[sea-surface-mapping](https://github.com/popperized/swc-lesson-pipelines/tree/master/pipelines/sea-surface-mapping)
-In this case, Popper helps make sure that a virtualenv is enabled and 
-running at the time of execution by throwing an exception if your 
-virtualenv is disabled.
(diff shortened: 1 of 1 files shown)


docs/sections/examples.md
```diff
@@ -6,7 +6,30 @@
...
+
+This tool is [virtualenv](https://virtualenv.pypa.io/en/stable/) and it can be used 
+to create an isolated environment with all the dependencies of a python application,
+including the version of the python runtime itself. 
+This is a lightweight way of creating portable pipelines.
+
+Popper pipelines automate and create an explicit record of the steps
+that need to be followed in order to create these isolated environments.
+
+Examples:
```
ivotron: We might want to create this as a markdown list

docs/sections/examples.md
```diff
@@ -6,7 +6,30 @@
...
+
+[sea-surface-mapping](https://github.com/popperized/swc-lesson-pipelines/tree/master/pipelines/sea-surface-mapping)
+In this case, Popper helps make sure that a virtualenv is enabled and 
+running at the time of execution by throwing an exception if your 
+virtualenv is disabled.
+
+Or if you want Popper to make the virtualenv for you, you can check the following pipeline:
```
ivotron: This might give the impression that popper is doing the work but actually the pipeline is the one that codifies the logic of creating the virtual environment. So maybe reword to something like: "Another alternative is to create a virtual environment as part of the execution of a pipeline. The `validator` pipeline ..."

Issue #328, opened 2018-06-03
cli: popper on popper


ivotron:
create a popper pipeline for testing the cli. Each stage can test distinct aspects of the `popper` command. This will basically break down into pieces the [`docs/ci/demo`](https://github.com/systemslab/popper/blob/master/docs/ci/demo) script.

ankan17:
Shouldn't this be closed now?? @ivotron

ivotron:
@ankan17 yes, thanks!

PR #366, opened 2018-06-21
cli: popper on popper

eltrufas:
Addresses #328


ci/common-setup.sh
```diff
@@ -0,0 +1,16 @@
...
+  rm -rf /tmp/mypaper
+  mkdir /tmp/mypaper
+  cd /tmp/mypaper
+  git init
+  popper init
+}
+
+if USE_VIRTUALENV
+then
+  source /tmp/popper-env/bin/activate
```
ivotron: On line 8 of `.travis.yml`, we are adding the `cli/bin` folder to `PATH` which makes `cli/bin/popper` the binary being executed when `popper` is used subsequently. After the virtualenv gets activated, do you know which of the two available `popper` commands gets invoked? My guess is that the one from virtualenv has higher precedence but it might be better to double check so that we are certain

.travis.yml
```diff
@@ -1,7 +1,13 @@
+
+---
 language: python
-python: 3.6
+python: 2.7
 services: docker
+install:
+- git clone --recursive https://github.com/systemslab/popper /tmp/popper
```
ivotron: we might want to avoid cloning master and, instead, delete this line and use the version that we are testing. So the `export` line below would be:

```bash
export PATH=$PATH:$PWD/cli/bin
```

.travis.yml
```diff
@@ -1,7 +1,13 @@
...
+- export PATH=$PATH:/tmp/popper/cli/bin
+- export PYTHONUNBUFFERED=1
 script:
-- PATH=$PATH:$PWD/cli/bin docs/ci/demo
-- docker build -t falsifiable/popper cli/
-- docker run --rm -e CI=1 -e POPPER_FIGSHARE_API_TOKEN="${POPPER_FIGSHARE_API_TOKEN}" -e POPPER_ZENODO_API_TOKEN="${POPPER_ZENODO_API_TOKEN}" -e POPPER_GITHUB_API_TOKEN="${POPPER_GITHUB_API_TOKEN}" -v `pwd`/docs/ci/demo:/demo --entrypoint=/demo falsifiable/popper
+  - USE_VIRTUALENV=1 popper run
```
ivotron: we also need to execute without creating a virtualenv, so that we can test [manual installs](https://github.com/systemslab/popper/tree/master/cli#manual) of popper.

ci/common-setup.sh
eltrufas: I did test this out by running `which popper` in the script after sourcing the virtualenv, which indeed outputs `/tmp/popper-env/bin/popper`. I'll put the `which` in `setup.sh`, just for the sake of debugging.

.travis.yml
eltrufas: yeah, I'll go ahead and do that.

.travis.yml
eltrufas: Done. Thanks!

eltrufas:
Moving the branch to the main repo to fix the gh api rate limiting.

PR #369, opened 2018-06-22
Fix cleanup

UjjwalAyyangar:
* Fixes popper cleanup 
* Fixes the bad argument bug in `popper init --existing`
* Adds a `popper reset` command for completely resetting a popper repository.


cli/popper/commands/cmd_cleanup.py
```diff
@@ -15,26 +14,34 @@ def cli(ctx):
     """Synchronize your pipelines and popper.yml file if any pipeline or stage
     has been deleted.
     """
-    pipeline_dir = os.path.join(pu.get_project_root(), 'pipelines')
+    #pipeline_dir = os.path.join(pu.get_project_root(), 'pipelines')
```
ivotron: instead of commenting, we can remove it

cli/popper/commands/cmd_init.py
```diff
@@ -30,8 +30,8 @@
 @click.option(
     '--infer-stages',
     help=('Infers the name of the stages from the existing bash scripts'
-          'on the folder while creating a new pipeline from the existing'
-          'folder.'),
+          ' on the folder while creating a new pipeline from the existing'
+          ' folder.'),
```
ivotron: maybe we could reword this to:

```
Infers the stages of a pipeline from bash script file names.
Used in conjunction with the --existing flag.
```

then user can read the help for `init` and for `--existing`

UjjwalAyyangar:
@ivotron - Requested changes have been made.

ci/test-reset.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
ivotron: this stages needs to be added to `.popper.yml`

UjjwalAyyangar:
Some more changes to be made. I will notify when they are completed.

ivotron:
thanks a lot!

PR #375, opened 2018-06-25
Fixes #374. Successfully remove paper pipeline with popper rm command

### LINKED ISSUE #374: cli: popper rm does not work for paper pipeline

ankan17:


cli/popper/commands/cmd_rm.py
```diff
@@ -29,14 +29,6 @@ def cli(ctx, pipeline):
         popper_config = pu.read_config()
         del popper_config['pipelines'][pipeline]
 
-        if 'stages' in popper_config:
```
ankan17: I could not understand if this portion is necessary. Is there any case when 'stages' is directly a key of config? @ivotron

UjjwalAyyangar: Some of the old popperized, repositories had stages as a key. It has been fixed now. 
This line can be removed.

Issue #89, opened 2017-04-05
ci: skip tests based on commit message


ivotron:
e.g. `popperci:skip` to entirely skip a commit or `popperci:whitelist[exp]` for running only the `exp` pipeline

PR #377, opened 2018-06-26
skips test based on commit messages

Franko1307:
Addresses https://github.com/systemslab/popper/issues/89

### DIFF
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -56,7 +56,9 @@ def cli(ctx, pipeline, timeout, skip, ignore_errors):
         pu.info("No pipelines defined in .popper.yml. "
                 "Run popper init --help for more info.", fg='yellow')
         sys.exit(0)
+
     if os.environ.get('CI', False):
+        # if commit is null, it throws an exception. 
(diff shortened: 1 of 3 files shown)


cli/popper/commands/cmd_run.py
```diff
@@ -54,6 +55,31 @@ def cli(ctx, pipeline, timeout, skip, ignore_errors):
...
+                # `popperci:whitelist[pipeline]` and gets the pipeline.
+                pipeline = re.search('popperci:whitelist\[(.+?)\]',
+                                     commit).group(1)
+                pu.info("Executing popperci:whitelist[{}]"
+                        .format(pipeline))
+            except AttributeError:
+                pipeline = None
+                pu.warn("Couldn't find pipeline associated with the "
+                        "popperci:whitelist flag. "
+                        "Assigning pipeline to None")
```
ivotron: this should be independent of a CI service. So we need to use instead `git` to look for the commit message, e.g. to run something like `git log -1 --pretty=%B` and store that on the string and parse it. Also, we can rename `popperci:` to `popper` (that `popperci` is legacy from the golang days)

Issue #357, opened 2018-06-17
cli: add support for generating '<service> ready' badges


ivotron:
Add support for generating badges for pipelines that denote that a pipeline (or project) is ready to be used on a particular infrastructure. The badge will read `<service> ready` where `<service> can be one of:

  * CloudLab
  * Chameleon
  * GCP
  * AWS

The URL associated to the badge will point to the section in the documentation where the guide for that particular service is documented. For example, the pipeline developed in #356 will have a `CloudLab ready` badge and the badge will point to the `Examples -> Infrastructure automation -> CloudLab using geni-lib` section.

Code might be implementing a fix for #139

PR #381, opened 2018-06-27
cli: add support for generating '<service> ready' badges 

eltrufas:
Currently the command generates "service-ready" badges linking to documenation, as specified in #357.

A caveat is that the badges generated link to the docs-reorg readthedocks branch, so they will need to be updated once the new docs go online.


ci/test-badge.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+popper badge errorout
+
+if [ $? -eq 0 ];
+then
+  exit 1
+fi
+
+set -e
```
ivotron: can we test that when given the proper flag/value, the command generates the badge link?

eltrufas:
I've added the requested test.

PR #394, opened 2018-07-06
Fixes #182 Add logic for notifying badge server on popper run

### LINKED ISSUE #182: cli: update status of pipeline against a badge server
Add logic for notifying a badge server about the status of a pipeline.

ankan17:


ci/test-badge-server.sh
```diff
@@ -0,0 +1,23 @@
...
+source common-setup.sh
+init_test
+sed -i '$s/.*/badge-server-url: http:\/\/127.0.0.1:5000/' .popper.yml
+popper init mypipeone
+git config --global user.email "testuser@example.com"
+git config --global user.name "Test User"
+git remote add origin https://github.com/systemslab/popper.git
+git add .
+git commit -m "Add mypipeone"
+popper run
```
ivotron: can we check that we added a new status to the server (e.g. using `curl`)?

ankan17: There will be no need of that if we are using the badge in our readme

ivotron:
woop!

Issue #392, opened 2018-07-06
cli: add a `--confirm` flag to `reset` command


ivotron:

UjjwalAyyangar:
@ivotron Can we close this issue?

PR #396, opened 2018-07-07
Confirmation prompt added to reset subcommand

UjjwalAyyangar:
Addresses #392 
Fixes the bug in `make_gh_request` function in `utils.py`

### DIFF
--- a/cli/popper/commands/cmd_reset.py
+++ b/cli/popper/commands/cmd_reset.py
@@ -17,6 +17,8 @@ def cli(ctx):
     pipelines and folders, leaving behind a newly created .popper.yml file.
     """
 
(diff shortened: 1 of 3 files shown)


cli/popper/commands/cmd_reset.py
```diff
@@ -17,6 +17,7 @@ def cli(ctx):
     pipelines and folders, leaving behind a newly created .popper.yml file.
     """
 
+    click.confirm("Do you want to continue?", abort=True)
```
ivotron: Can we add a sentence like "This will remove all the pipeline files in this project, do you want to continue?"

UjjwalAyyangar:
@ivotron : Requested changes have been made.

PR #416, opened 2018-07-21
Fixes #410. Add pipeline to specific folder provided as a name

### LINKED ISSUE #410: cli: allow to specify a custom path for 'popper add'
currently, the `popper add` command places pipelines in folders whose name is similar to the one of the remote pipeline. For this issue, we'll modify it so that we can specify the name of the pipeline, similarly to how it is done for the `popper init` command. For example:

```bash
popper add popperized/swc-lesson-pipelines/sea-surface-mapping sea-mapping
```

the above will place the lesson on `pipelines/sea-mapping`. A folder path will also be supported, for example:

```bash
popper add popperized/nwp-popper/sandy other/path/sandy
```

the above will place the pipeline on `other/path/sandy`

ankan17:


ci/test-add.sh
```diff
@@ -14,6 +14,18 @@ test -f pipelines/co2-emissions/run.sh
...
 test -f pipelines/co2-emissions/validate.sh
 
+init_test
+popper add popperized/swc-lesson-pipelines/co2-emissions
+test -d pipelines/co2-emissions
+test -f pipelines/co2-emissions/README.md
+popper rm co2-emissions
+
+init_test
+popper add popperized/swc-lesson-pipelines/co2-emissions swc-lesson-pipelines
```
ivotron: When a `/` is in the name, we treat the string as a path. The idea for this is to work in the same way as the `popper init` command works, that is, the second argument for the command is interpreted in the following way. If `<name>` is a string without `/`, then put it in the `pipelines/` folder. If it does contains a `/`, then treat it as a path, where the last component of the path (i.e. the `basename` of the string) is the name of the pipeline and the preceding substring is the folder.

So the following tests would have to pass:

```bash
init_test
popper add popperized/swc-lesson-pipelines/co2-emissions foo/emissions

test -d foo/emissions
cat .popper.yml | grep 'emissions:'
cat .popper.yml | grep 'path: foo/emissions'
test -f foo/emissions/run.sh

init_test
popper add popperized/swc-lesson-pipelines/co2-emissions foo

test -d pipelines/foo
cat .popper.yml | grep 'foo:'
cat .popper.yml | grep 'path: pipelines/emissions'
test -f pipelines/foo/run.sh

init_test
popper add popperized/swc-lesson-pipelines/co2-emissions foo/
test -d foo/
cat .popper.yml | grep 'foo:'
cat .popper.yml | grep 'path: foo'
test -f foo/run.sh
```

cli/popper/commands/cmd_add.py
```diff
@@ -110,3 +132,22 @@ def get_config(owner, repo):
...
+            if e.errno == EPERM or e.errno == EACCES:
+                pu.fail(
+                    "Could not create the necessary path.\n"
+                    "Please make sure you have the correct permissions."
+                )
+            elif e.errno == ENOENT:
+                create_path(os.path.join(os.path.split(path)[0]))
+                os.mkdir(path)
+            else:
+                pu.fail("Failed due to unknown reasons.")
```
ivotron: there is `os.makedirs()` that implements this functionality

cli/popper/commands/cmd_add.py
```diff
@@ -47,16 +45,35 @@ def cli(ctx, pipeline, folder, branch):
...
+    if folder:
+        path, basename = os.path.split(folder)
+        if not basename:  # Only true when trailing slash is present
+            path, basename = os.path.split(path)
+        if basename:
+            new_pipe_name = basename  # Use basename as the name of pipeline
+        else:
+            pu.fail("The path is invalid.")
+    else:  # If no pipeline name is provided, use the original pipeline name
+        new_pipe_name = pipe_name
```
ivotron: can we consolidate this and the logic in https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_init.py#L93 so that we create a `utils.get_name_and_path_for_new_pipeline()` function that takes as input the name of the pipeline and returns a two-element tuple `(name, path)`.

cli/popper/commands/cmd_add.py
```diff
@@ -47,16 +45,35 @@ def cli(ctx, pipeline, folder, branch):
...
-        pu.fail("Pipeline {} already in repo.".format(pipe_name))
+    if new_pipe_name in config['pipelines']:
+        pu.fail("Pipeline {} already in repo.".format(new_pipe_name))
 
     project_root = pu.get_project_root()
+
+    if not folder:  # Put it in the pipelines directory, default case
+        folder = os.path.join('pipelines', new_pipe_name)
+    elif '/' not in folder:  # Put it in the path inside pipelines directory
+        folder = os.path.join('pipelines', folder)
```
ivotron: with a `utils.get_name_and_path_for_new_pipeline()` function the above won't be needed

ankan17:
Can this be merged? @ivotron

cli/popper/commands/cmd_add.py
```diff
@@ -70,24 +76,29 @@ def cli(ctx, pipeline, folder, branch):
...
+    with tarfile.open(mode='r:gz', fileobj=BytesIO(r.content)) as t:
         t.extractall()
 
-    os.rename('{}-{}/pipelines/{}'.format(
-        repo, branch, pipe_name), os.path.join(folder, pipe_name))
-    shutil.rmtree('{}-{}'.format(repo, branch))
+    try:
+        os.rename('{}-{}/pipelines/{}'.format(
+            repo, branch, pipe_name), pipelines_dir)
+    except OSError:
```
ivotron: is the folder not being empty the _only_ cause of an `OSError`? Maybe we should explicitly check whether the given folder is not empty. In general, exceptions are usually used to check for unexpected behavior, as opposed to control the flow of a program. A common practice is to check what the code is expecting when the exception is thrown instead. In this case, that the path is not empty (see comment above)

cli/popper/commands/cmd_add.py
```diff
@@ -47,21 +44,30 @@ def cli(ctx, pipeline, folder, branch):
...
     project_root = pu.get_project_root()
+
     pipelines_dir = os.path.join(project_root, folder)
 
     if not os.path.exists(pipelines_dir):
-        os.mkdir(pipelines_dir)
+        try:
+            os.makedirs(pipelines_dir)
+        except (OSError, IOError) as e:
+            pu.fail("Could not create the necessary path.\n")
```
ivotron: It would be also be useful to fail if the path already exists.

cli/popper/commands/cmd_add.py
```diff
@@ -47,21 +44,30 @@ def cli(ctx, pipeline, folder, branch):
...
 
     gh_url = 'https://github.com/{}/{}/'.format(owner, repo)
     gh_url += 'archive/{}.tar.gz'.format(branch)
 
-    pu.info("Downloading pipeline {}... ".format(pipe_name))
+    pu.info(
+        "Downloading pipeline {} as {}...".format(pipe_name, new_pipe_name)
```
ivotron: maybe say 'downloading pipeline {} to {}` instead as `as`

ankan17: Since I am using pipeline names for both the placeholders, that's why it says `as`. For example, if you are downloading pipeline1 but named it as pipeline2, it'll say `Downloading pipeline pipeline1 as pipeline2...` which seems right, imo? pipeline2 is the name, not the path. What do you say? @ivotron

ivotron: sure, that's ok

cli/popper/commands/cmd_add.py
ivotron: Yes, we should do both.

Issue #362, opened 2018-06-19
cli: matrix execution


ivotron:
Issue #332 introduced the definition of a list of executions, where each execution is defined by a set of environment variables. In addition, #213 will implement the option of executing a pipeline on multiple environments.

For this issue, we will work on the changes to the `run` subcommand so that it can execute the bi-dimensional `(env_vars x exec_envs)` matrix. In summary:

  * create a list of executions by combining all environment variable sets and all execution environments for a pipeline.
  * iterate over each element in the list
  * create output folders for each element

PR #420, opened 2018-07-24
cli: matrix execution

Franko1307:
https://github.com/systemslab/popper/issues/362


cli/popper/commands/cmd_run.py
```diff
@@ -96,6 +96,39 @@ def cli(ctx, pipeline, timeout, skip, ignore_errors, output,
         pu.fail("Failed to execute pipeline")
 
 
+def check_for_env_vars(args):
```
ivotron: We can avoid having this function if we create a list of executions. So first get a list of executions for a pipeline, and then we pass the list to the `run_pipeline` function. For pipelines where there's no `env_vars` defined, we create one execution (i.e. the empty set of vars).

cli/popper/commands/cmd_run.py
```diff
@@ -96,6 +96,39 @@ def cli(ctx, pipeline, timeout, skip, ignore_errors, output,
...
+    env_vars = args[-1]
+    if env_vars:
+        return run_env_vars(args)
+    else:
+        return run_pipeline(*args[:-1])
+
+
+def run_env_vars(args):
```
ivotron: if the above gets implemented, then we don't need a specific case for `run_env_vars` vs. `run_pipeline`, in other words, the logic of handling a list of executions will be part of the `run_pipeline` function.

ivotron:
awesome! thanks a lot for working on this. would it be possible to rebase master?

PR #421, opened 2018-07-24
Fixes #418 Notify the server about the branch

### LINKED ISSUE #418: badges: add support for branches when updating the badge status
server needs to understand branches and return only badges for master by default

ankan17:
The code for the server has been modified.

### DIFF
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -145,11 +145,26 @@ def update_badge(status):
             'status': status
         }
         try:
-            r = requests.post(badge_server_url, data=data)
-            if r.status_code != 201:
-                pu.warn("Could not create a record on the badge server")
-        except requests.exceptions.RequestException:
-            pu.warn("Could not communicate with the badge server")
+            commit_id = check_output(['git', 'rev-parse', 'HEAD'])[:-1]
+            branch_name = check_output(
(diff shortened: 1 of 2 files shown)


cli/popper/commands/cmd_run.py
```diff
@@ -130,11 +130,26 @@ def update_badge(status):
             'status': status
         }
         try:
-            r = requests.post(badge_server_url, data=data)
-            if r.status_code != 201:
-                pu.warn("Could not create a record on the badge server")
-        except requests.exceptions.RequestException:
-            pu.warn("Could not communicate with the badge server")
+            commit_id = check_output(['git', 'rev-parse', 'HEAD'])[:-1]
```
ivotron: there's a [`get_head_commit()`](https://github.com/systemslab/popper/blob/master/cli/popper/utils.py#L291) function available now, can we use that?

cli/popper/commands/cmd_run.py
```diff
@@ -130,11 +130,26 @@ def update_badge(status):
...
+            branch_name = check_output(
+                ['git', 'rev-parse', '--abbrev-ref', 'HEAD']
+            )[:-1]
+            data = {
+                'timestamp': int(time.time()),
+                'commit_id': commit_id,
+                'branch': branch_name,
+                'status': status
+            }
```
ivotron: looks like this is duplicating what's on line 127, should we consolidate it?

PR #423, opened 2018-07-24
Update docs to use popper run

eltrufas:
Updates the cli features section to use `popper run` instead of `popper check`.

[skip ci]

### DIFF
--- a/docs/sections/cli_features.md
+++ b/docs/sections/cli_features.md
@@ -57,7 +57,7 @@ popper run
 Once a pipeline is run, one can show the logs:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/sections/cli_features.md
```diff
@@ -54,7 +54,7 @@ or to execute all the pipelines in a project:
 popper run
 ```
 
-Once a pipeline is checked, one can show the logs:
+Once a pipeline is run, one can show the logs:
 
 ```bash
 ls -l pipelines/myexp/popper_logs
```
ivotron: I think we also need to change this to `pipelines/myexp/popper/`??

Issue #425, opened 2018-07-25
Skipping based on commit message fails on python 3


eltrufas:
Currently, when I run the `test-run.sh` script locally on python 3.6 it'll fail with this message:

```
+ popper init --stages=setup mypipeone
Initialized pipeline mypipeone
+ popper init --stages=setup mypipetwo
Initialized pipeline mypipetwo
+ popper init --stages=setup mypipethree
Initialized pipeline mypipethree
+ git add .
+ git commit --allow-empty -m 'popper:whitelist[mypipeone] this is a test'
[master (root-commit) 7a15fe7] popper:whitelist[mypipeone] this is a test
 8 files changed, 48 insertions(+)
 create mode 100644 .gitignore
 create mode 100644 .popper.yml
 create mode 100644 pipelines/mypipeone/README.md
 create mode 100755 pipelines/mypipeone/setup.sh
 create mode 100644 pipelines/mypipethree/README.md
 create mode 100755 pipelines/mypipethree/setup.sh
 create mode 100644 pipelines/mypipetwo/README.md
 create mode 100755 pipelines/mypipetwo/setup.sh
+ popper run --no-badge-update
Traceback (most recent call last):
  File "/Users/trufas/popper/popper-fork/cli/bin/popper", line 29, in <module>
    popper.cli.cli()
  File "/Users/trufas/.pyenv/versions/popper/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/trufas/.pyenv/versions/popper/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/trufas/.pyenv/versions/popper/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/trufas/.pyenv/versions/popper/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/trufas/.pyenv/versions/popper/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/trufas/.pyenv/versions/popper/lib/python3.6/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
[...]

PR #426, opened 2018-07-25
Fixes skipping based on commit message on python 3

eltrufas:
Addresses #425


ankan17:
@eltrufas Please rebase your branch with `master` so that travis tests this with python3 too. That way we can be sure that the error is gone.

ivotron:
awesome, thanks a lot @eltrufas @ankan17

Issue #422, opened 2018-07-24
`popper init` writes outdated entries to `.gitignore`


eltrufas:
At the moment `popper init` will add `popper_logs` and `popper_status` to `.gitignore`. These both got changed into a single `popper` directory. Update `popper init` to write the appropriate files.

PR #427, opened 2018-07-25
Add correct entries to '.gitignore'

eltrufas:
Addresses #422.


eltrufas:
Not sure why this is failing. Might be that adding `popper` to the `.gitignore` file has some weird interaction with the tests.

eltrufas:
Turns out you need (the -x)[https://git-scm.com/docs/git-clean/2.3.3] flag for `git clean` to also remove ignored files.

ivotron:
thanks a lot!

Issue #393, opened 2018-07-06
1.1.0 umbrella issue


ivotron:
This issue tracks the tasks needed for 1.1.0 version:

## CLI features

  * [x] add a `--confirm yes` flag to `reset` subcommand (#392)
  * [x] support for multiple execution environments (#213)
  * [x] support for arbitrary execution environments, not just the ones in our docker hub (#307)
  * [x] `require` for env variables (#371).
  * [x] matrix execution (#362)
  * [x] support for Jenkins (#130)

## docs

  * new reorganized docs/ folder (#355)
      * cli features
          * [x] search and import
          * [x] visualizing a pipeline
          * [x] adding metadata to a project
          * [x] archiving a project and obtaining DOIs
      * other resources
          * [x] automated artifact evaluation
  * new CI section
      * [ ] describe environments
      * [ ] describe list of executions
      * [x] describe how matrix execution works
  * landing page updates:
      * [ ] update screencast on landing page
      * [x] update links
  * update README
      * [x] add logo
      * [x] update links

## Example pipelines

  * [x] GCE
  * [x] Chameleon
  * [x] CloudLab

## Blog
  * [x] Create schedule for weekly posts

PR #448, opened 2018-08-03
Update readme

Franko1307:
update README
  add logo
  add screencast to README
  update links

addresses https://github.com/systemslab/popper/issues/393


README.md
```diff
@@ -1,4 +1,6 @@
-# Popper
+<img src="docs/figures/popper_logo.png" width="200">
```
ankan17: This can be changed to the markdown syntax instead of mixing html and markdown.

README.md
```diff
@@ -14,12 +16,15 @@ Popper is a convention and CLI tool for generating reproducible papers. This rep
 
 Quick links to other resources:
 
-  * [5-minute screencast demo of the CLI](https://asciinema.org/a/xzEYfI3U4H5CWdmjBe0O2CAw8).
   * [Slidedeck introducing the convention](https://www.slideshare.net/ivotron/the-popper-experimentation-protocol-and-cli-tool-86987253).
   * [10-min video recording of intro and demo](https://air.mozilla.org/mozilla-open-leaders-round-4-final-demos-open-succulent-project/#@52m0s).
   * [Software Carpentry formatted Lesson](https://popperized.github.io/swc-lesson/).
   * [List of repositories that follow the convention](https://github.com/popperized).
 
+#### 5 minute screencast demo of the CLI
```
ankan17: I think it will be better if the screencast can be moved above the quick links section.

README.md
Franko1307: Yeah sure, the problem with the Markdown syntax is that it doesn't support a size parameter on GitHub. Take a look of how the logo looks like and if you're okay with the size I'll use that syntax. Or if you know a way to change the size using the Markdown syntax please let me know.

ankan17: This is simple. You can use a resized logo instead. You can just resize it and add a `small_logo.png` or something like that and then use it.

Franko1307: I guess you're right. I made the changes :smile:

README.md
```diff
@@ -12,14 +11,20 @@ Popper is a convention and CLI tool for generating reproducible papers. This rep
     convention and the CLI tool.
   * Contents of our [landing page](http://falsifiable.us) ([`gh-pages`](https://github.com/systemslab/popper/tree/gh-pages) branch).
 
+
+#### 5 minute screencast demo of the CLI
+  
+  [![asciicast](https://asciinema.org/a/xzEYfI3U4H5CWdmjBe0O2CAw8.png)](https://asciinema.org/a/xzEYfI3U4H5CWdmjBe0O2CAw8)
```
UjjwalAyyangar: Can the size of the image be reduced?

ivotron: It might be better to leave it out as we had it before. Or to reduce the size of it and put it to the end.

README.md
```diff
@@ -1,9 +1,8 @@
-# Popper
-
+![popper logo](docs/figures/rsz_popper_logo.png)
+#
```
ivotron: Can we reduce the size of the image and/or center it? Also, having the badges next to the title of the project (like the [hashicorp](https://github.com/hashicorp) projects), and then the logo something like:

# Popper [Build Status](https://travis-ci.org/systemslab/popper) [Join the chat at https://gitter.im/systemslab/popper](https://gitter.im/falsifiable-us/popper?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)  [PyPI version](https://badge.fury.io/py/popper) [GitHub license](https://github.com/systemslab/popper/blob/master/LICENSE)[Popper Status](http://popper.rtfd.io/en/latest/sections/badge_server.html)

<p align="center" style="text-align:center;">
  
</p>

Popper is a convention and CLI tool for generating reproducible papers...

Franko1307: It should be okay now @ivotron

PR #449, opened 2018-08-03
new CI section

Franko1307:
describe how matrix execution works

### DIFF
--- a/docs/sections/ci_features.md
+++ b/docs/sections/ci_features.md
@@ -73,7 +73,7 @@ following command:
 
 `popper env your-popper-pipeline --add debian-9`
 
-###### NOTE: Docker is required in order to run your project on an execution environment different from host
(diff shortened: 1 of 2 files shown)


docs/sections/ci_features.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+Please notice that Popper automatically uses host as your default execution environment.
+Then, so far we have a matrix execution which consists in the combinations between host
+and your sets of environment variables. Let's add Debian 9 as an execution environment with the
+following command:
+
+`popper env your-popper-pipeline --add debian-9`
+
+###### NOTE: Docker is required in order to run your project on an execution environment different from host
```
ivotron: This can be a `> **NOTE**` instead of a 6-level header

Franko1307: Sure, I made the change. @ivotron

Issue #433, opened 2018-07-30
cli: add project README content as part of 'popper init'


ivotron:
Similar to #432, we will add more content to the README of the project (this depends on #434  to be done. For example:

# <repo-name>

This repository contains [Popper](https://github.com/systemslab/popper) pipelines. To show a list of available pipelines using the [`popper` CLI tool](https://github.com/systemslab/popper):

```bash
cd <repo-name>
popper ls
```

to execute one of the pipelines:

```bash
popper run <pipeline-name>
```

where `<pipeline-name>` is one of the pipelines in the repository. For more on what other information from this repository is available, you can run:

```bash
popper --help
```

Issue #432, opened 2018-07-30
cli: add README content as part of 'popper init <pipeline>'


ivotron:
Currently, when `popper init <pipeline>` is executed, we only add a header to the `README.md` file. For this issue, we will add more content to the generated README, with a list of stages and generic info on how to add a pipeline using `popper add`, and how to run it using `popper run`.

Idea for a content template follows (between horizontal lines):

-----------------------------------------------

# `<pipeline-name>`

**TODO**: insert high-level description of the pipeline. It consists of the following stages:

  * [`<one>`](./one.sh). **TODO**. Add high-level description of stage `<one>`.
  * [`<two>`](./two.sh). **TODO**. Add high-level description of stage `<two>`.

# Obtaining the pipeline

To add this pipeline to  the [`popper` CLI tool](https://github.com/systemslab/popper):

```bash
cd your-repo
popper add org/<name-of-repo>/<pipeline-name>
```

**TODO**: replace `org` appropriately.

# Running the pipeline

To run the pipeline using the [`popper` CLI tool](https://github.com/systemslab/popper):

```bash
cd <name-of-repo>
popper run <pipeline-name>
```

The pipeline expects the following environment variables:

  * `<ENV_VAR1>`. Description of environment variable.
  * `<ENV_VAR2>`. Description of environment variable.

For example, the following is an execution with all expected 
variables:

```bash
export <ENV_VAR1>=value-for-<ENV_VAR_1>
export <ENV_VAR2>=value-for-<ENV_VAR_2>

popper run <pipeline-name>
```

# Dependencies

**TODO**: add list of dependencies, for example:

  * Python.
  * C++ compiler.
  * [Docker](https://docker.com) (for generating plots).
  * etc.

-----------------------------------------------

In the above template, the content in `<>` needs to be replaced by the proper values that are given as arguments to the `popper init` command, such as name of pipeline for the `<pipeline-name>` placeholder, name of stages, etc. The `**TODO**` tasks are intended for users to complete.

Issue #243, opened 2018-03-04
cli: add test for popper version


UjjwalAyyangar:
Need a bash script that can check if  the popper version and python setup.py sdist both show the same version. This bash script can then be invoked from .travis.yml so that we can check this whenever we run tests.

UjjwalAyyangar:
I'll work on it and send in a  pull request.

ivotron:
Thanks @UjjwalAyyangar , this test can be added to the [`docs/ci/demo`](https://github.com/systemslab/popper/blob/master/docs/ci/demo) bash script.

ankan17:
Is this closed @UjjwalAyyangar ?

UjjwalAyyangar:
No. Will include the script in my current PR. Thanks for notifying.

PR #457, opened 2018-08-08
CLI : Auto generates readme templates for popper

UjjwalAyyangar:
Addresses :
* #432
*  #433 
*  #458 (reset subcommand for resetting a popper repository)
* #243 

The current sample can be seen [here](https://github.com/UjjwalAyyangar/MockPipeline/tree/master/pipelines/dummy_pipeline)

### DIFF
--- a/cli/popper/template.py
+++ b/cli/popper/template.py
@@ -10,6 +10,31 @@ class ReadMe:
         with open(os.path.join(path, 'README.md'), 'w') as f:
             f.write(content)
 
+    def init_project(self):
+        content = """
+        # <repo-name>
+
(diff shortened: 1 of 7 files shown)


UjjwalAyyangar:
Since the pep8-linting is not required for comments

@pep8speaks Keep Quiet

cli/popper/template.py
```diff
@@ -0,0 +1,143 @@
...
+# Obtaining the pipeline
+
+To add this pipeline to  the [`popper` CLI tool](https://github.com/systemslab/popper):
+
+```bash
+cd your-repo
+popper add org/{0}/{1}
+```
+
+**TODO**: replace `org` appropriately.
```
ivotron: We could try to get the `org` name using git first? And if that fails, then we add the `**TODO**` node.

Issue #499, opened 2019-02-24
cli: add --recursive flag to 'popper run' to run any workflow in the repo


ivotron:
Find any `main.workflow` and run it

realsdx:
I want to work on this issue. How should i proceed?

ivotron:
Hi @realsdx, thanks a lot for offering to work on this! For this, the [`cmd_run.py`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_run.py) script needs to be updated so that a `Workflow` object is created (and executed) for each `main.workflow` file found in any descendent of `$PWD`.

PR #523, opened 2019-03-06
Written fix for issue #499 

akashrajr1:
Added --recursive for issue #499. @ivotron please review it

### DIFF
--- a/ci/test/recursive
+++ b/ci/test/recursive
@@ -28,4 +28,23 @@ do
     create_workflow $i
 done
 
-popper run --recursive
+popper run --recursive >> out
+
+files=`grep "^Found .*" out | cut -f6 -d' ' | cut -d'/' -f4-`
+
+counter=0
+set +x
+
+for workflow in "${workflows[@]}"
+do
+    for file in $files
+    do
+        if [ $workflow = $file ]
+        then
+            echo "$file was run!"
+            ((counter++))
+        fi
+    done
+done
+
+echo "Ran $counter/5 cases successfully"
--- a/ci/test/recursive
+++ b/ci/test/recursive
@@ -1,0 +1,31 @@
+#!/bin/bash
+set -ex
+
+source ./common
+init_test_repo
+cd $test_repo_path
+
+function create_workflow(){
+    cat > $1 << EOF
+workflow "test" {
+  resolves = "show env"
+}
+
+action "show env" {
+  uses = "actions/bin/sh@master"
+  args = "ls"
+}
+EOF
+}
+
+mkdir another
+mkdir another/another
+
+workflows=( "main.workflow" "another/main.workflow" "another/another/main.workflow" "try.workflow" "another/demo.workflow" )
+
+for i in "${workflows[@]}"
+do
+    create_workflow $i
+done
+
+popper run --recursive
--- a/ci/test/recursive
+++ b/ci/test/recursive
@@ -48,3 +48,4 @@ do
 done
 
 echo "Ran $counter/5 cases successfully"
+test $counter -eq 5 
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -34,10 +34,25 @@ from popper.cli import pass_context
     required=False,
     default=None
 )
+@click.option(
+    '--recursive',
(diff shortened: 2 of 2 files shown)


cli/popper/commands/cmd_run.py
```diff
@@ -34,10 +34,25 @@
     required=False,
     default=None
 )
+@click.option(
+    '--recursive',
+    help='Run any main.workflow found recursively from current path. ',
```
ivotron: This mentions `main.workflow` files but the search below is for any `.workflow` file. It's better if we make it for any `.workflow` file, so we need to update the `help` string so it reads any `Any .workflow file`.

cli/popper/commands/cmd_run.py
```diff
@@ -34,10 +34,25 @@
...
     """Executes one or more pipelines and reports on their status.
     """
+
+    if recursive:
+        for root, dirs, files in os.walk('.'):
+            for file in files:
+                if file.endswith('.workflow'):
+                    wfile = os.path.join(root, file)
+                    wfile = os.path.abspath(wfile)
+                    pu.info("Found and running workflow at "+wfile+"\n")
```
ivotron: This will only run the last `.workflow` file found in the folder hierarchy. Instead we would like to create a `Workflow` object for each of the `.workflow` files found. This would be solved by indenting lines 56-71 to the innermost level of the loop.

akashrajr1:
Thanks for the comments, will resolve and send a PR soon.
It seems I should have clarified about #499 beforehand. Does it mean to execute all files named `main.workflow`  or only `.workflow` files? 
In case it is the former might I suggest, the addition of `file` argument so that we can use `popper run --recursive main.workflow` so that only `main.workflow` files found recursively are run. It is also has the added functionality of letting the user execute any workflow files he/she wishes to. 
If it is the latter, we can use the current implementation of using it as toggle flag.

cli/popper/commands/cmd_run.py
akashrajr1: changed it to `add .workflow`

cli/popper/commands/cmd_run.py
akashrajr1: If 56-71 are shifted, won't the rest of code be dependent on `recursive` variable? Used an `else` at line 56 instead so that original functionality is maintained.

akashrajr1:
@ivotron I have left `--recursive` with its original functionality i.e. as a toggle flag

akashrajr1:
@ivotron an issue I was facing was that `--wfile` is independent of `--recursive`, so I was trying to resolve that. All of my previous commits had some dependency between them, so I separated out the part which is supposed to run the pipeline into a function and applied it independently for `--wfile` and `--recursive` options.

ci/test/recursive
```diff
@@ -0,0 +1,31 @@
...
+mkdir another/another
+
+workflows=( "main.workflow" "another/main.workflow" "another/another/main.workflow" "try.workflow" "another/demo.workflow" )
+
+for i in "${workflows[@]}"
+do
+    create_workflow $i
+done
+
+popper run --recursive
```
ivotron: we can check here whether each of them got invoked. Maybe by saving output to a file and then running `grep` on that file to search for output that indicates that all workflows were executed.

akashrajr1: had to unset `set -x`  in this [line](https://github.com/systemslab/popper/blob/158160ffb54ea0e84f08a225c85096443403f694/ci/test/recursive#L36). `set -x` wasn't able to follow through nested for loops and proper output wasn't displayed.

ci/test/recursive
```diff
@@ -0,0 +1,50 @@
...
+    do
+        if [ $workflow = $file ]
+        then
+            echo "$file was run!"
+            ((counter++))
+        fi
+    done
+done
+
+echo "Ran $counter/5 cases successfully"
```
ivotron: this is awesome! the only remaining thing to do is to check whether `counter` is actually 5, something like:

```bash
test $counter == 5
```

which will cause the script to fail (because of the `-e` flag at the top) whenever the counter does not equal to 5

ivotron:
thanks a lot!

PR #527, opened 2019-03-06
Fixes #502 Added --verbose flag

### LINKED ISSUE #502: cli: add --verbose flag to 'popper run' command
For this issue, we will add a `--verbose` flag to the [`popper run`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_run.py) command that will instruct popper to print to `stdout` the output of all the commands being invoked such as `docker build`, `docker run`, etc..

JayjeetAtGithub:
When `popper run --verbose` is run, the ouput of all the commands invoked get printed to the output.

### DIFF
--- a/cli/popper/gha.py
+++ b/cli/popper/gha.py
@@ -13,6 +13,7 @@ import sys
 class Workflow(object):
     """A GHA workflow.
     """
+
     def __init__(self, wfile, workspace):
         if not wfile:
             if os.path.isfile("main.workflow"):
@@ -163,21 +164,21 @@ class Workflow(object):
 
             cloned.add('{}/{}'.format(user, repo))
 
-    def instantiate_runners(self):
+    def instantiate_runners(self, verbose=False):
         """Factory of ActionRunner instances, one for each action"""
         for _, a in self.wf['action'].items():
             if 'docker://' in a['uses']:
                 a['runner'] = DockerRunner(a, self.workspace,
-                                           self.env, self.timeout)
+                                           self.env, self.timeout, verbose)
                 continue
 
             if './' in a['uses']:
                 if os.path.exists(os.path.join(a['uses'], 'Dockerfile')):
                     a['runner'] = DockerRunner(a, self.workspace,
-                                               self.env, self.timeout)
+                                               self.env, self.timeout, verbose)
                 else:
                     a['runner'] = HostRunner(a, self.workspace,
-                                             self.env, self.timeout)
+                                             self.env, self.timeout, verbose)
                 continue
 
             dockerfile_path = os.path.join(a['repo_dir'], a['action_dir'],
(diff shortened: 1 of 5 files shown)


JayjeetAtGithub:
@ivotron Please review

JayjeetAtGithub:
@ivotron Can you please tell me how to show the output in the `execute()` in `gha.py` ?

cli/popper/gha.py
```diff
@@ -182,22 +183,22 @@ def instantiate_runners(self):
                 a['runner'] = HostRunner(a, self.workspace,
                                          self.env, self.timeout)
 
-    def run(self, action_name=None, reuse=False):
+    def run(self, action_name=None, reuse=False, verbose=False):
         """Run the pipeline or a specific action"""
         os.environ['WORKSPACE'] = self.workspace
 
         self.download_actions()
         self.instantiate_runners()
```
ivotron: as a way of minimizing the number of functions that get the `verbose` parameter, we can pass it to the `instantiate_runners` method and then get it on the constructor of each runner, and make a `verbose` member of the base `AcitonRunner` class.

cli/popper/utils.py
```diff
@@ -483,7 +483,8 @@ def exec_cmd(cmd, ignoreerror=False):
         fail("Command '{}' failed: {}\n".format(cmd, ex))
 
     output = output.decode('utf-8')
-
+    if verbose:
+        info("{}: \n{}".format(cmd, output))
```
ivotron: while this works, this is going to print everything at once after the command has finished. Also, commands that fail will result in the code not reaching this point (they exit in line 483 above where the `fail` method is invoked).

Instead, what we would like is to stream the output of the command to `stdout`

cli/popper/gha.py
```diff
@@ -251,32 +254,44 @@ def execute(self, cmd, log_tag):
         out_fname = os.path.join(self.log_path, log_tag + '.out')
         err_fname = os.path.join(self.log_path, log_tag + '.err')
 
-        with open(out_fname, "w") as outf, open(err_fname, "w") as errf:
-            p = subprocess.Popen(cmd, stdout=outf, stderr=errf, shell=True,
-                                 preexec_fn=os.setsid)
+        if self.verbose:
+            rc = pu.exec_cmd_verbose(cmd)[0]
```
ivotron: this will result in the log files not being created, which is not the expected behavior. Instead, we would like a way of streaming to both, `stdout` and the log files.

cli/popper/gha.py
```diff
@@ -313,15 +328,19 @@ def run(self, reuse):
...
 
         return False
 
     def docker_rm(self):
-        pu.exec_cmd('docker rm {}'.format(self.cid), ignoreerror=True)
+        pu.exec_cmd(
+            'docker rm {}'.format(
+                self.cid),
+            self.verbose,
+            ignoreerror=True)
```
ivotron: same as mentioned before regarding indentation style

cli/popper/utils.py
```diff
@@ -474,9 +474,30 @@ def get_name_and_path_for_new_pipeline(folder, pipeline_name=''):
     return new_pipeline_name, path
 
 
-def exec_cmd(cmd, ignoreerror=False):
+def exec_cmd_verbose(command):
```
ivotron: Given that this is similar to what's on `ActionRunner.execute`, we can take this opportunity to create a utility function for executing external commands with parameters to suppress output, only write output to `.out` and `.err` logs, or both.

JayjeetAtGithub: Does suppressing output mean printing ouput to STDOUT ?

ivotron: Sorry, I meant, four options:
  * suppress output (don't write output at all)
  * write to `stdout`
  * write to log files
  * both (write to `stdout` and log files)

JayjeetAtGithub: Thanks ! @ivotron

JayjeetAtGithub:
@ivotron Can't understand why the docker build is failing. Please see

ivotron:
Sorry @JayjeetAtGithub , can you please rebase master?

JayjeetAtGithub:
@ivotron Rebased it

ivotron:
thanks a lot!

Issue #504, opened 2019-03-04
cli: validate .workflow file


ivotron:
As part of `popper run`, add a routine that validates a `.workflow` file in order to check that the expected elements of the workflow are present, as described in the Github actions documentation for both [Workflow](https://developer.github.com/actions/creating-workflows/workflow-configuration-options/#workflow-blocks) and [Action](https://developer.github.com/actions/creating-workflows/workflow-configuration-options/#action-blocks) blocks.

The idea is to show better messages to a user in the command line as opposed to the current behavior which is a Python dump of the trace. For example, for a `.workflow` file that is missing a `resolves` keyword in the Workflow block, this is the current output:

```
$ popper run
Traceback (most recent call last):
  File "/Users/ivo/src/popper/cli/bin/popper", line 30, in <module>
    popper.cli.cli()
  File "/Users/ivo/src/popper/cli/bin/vendor/click/click/core.py", line 731, in __call__
    return self.main(*args, **kwargs)
  File "/Users/ivo/src/popper/cli/bin/vendor/click/click/core.py", line 706, in main
    rv = self.invoke(ctx)
  File "/Users/ivo/src/popper/cli/bin/vendor/click/click/core.py", line 1080, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/ivo/src/popper/cli/bin/vendor/click/click/core.py", line 907, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/ivo/src/popper/cli/bin/vendor/click/click/core.py", line 544, in invoke
    return callback(*args, **kwargs)
  File "/Users/ivo/src/popper/cli/bin/vendor/click/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/Users/ivo/src/popper/cli/bin/vendor/click/click/core.py", line 544, in invoke
    return callback(*args, **kwargs)
  File "/Users/ivo/src/popper/cli/popper/commands/cmd_run.py", line 35, in cli
    pipeline = Workflow(wfile, workspace)
  File "/Users/ivo/src/popper/cli/popper/gha.py", line 37, in __init__
    self.normalize()
  File "/Users/ivo/src/popper/cli/popper/gha.py", line 77, in normalize
    self.wf['resolves'] = wf_block['resolves']
KeyError: 'resolves'
```

Instead, we'd like something more friendly, like an `Expecting 'resolves' keyword` message.

[...]

PR #529, opened 2019-03-06
Fixes validation of .workflow

JayjeetAtGithub:
Adds validation rules for `.workflow` files


cli/popper/gha.py
```diff
@@ -53,6 +55,61 @@ def __init__(self, wfile, workspace):
...
+        """
+        Validates the .workflow file
+        """
+        if not self.wf.get('workflow', None):
+            pu.fail('Workflow block must be present\n')
+        for _, wf_block in dict(self.wf['workflow']).items():
+            if not wf_block.get('resolves', None):
+                pu.fail('[resolves] attribute must be present\n')
+            else:
+                if not isinstance(
```
ivotron: It might be cleaner to do this check in the `normalize()` function, since we are doing type checking there already

cli/popper/gha.py
```diff
@@ -53,6 +55,61 @@ def __init__(self, wfile, workspace):
...
+                        wf_block['resolves'],
+                        str) and not isinstance(
+                        wf_block['resolves'],
+                        list):
+                    pu.fail('[resolves] attribute must be string or list\n')
+            if wf_block.get('on', None):
+                if not isinstance(wf_block['on'], str):
+                    pu.fail('[on] attribute must be a string\n')
+        if not self.wf.get('action', None):
+            pu.fail('Action block must be present\n')
```
ivotron: "At least one action block must be present"

cli/popper/gha.py
```diff
@@ -53,6 +55,61 @@ def __init__(self, wfile, workspace):
...
+        for a_name, a_block in self.wf['action'].items():
+            if not a_block.get('uses', None):
+                pu.fail('[uses] attribute must be present\n')
+            else:
```
ivotron: same as above, let's move this to `normalize()`

cli/popper/gha.py
```diff
@@ -53,6 +55,61 @@ def __init__(self, wfile, workspace):
...
+                if not isinstance(a_block['uses'], str):
+                    pu.fail('[uses] block must be a string\n')
+            if a_block.get('needs', None):
```
ivotron: same as above: move to `normalize()`

cli/popper/gha.py
```diff
@@ -53,6 +55,61 @@ def __init__(self, wfile, workspace):
...
+                if not isinstance(
+                        a_block['needs'],
+                        str) and not isinstance(
+                        a_block['needs'],
+                        list):
+                    pu.fail('[needs] attribute must be a list or string\n')
+            if a_block.get('runs', None):
+                if not isinstance(
```
ivotron: move to `normalize()`

cli/popper/gha.py
```diff
@@ -53,6 +55,61 @@ def __init__(self, wfile, workspace):
...
+                        a_block['runs'],
+                        str) and not isinstance(
+                        a_block['runs'],
+                        list):
+                    pu.fail('[runs] attribute must be a list or string\n')
+            if a_block.get('args', None):
+                if not isinstance(
```
ivotron: move to `normalize()`

JayjeetAtGithub:
@ivotron Can you please tell me what logic will be in the `validate` function if the validation is handled in `normalize` function ?

cli/popper/gha.py
JayjeetAtGithub: This has been fixed

ivotron:
hey @JayjeetAtGithub 

> As of now , before normalizing , all the validation errors are being catched in validate() and only if everthing is valid , we move on to normalizing in normalize()

you're right. What I meant in my review was that, we wouldn't be doing any type-checking as part of the `validate()` method that you're implementing. Instead we would do type-checking in the `normalize` function. Maybe we can rename it to `normalize_types()` or something like that. And your method could be `validate_syntax` or `parse`?

ivotron:
hey @JayjeetAtGithub . Is there anything you would like to discuss regarding this? Thanks!

JayjeetAtGithub:
@ivotron Can you please tell what will be in there in `validate_syntax()` , since what my pr adds is only type checking , and if i do that in  `normalize_types()` , then i cant understand what `validate_syntax() / parse()` will do ? Do i need to add some other logic ?

ivotron:
@JayjeetAtGithub `validate_syntax` will only check for the presence of required elements, as specified in #504 . The `normalize_types` will look at types

ivotron:
Thanks a lot @JayjeetAtGithub !  There are a couple of pep8 issues that pep8speaks is complaining about. Also, could you please rebase master again? Sorry 😩

ivotron:
thanks a lot for your patience @JayjeetAtGithub !

Issue #513, opened 2019-03-04
cli: show ascii-art diagram of workflow


ivotron:
Add a `popper view` command that prints a diagram of workflow to the terminal

akashrajr1:
@ivotron For this [workflow](https://github.com/cplee/github-actions-demo/blob/master/.github/main.workflow), the output should be something like this?  
```
install  
   |  
   |______________  
   |             |  
  test		lint  
   |_____________|  
   |   
   |  
branch-filter  
   |  
   |  
   |  
deploy  
```

ivotron:
Yeah, that's the idea. There might be a ascii diagraming library out there that we can use. Or maybe we could first export the diagram to the .dot format and then use a graphviz library to display it

akashrajr1:
I was having a look through Graph::Easy. But I don't know how effectively it will handle the cycles in the graph. How complex can these workflow graphs be?

ivotron:
this is an alternative for python https://github.com/junjihashimoto/py-stacked-dag

PR #530, opened 2019-03-09
Add command - popper dot

akashrajr1:
Resolve issure #513. `popper view`  now displays ascii-art diag for a workflow.

Example: This [workflow](https://github.com/cplee/github-actions-demo/blob/master/.github/main.workflow) is represented as 

```Ascii for test-and-deploy workflow
             +---------------+
             |    deploy     |
             +---------------+
               |
               |
               v
+------+     +---------------+
| test | <-- | branch_filter |
+------+     +---------------+
  |            |
  |            |
  |            v
  |          +---------------+
  |          |     lint      |
  |          +---------------+
  |            |
  |            |
  |            v
  |          +---------------+
  +--------> |    install    |
             +---------------+
```

### DIFF
--- a/ci/test/dot
+++ b/ci/test/dot
@@ -5,6 +5,16 @@ source ./common
 init_test_repo
 cd $test_repo_path
 
+#####
+# Logic
+# The actual output is compared to expected output
+# The actual output is presorted
+# The expected output is sorted
+# Sorting is done as output is sometimes unpredictable
+#####
+
+# From https://raw.githubusercontent.com/cplee/github-actions-demo/master/.github/main.workflow
+
 cat <<EOF > main.workflow
 workflow "test-and-deploy" {
   on = "push"
@@ -47,8 +57,140 @@ EOF
 popper dot > out
 actual_output=`tail -n+2 out | head -n-1 | cut -f2 | sort`
 
-expected_output="test -> install;
+expected_output="branch_filter -> lint;
 branch_filter -> test;
-lint -> install;
 deploy -> branch_filter;
-branch_filter -> lint;"
+lint -> install;
+test -> install;"
+
+if [ "$actual_output" = "$expected_output" ]
+then
+  echo "success" >> success
+fi
+
+
+# From https://raw.githubusercontent.com/actions/example-aws/master/.github/main.workflow
+
+cat <<EOF > aws.workflow
+workflow "Build and Deploy" {
+  on = "push"
+  resolves = ["List Public IP"]
+}
+
+# Build
+
+action "Build Docker image" {
+  uses = "actions/docker/cli@master"
+  args = ["build", "-t", "aws-example", "."]
+}
+
+# Deploy Filter
+action "Deploy branch filter" {
+  needs = ["Push image to ECR"]
+  uses = "actions/bin/filter@master"
+  args = "branch master"
+}
+
+# AWS
+
+action "Login to ECR" {
+  uses = "actions/aws/cli@master"
+  secrets = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]
+  env = {
+    AWS_DEFAULT_REGION = "us-west-2"
+  }
+  args = "ecr get-login --no-include-email --region $AWS_DEFAULT_REGION | sh"
+}
+
+action "Tag image for ECR" {
+  needs = ["Build Docker image"]
+  uses = "actions/docker/tag@master"
+  env = {
+    CONTAINER_REGISTRY_PATH = "377117578606.dkr.ecr.us-west-2.amazonaws.com"
+    IMAGE_NAME = "aws-example"
+  }
+  args = ["$IMAGE_NAME", "$CONTAINER_REGISTRY_PATH/$IMAGE_NAME"]
+}
+
+action "Push image to ECR" {
+  needs = ["Login to ECR", "Tag image for ECR"]
+  uses = "actions/docker/cli@master"
+  env = {
+    CONTAINER_REGISTRY_PATH = "377117578606.dkr.ecr.us-west-2.amazonaws.com"
+    IMAGE_NAME = "aws-example"
+  }
+  args = ["push", "$CONTAINER_REGISTRY_PATH/$IMAGE_NAME"]
+}
+
+action "Store Kube Credentials" {
+  needs = ["Push image to ECR"]
+  uses = "actions/aws/kubectl@master"
+  secrets = ["KUBE_CONFIG_DATA"]
+}
+
+action "Configure Kube Credentials" {
+  needs = ["Login to ECR"]
+  uses = "actions/aws/cli@master"
+  env = {
+    CLUSTER_NAME = "devel2"
+    AWS_DEFAULT_REGION = "us-west-2"
+  }
+  secrets = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]
+  args = "eks update-kubeconfig --name $CLUSTER_NAME --region $AWS_DEFAULT_REGION"
+}
+
+# Example Local Action to use `aws-iam-authenticator`
+action "Deploy to EKS" {
+  needs = ["Store Kube Credentials", "Deploy branch filter"]
+  # ["Configure Kube Credentials"]
+  uses = "./.github/actions/eks-kubectl"
+  runs = "sh -l -c"
+  args = ["SHORT_REF=$(echo $GITHUB_SHA | head -c7) && cat $GITHUB_WORKSPACE/config.yml | sed 's/TAG/'\"$SHORT_REF\"'/' | kubectl apply -f - "]
+  secrets = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]
+}
+
+action "Verify EKS deployment" {
+  needs = [
+    "Push image to ECR",
+    "Deploy to EKS",
+  ]
+  # :point_down: use this for self-contained kubectl config credentials
+  #uses = "docker://gcr.io/cloud-builders/kubectl"
+  uses = "./.github/actions/eks-kubectl"
+  args = ["rollout status deployment/aws-example-octodex"]
+  secrets = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]
+}
+
+action "List Public IP" {
+  needs = "Verify EKS deployment"
+  uses = "./.github/actions/eks-kubectl"
+  args = ["get services -o wide"]
+  secrets = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]
+}
+EOF
+
+popper dot --wfile aws.workflow > out
+actual_output=`tail -n+2 out | head -n-1 | cut -f2 | sort`
+
+expected_output="Deploy_branch_filter -> Push_image_to_ECR;
+Deploy_to_EKS -> Deploy_branch_filter;
+Deploy_to_EKS -> Store_Kube_Credentials;
+List_Public_IP -> Verify_EKS_deployment;
+Push_image_to_ECR -> Login_to_ECR;
+Push_image_to_ECR -> Tag_image_for_ECR;
+Store_Kube_Credentials -> Push_image_to_ECR;
+Tag_image_for_ECR -> Build_Docker_image;
+Verify_EKS_deployment -> Deploy_to_EKS;
+Verify_EKS_deployment -> List_Public_IP;
+Verify_EKS_deployment -> Push_image_to_ECR;"
+
+if [ "$actual_output" = "$expected_output" ]
+then
+  echo "success" >> success
+fi
+
+
+counter=`cat success | wc -l`
+
+echo "Ran $counter/2 cases successfully"
+test $counter -eq 2
--- a/ci/test/dot
+++ b/ci/test/dot
@@ -13,46 +13,10 @@ cd $test_repo_path
 # Sorting is done as output is sometimes unpredictable
 #####
 
-# From https://raw.githubusercontent.com/cplee/github-actions-demo/master/.github/main.workflow
 
-cat <<EOF > main.workflow
-workflow "test-and-deploy" {
-  on = "push"
-  resolves = "deploy"
-}
-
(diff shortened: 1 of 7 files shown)


akashrajr1:
@ivotron the dependencies for perl libraries. I don't know how to resolve them
```
sudo apt install cpanminus
sudo cpanm Graph::Easy
```

cli/popper/commands/cmd_view.py
```diff
@@ -0,0 +1,66 @@
...
+    if not isinstance(cur_action, str):
+        cur_action = cur_action[0]
+
+    graph = add(parent_action, cur_action, wf["action"], graph) + "}"
+
+    with open('ascii.dot', 'w') as fp:
+        fp.write(graph)
+
+    os.system("awk '!seen[$0]++' ascii.dot > out")
+    os.system('cat out | graph-easy --from=dot --as_ascii')
```
ivotron: hey @akashrajr1 , thank you so much for working on this! The goal is to have popper rely only on python modules. So what we can do is to leave this out and put it in the documentation. If we do this, then we maybe should rename this command to `popper dot` and print the `.dot` graph to `stdout`.

Another alternative would be to find a Python library that prints to ascii a `.dot` graph. Or maybe we can do both (a `popper dot` and a python-based `popper view`)?

akashrajr1: I actually, went by your original suggestion and tried to use this [python module](https://github.com/junjihashimoto/py-stacked-dag), but I was having problems running it. I found that `__init__.py` was corrupt for one of the directories, and I had to recreate it to make it work. So it brings in the question of usability. (I will have one more look, in case I missed something)
So I was wondering if we could use perl libraries instead , but no worries. 

Doing both commands sounds good to me. But for now, we can just print the `.dot` file to `stdout` using `popper dot` for now, until a solution presents itself for `popper view`. We can add a recommendation in the documentation to use 
```
apt install cpanminus
cpanm Graph::Easy
```

akashrajr1: @ivotron maybe when L39 fails for `graph-easy` we can prompt instruction to install it?

ivotron: The problem with instructions is that we don't know what the user is using. Are users on a mac, ubuntu, centos, arch, etc?

akashrajr1: @ivotron please review. Updated as requested

akashrajr1:
@ivotron please review. Updated as requested

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,66 @@
...
+
+import click
+import hcl
+import popper.utils as pu
+from ..cli import pass_context
+import types
+
+
+@click.argument(
+    'wfile', required=True)
```
ivotron: Can we make this mimic the behavior of `popper run`, where we look for `main.workflow` in the `$PWD`, or in `./github`. We might be able to move this functionality to `utils.py` so that `run` and `dot` can share it.

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,66 @@
...
+                                 '[Used for graphical representations]')
+@pass_context
+def cli(ctx, wfile):
+    """Creates a dot file"""
+    if not os.path.isfile(wfile):
+        pu.fail("File {} not found.\n".format(wfile))
+        exit(1)
+
+    with open(wfile, 'r') as fp:
+        wf = hcl.load(fp)
```
ivotron: I think we are OK if we print to `stdout` instead to a file. This makes it compatible with Unix conventions, and if users want to store it in a file they can do `popper dot > foo.dot`. Also, printing to `stdout` allows users to pipe and do things like `popper dot | dot -png -o wf.png`

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if not isinstance(cur_action, str):
+        cur_action = cur_action[0]
+
+    graph = add(parent_action, cur_action, wf["action"], graph) + "}\n"
+
+    with open('ascii.dot', 'w') as fp:
+        fp.write(graph)
+
+    os.system("awk '!seen[$0]++' ascii.dot > out")
```
ivotron: can you please explain why we need to do this? Is this due to the fact that one more than one action can depend on the same action? If that's the case, then maybe we can do this in python. This help to remove the dependency with `awk`, which in fact is also present [here](https://github.com/systemslab/popper/blob/master/cli/popper/scm.py#L61).

cli/popper/commands/cmd_dot.py
akashrajr1: Yeah, sounds good to me. Should I also move `-- recursive` functionality to `utils.py`, so `popper dot` can also use it?

cli/popper/commands/cmd_dot.py
akashrajr1: Yes you are right, L40 was to curb duplicate statements because duplicate edges were generated due to multiple actions depending on the same action. Though it seems like as a part of fix, I forgot to integrate it properly. Thanks for pointing it out. 
I think I can remove the generation of duplicate edges using a visited array to keep track of actions visited. It will also remove the dependency on `awk`. 
( I thought that `awk` was fine as it was being used at other parts of code. I will keep the code pure python from now on)

cli/popper/commands/cmd_dot.py
akashrajr1: Agreed. Will fix it.

cli/popper/commands/cmd_dot.py
akashrajr1: @ivotron let me know about moving the `--recursive` feature to `utils` so that both `run` and `dot` can use it

akashrajr1:
@ivotron please have a look. thanks!

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,63 @@
...
+        '[default: ./github/main.workflow OR ./main.workflow]'
+    ),
+    required=False,
+    default=None
+)
+@click.command('dot', short_help='Generates a dot file '
+                                 '[Used for graphical representations]')
+@pass_context
+def cli(ctx, wfile):
+    """Creates a dot file"""
```
ivotron: this description needs to be updated,

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,63 @@
...
+@click.option(
+    '--wfile',
+    help=(
+        'File containing the definition of the workflow. '
+        '[default: ./github/main.workflow OR ./main.workflow]'
+    ),
+    required=False,
+    default=None
+)
+@click.command('dot', short_help='Generates a dot file '
```
ivotron: the description needs to be updated

cli/popper/utils.py
```diff
@@ -258,3 +258,28 @@ def get_git_files():
     """
     gitfiles, _ = exec_cmd("git ls-files")
     return gitfiles.split("\n")
+
+
+def find_default_wfile(wfile):
```
ivotron: awesome 😄

cli/popper/commands/cmd_dot.py
ivotron: @akashrajr1 yeah, that would be useful as well!

akashrajr1: Thanks, will update it soon

akashrajr1:
thanks @ivotron, will incorporate these changes!

cli/popper/commands/cmd_dot.py
akashrajr1: Moved recursive feature to `find_recursive_wfile()` the in utils. It will now return a list of wfile locations, which can be iterated thrgh.

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,74 @@
...
+    help=(
+        'File containing the definition of the workflow. '
+        '[default: ./github/main.workflow OR ./main.workflow]'
+    ),
+    required=False,
+    default=None
+)
+@click.option(
+    '--recursive',
+    help='Run any .workflow file found recursively from current path. ',
```
ivotron: this description also needs to be updated since `popper dot` is not actually running the workflows

cli/popper/commands/cmd_dot.py
ivotron: this still holds

cli/popper/commands/cmd_dot.py
ivotron: still holds

cli/popper/commands/cmd_dot.py
akashrajr1: Thanks! I missed this

akashrajr1:
@ivotron python2 [build](https://travis-ci.org/systemslab/popper/jobs/509201342#L743-L765) is failing but python3 [build](https://travis-ci.org/systemslab/popper/jobs/509201343#L451-L453) is successful. Any idea what might be the problem?

ivotron:
@akashrajr1 I suspect this is due to the python 2/3 incompatibility when dealing with strings. One way of addressing this is to implement this functionality as part of the `Workflow` class. The `Workflow` constructor does a couple of things like validating the `.workflow` file syntax and normalizing the representation. So you might want to create a `Workflow.to_dot()` method that implements this logic, as opposed to implementing it on the `cmd_dot` script

akashrajr1:
Sure @ivotron, I will give that a try

akashrajr1:
@ivotron please review the PR! Thanks

ci/test/dot
```diff
@@ -0,0 +1,85 @@
...
+export PHONY_SECRET=foo
+
+popper dot > out
+actual_output=`tail -n+2 out | head -n-1 | cut -f2 | sort`
+
+expected_output="branch_filter -> lint;
+branch_filter -> test;
+deploy -> branch_filter;
+lint -> install;
+test -> install;"
```
ivotron: deploy is the last action as shown [here](https://github.com/cplee/github-actions-demo/blob/master/.github/main.workflow)

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,71 @@
...
+    default=None
+)
+@click.option(
+    '--recursive',
+    help='Generate .dot file for any.workflow file '
+         'found recursively from current path.',
+    required=False,
+    is_flag=True
+)
+@click.command('dot', short_help='Generates a .dot format for a given workflow'
```
ivotron: "Generate a graph in the .dot format"

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,71 @@
...
+               '\n[.dot is a format for graphical representation]')
```
ivotron: the square brackets are used to describe default options. I'd leave this line out.

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,71 @@
...
+@pass_context
+def cli(ctx, wfile, recursive):
+    """
+    Analyzes a workflow file and then creates a dot representation for the same
```
ivotron: I'd leave the "analyzes" out. Maybe "Creates a graph in the .dot format representing the workflow"

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,71 @@
...
+    .dot represents the order in which actions in the workflow must be executed
```
ivotron: this might be left out

cli/popper/commands/cmd_dot.py
```diff
@@ -0,0 +1,71 @@
...
+        parent_action = cur_action = action
+
+        graph = list()
+        graph = add(parent_action, cur_action, wf['action'], graph)
+        graph = ''.join(list(set(graph)))
+        graph = "digraph G {\n" + graph + "}\n"
+        pu.info(graph)
+
+
+# Recursively go through "needs" and add corresponding actions to graph
```
ivotron: seems like the logic is inverted and this ends up generating an inverted representation of the graph.

One alternative is to use the `next` attribute of each node, which represent forward-going edges ([this block of code](https://github.com/systemslab/popper/blob/master/cli/popper/gha.py#L168-L171) adds these). So you can start from the root and then use `next` to get to the subsequent nodes.

ci/test/dot
akashrajr1: The `expected_output` is actual a list of sorted edges. Sorting is needed as graph nodes on the same level may be executed in arbitrary order. Eg. `lint -> install` and `test -> install` were appearing out of order. So sort was used to rectify this, so that comparison is done properly.
But I think `expected_output` is misleading, will rename it to something else.

cli/popper/commands/cmd_dot.py
akashrajr1: Yes, I had to invert the order to get the output below by using `popper dot | graph-easy`. This is I think because of how directed graphs are. Should I change it to an  undirected graph?

> Resolve issure #513. `popper view` now displays ascii-art diag for a workflow.
> 
> Example: This [workflow](https://github.com/cplee/github-actions-demo/blob/master/.github/main.workflow) is represented as
> 
> ```
>              +---------------+
>              |    deploy     |
>              +---------------+
>                |
>                |
>                v
> +------+     +---------------+
> | test | <-- | branch_filter |
> +------+     +---------------+
>   |            |
>   |            |
>   |            v
>   |          +---------------+
>   |          |     lint      |
>   |          +---------------+
>   |            |
>   |            |
>   |            v
>   |          +---------------+
>   +--------> |    install    |
>              +---------------+
> ```

ivotron: in the diagram you linked to, the order of actions is inverted. The graph should look like this:

```
install -> test -> branch-filter -> deploy
           -> lint 
```

the first action that executes is the install action, followed by test and lint, then branch-filter and then deploy. In the diagram you posted above, deploy is the first action that gets executed

akashrajr1: Oh.. I will change it to reflect that

akashrajr1:
@ivotron this PR is ready for review

ivotron:
thanks a lot for your patience @akashrajr1 !

PR #524, opened 2019-03-06
fix python 2 to 3 compatibility

barbaraDorame:
fixes #521

PR #532, opened 2019-03-09
Adds parallel execution of actions in stages

JayjeetAtGithub:
Runs the actions in a stage in parallel


JayjeetAtGithub:
@ivotron Can you please help me on how to add the `futures` dependency to the codebase. I added it to `setup.py` first , but build failed. Then i tried sometime more, but still failing.

JayjeetAtGithub:
```
$ bash -c "popper run --wfile ci/cli.workflow"
Error: Command 'run' doesn't exist. 
Type 'popper --help' for more.
The most similar commands are: run, version
No module named concurrent.futures
The command "bash -c "popper run --wfile ci/cli.workflow"" exited with 2.
```

This error comes during python2.7 build while python3.7 passes

ivotron:
sorry @JayjeetAtGithub can you please rebase master?

JayjeetAtGithub:
@ivotron  I have rebased, but my python2.7 build is failing ? Can you please help a bit

.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "cli/bin/vendor/python-future"]
 	path = cli/bin/vendor/python-future
 	url = https://github.com/PythonCharmers/python-future.git
+[submodule "cli/bin/vendor/pythonfutures"]
+	path = cli/bin/vendor/pythonfutures
+	url = https://github.com/agronholm/pythonfutures.git
```
ivotron: the `futures` module is already being added, so there's no need to add it again. This was done as part of #524

cli/bin/popper
```diff
@@ -16,6 +16,7 @@ else:
 sys.path.insert(0, yaml_lib)
 if sys.version_info[0] < 3:
     sys.path.insert(0, os.path.join(vendordir, 'python-future', 'src'))
+    sys.path.insert(0, os.path.join(vendordir, 'pythonfutures', 'src'))
```
ivotron: same as above

cli/setup.py
```diff
@@ -20,7 +20,8 @@
         'lark-parser==0.5.6',
         'pyhcl',
         'pyyaml',
-        'pyaes'
+        'pyaes',
+        'futures; python_version == "2.7"'
```
ivotron: same as above

JayjeetAtGithub:
@ivotron The module that i added as a part of this pr is [futures](https://pypi.org/project/futures/) which is different from that added in #524.I removed it, but build is still failing.

cli/bin/popper
ivotron: @JayjeetAtGithub this does not work because the `pythonfeatures` repo does not have a `src` folder. The library is at the root of the project

ivotron:
@JayjeetAtGithub oh, I see. In that case, what you had before should work. I left a comment in one of our threads.

JayjeetAtGithub:
@ivotron Thanks a lot for pointing it out. Updated the PR

ivotron:
I hate to ask this again: could you please rebase master again? Sorry 😩

JayjeetAtGithub:
@ivotron Added --serial flag

JayjeetAtGithub:
Thanks ! Will be updating soon

JayjeetAtGithub:
@ivotron Added tests to check thread safety with stages each containing 10 actions.

ivotron:
@JayjeetAtGithub thanks a lot for addressing the previous comments!

The output on TravisCI seems to indicate that only the test for `reuse` is being executed, can you please check? One thing I suspect it will be an issue is the printing of messages to the terminal. We might need to serialize access to stdout within the `utils.exec_cmd` function. In general, we need to check all the codebase to ensure that there are no thread-safety issues. Another one from the top of my head is the `Workflow.get_stages` generator method, which will likely require the use of thread-safe iterators.

Also, we will need to run the tests on travis using the `--serial` flag (i.e. we need to modify the `.travis.yml` so that this flag is passed), otherwise all the tests we run will be implicitly testing this functionality as well. Another alternative is to invert the behavior and add a `--parallel` flag and always run serially by default.

JayjeetAtGithub:
> One thing I suspect it will be an issue is the printing of messages to the terminal. We might need to serialize access to stdout within the utils.exec_cmd function. 

@ivotron Can you explain this a bit ?

JayjeetAtGithub:
> The output on TravisCI seems to indicate that only the test for reuse is being executed, can you please check?

Yes , only reuse is getting executed . Sorry for not noticing it before.

ivotron:
> > One thing I suspect it will be an issue is the printing of messages to the terminal. We might need to serialize access to stdout within the utils.exec_cmd function.
> 
> @ivotron Can you explain this a bit ?

sure! Currently, [this function](https://github.com/systemslab/popper/blob/1f04249de147b7f46517531f070f7ea33827dec3/cli/popper/utils.py#L89) makes reference to `stdout` via the `info()` function. When `popper run` is executed (with `verbose=True` and `log_file=True`), multiple threads executing this function will be printing lines of output that can correspond to multiple threads. All the output will likely end up being interleaved.

One pragmatic solution is to notify users that this is likely to happen when they execute in parallel mode, and that we advice them to use `--quiet` to avoid confusion.

JayjeetAtGithub:
@ivotron  Summary of changes made:

- Changed `--serial` to `--parallel` and changed the tests accordingly .
- Made the generator function `Workflow.get_stages` thread safe by making a custom decorator `@threadsafe_generator` which is defined in `utils.py` file.
- The above change resulted in execution of all tests  and proper output in build logs which was not happening before.
- Added a warning message about interleaved output on `popper run --parallel`.

JayjeetAtGithub:
@ivotron I cant understand why some of these actions are failing silently. Can you please tell me ?

JayjeetAtGithub:
Hi , @ivotron I have modified the `ThreadPoolExecutor` part in a way, so that if any one of the actions in a stage ( where all the actions are running parallely ) do fail, then the execution stops with `exit code 1` and hence the build fails. I have also corrected the tests, previously due to which all the `[ls...]` actions failed.

JayjeetAtGithub:
@ivotron Ok !  Updating soon !

JayjeetAtGithub:
@ivotron Made the changes accordingly. Added tests to check whether all the required `.log` files are being generated in `popper_logs` directory in the workspace for every action.

JayjeetAtGithub:
Hi @ivotron, Changes made :
- Fixed PEP8 issues
- Modified the `run_stage()` fn. by adding `pu.info('Action ran successfully !')` whenever an action runs successfully in parallel mode. Then i added tests to match the count of this info statement in output with the total no. of actions. Hope this serves the purpose. :smiley:

JayjeetAtGithub:
@ivotron Added failure tests

cli/popper/gha.py
```diff
@@ -290,12 +292,30 @@ def run(self, action_name=None, reuse=False):
...
+
+    def stage_exec(self, action, reuse):
+        self.wf['action'][action]['runner'].run(reuse)
+
+    def run_stage(self, stage, reuse=False, parallel=False):
+        if parallel:
+            with ThreadPoolExecutor(max_workers=mp.cpu_count()) as ex:
+                flist = {
+                    ex.submit(
+                        self.stage_exec,
```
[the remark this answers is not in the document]
JayjeetAtGithub: Done

cli/popper/gha.py
```diff
@@ -290,12 +292,30 @@ def run(self, action_name=None, reuse=False):
...
+                        a,
+                        reuse): a for a in stage}
```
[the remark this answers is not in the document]
JayjeetAtGithub: Done

cli/setup.py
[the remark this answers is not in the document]
JayjeetAtGithub: Done

ivotron:

Issue #505, opened 2019-03-04
cli: add support for handling ctrl-c signal properly


ivotron:
As of today, if users run [`popper run`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_run.py) and they kill it by pressing `ctrl-c`, any initiated containers will remain running as only the Python app is interrupted. For this issue, we will catch the `ctrl-c` event and cleanup all subprocess  and running containers. For this to be done, we need to either name containers deterministically or keep a list of container IDs that were initiated by popper

akashrajr1:
Essentially we keep a list of subprocesses and docker CIDs stored during execution, and once we catch the interrupt we kill/delete all subprocesses and docker CIDS?

ivotron:
@akashrajr1 yes, that would work

PR #537, opened 2019-03-14
Cleanup on Ctrl+C Signal

akashrajr1:
Issue #505: Cleanup docker containers created during `popper run`

Output generated on running `popper run` on this [workflow](https://github.com/cplee/github-actions-demo/blob/master/.github/main.workflow)
```
Already on 'master'
[install] docker pull node:11.6.0
[install] docker create node:11.6.0 npm install
[install] docker start ..
[test] docker pull node:11.6.0
[test] docker create node:11.6.0 npm test
[test] docker start .
[lint] docker build -t action/jshint /root/github-actions-demo/./.github/actions/jshint
[lint] docker create action/jshint 
[lint] docker start .
[branch-filter] docker build -t actions/bin /tmp/actions/actions/bin/./filter
[branch-filter] docker create actions/bin branch master
[branch-filter] docker start .
[deploy] docker build -t actions/bin /tmp/actions/actions/bin/./sh
^C

Deleted install
Deleted test
Deleted lint
Deleted branch-filter
Terminated
```

### DIFF
--- a/ci/test/interrupt
+++ b/ci/test/interrupt
@@ -20,30 +20,8 @@ kill -2 $pid
 sleep 5
 check="deploy branch-filter test lint install"
 actual=`docker ps -a --format "{{.Names}}"`
-for a in $actual
-do 
-  for b in $check
-  do 
-    if [ $a = $b ] 
-    then echo "fail" >> fail
-    fi
-  done
-done
 
-counter=`cat fail|wc -l`
-
-#A test for --parallel run
-popper run & pid=$!
-#Give popper run some time to create dockers
-sleep 8
-
-# -2 is for SIGINT or Ctrl-C
-kill -2 $pid
-
-#Give popper run some time to remove dockers and then check
-sleep 5
-
-actual=`docker ps -a --format "{{.Names}}"`
+cd ..
 for a in $actual
 do
   for b in $check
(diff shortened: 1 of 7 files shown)


ivotron:
thanks a lot @akashrajr1! Would it be possible to rebase master 😩 sorry, it seems that my latest merge will cause multiple conflicts with your PR

akashrajr1:
@ivotron not sure how to proceed from here onwards. Should I wait for #532 ( parallel actions) to be over?

ivotron:
@akashrajr1 thanks a lot for working on this! IMO this issue can be worked independent, if we do the following:

  * ensuring that the signal handling mechanism works in a multi-threaded scenario. For example, that `signal_handler` runs in a single thread, and that when it does so, all the running threads are interrupted. This might be the way this already works, but we need to double-check.
  * adding tests for this feature. This can be done [in bash](https://stackoverflow.com/q/14696427).

akashrajr1:
The build fail in python 3 [here](https://travis-ci.org/systemslab/popper/jobs/510687357#L294-L298) is addressed in this PR #556

ivotron:
thanks a lot @akashrajr1! As a way of double-checking all is fine, can you please rebase master? thanks!

akashrajr1:
@ivotron success on first try :)
Please review it, thanks!

cli/popper/cli.py
```diff
@@ -67,3 +69,30 @@ def get_command(self, ctx, name):
...
+def signal_handler(sig, frame):
+
+    pu.info("\n")
+    cmd_out = pu.exec_cmd('docker ps -a --format "{{.Names}}"')
+    cmd_out = cmd_out[0].splitlines()
+    cmd_out = set(cmd_out)
+
+    for img in set(docker_list).intersection(cmd_out):
+        pu.exec_cmd('docker stop {}'.format(img))
+        pu.exec_cmd('docker rm -f {}'.format(img))
```
ivotron: users might have provided the `--reuse` flag to `popper run`, so we need to take that into account.

cli/popper/utils.py
```diff
@@ -187,11 +188,13 @@ def b(t):
                 info('DEBUG: subprocess.Popen() with combined stdout/stderr\n')
             p = Popen(cmd, stdout=PIPE, stderr=STDOUT, shell=True,
                       universal_newlines=True)
+            popper.cli.process_list.append(p.pid)
         else:
             if debug:
                 info('DEBUG: subprocess.Popen() with separate stdout/stderr\n')
             p = Popen(cmd, stdout=outf, stderr=errf, shell=True,
                       universal_newlines=True)
+            popper.cli.process_list.append(p.pid)
```
ivotron: since this is done regardless of the value of `verbose`, it can be done outside the conditional block

ci/test/interrupt
```diff
@@ -0,0 +1,35 @@
...
+  for b in $check
+  do 
+    if [ $a = $b ] 
+    then echo "fail" >> fail
+    fi
+  done
+done
+
+counter=`cat fail|wc -l`
+test $counter -eq 0
```
ivotron: would it be possible to add a test with `popper run --parallel`? We can take one of the workflows for those tests and make the actions run for longer, and then check if the signal handler properly cleans up all the concurrently running containers.

cli/popper/cli.py
ivotron: Something else I forgot to mention is that, this also affects the message presented to the user. When `--reuse` is given, we are only stopping containers; when it isn't, we are destroying them as well.

We can also print a message like `^C signal detected, stopping running processes"

akashrajr1: Oh right. I will take that into account

ci/test/interrupt
akashrajr1: Yeah, sure. Will try it out

akashrajr1:
@ivotron PR is ready for review! Thanks!

ci/test/interrupt
```diff
@@ -0,0 +1,58 @@
...
+    if [ $a = $b ] 
+    then echo "fail" >> fail
+    fi
+  done
+done
+
+counter=`cat fail|wc -l`
+
+#A test for --parallel run
+popper run & pid=$!
```
ivotron: this has to make use of the `--parallel` flag. In addition, the workflow being executed does not contain multiple actions in a single stage, so running with the `--parallel` flag will end up being the same as running without it. We can reuse the workflow defined in the [test for the parallel](https://github.com/systemslab/popper/blob/master/ci/test/parallel_stage_exec) feature here, so that we ensure multiple actions run on the same stage. We might need to modify the actions so that they run for longer (e.g. instead of running `ls`, we can run `sleep 100` or something alike).

akashrajr1: I think I was using `--parallel` option. But I might have overwritten it somehow, hence the miss. Will change the workflow and update the PR

akashrajr1:
@ivotron please have a look. thanks!

ivotron:
thanks a lot for your patience @akashrajr1 !

PR #544, opened 2019-03-15
Fixes bug in validate_syntax

JayjeetAtGithub:
There was a bug in the PR #529 where if a `resolves` or a `uses` attribute was found , it returned which lead to the function not validating the rest of the block and hence errors were thrown afterwards in `normalize` and `check_secrets`. #529  got merged before i could update the PR, hence this new PR.

### DIFF
--- a/cli/popper/gha.py
+++ b/cli/popper/gha.py
@@ -60,20 +60,24 @@ class Workflow(object):
     def validate_syntax(self):
         """ Validates the .workflow file.
         """
+        resolves_present = False
+        uses_present = False
         if not self.wf.get('workflow', None):
             pu.fail('A workflow block must be present\n')
         else:
             for _, wf_block in dict(self.wf['workflow']).items():
                 if wf_block.get('resolves', None):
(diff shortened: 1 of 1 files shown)


ivotron:
Interesting, can we add more tests to catch this scenario?

JayjeetAtGithub:
I guess no more tests are needed as i have included all kinds of test cases in validate test.

JayjeetAtGithub:
@ivotron [here](https://travis-ci.org/systemslab/popper/jobs/506589554#L498) is the build log where the workflow execution was failing silently.
The build log of this PR which solves the above issue is , [here](https://travis-ci.org/systemslab/popper/jobs/506605419#L312) where the validation is working successfully and a nice message is shown instead of the dirty stacktrace.

JayjeetAtGithub:
> Interesting, can we add more tests to catch this scenario?

This scenario will not occur anymore

ivotron:
thanks a lot!

Issue #536, opened 2019-03-13
ci: missing action "test recursive" in cli.workflow


akashrajr1:
@ivotron I just realised.. shouldn't there be an action `test recursive` for [cli.workflow](https://github.com/systemslab/popper/blob/master/ci/cli.workflow)?

ivotron:
oh, you're right! sorry I missed that.

akashrajr1:
@ivotron thanks for the reaffirmation. Will send a PR to fix it.

PR #545, opened 2019-03-15
Fix missing action recursive in cli.workflow

akashrajr1:
Addresses #536

### DIFF
--- a/ci/cli.workflow
+++ b/ci/cli.workflow
@@ -33,6 +33,11 @@ action "test scaffold" {
   runs = "scaffold"
 }
 
+action "test recursive" {
+  uses = "./ci/test"
+  runs = "recursive"
+}
+
 action "end" {
   uses = "./ci/test"
   runs = "version"
@@ -42,6 +47,7 @@ action "end" {
     "test reuse",
     "test actions-demo",
     "test validate",
-    "test scaffold"
+    "test scaffold",
+    "test recursive"
   ]
 }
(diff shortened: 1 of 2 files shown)


akashrajr1:
@ivotron can you help me resolve `travis ci` build fail?

akashrajr1:
[This](https://travis-ci.org/systemslab/popper/jobs/507535783#L897-L902) seems to indicate only `main.workflow` was run. But [these](https://travis-ci.org/systemslab/popper/jobs/507535783#L890-L894) indicate that `popper run --recursive` has generated output for these workflows in out and thus got the output it needed
And since `set -x` wasn't working properly for nested loops I had to reset it here. Maybe just hit and trial will help?https://github.com/systemslab/popper/blob/8dbffd68c68b34e045a26ee0c2aa0082982ac11b/ci/test/recursive#L36-L48

akashrajr1:
@ivotron just found something instrumental!  
https://github.com/systemslab/popper/blob/8dbffd68c68b34e045a26ee0c2aa0082982ac11b/ci/test/recursive#L45
Arithmetic expansions actually seem to break the code for some unknown reason! So I had to bypass it by counting no of `success`
https://github.com/systemslab/popper/blob/a2877e8d407ceca4211d4f2d262af6818e9acc38/ci/test/recursive#L44-L49
All of them seem to fail. For future references, these commands fail the build
```
i=$((i+1)) 
i=$((i++)) 
((i=i+1))  
((i+=1))   
((i++))    
((++i))    
let "i=i+1"
let "i+=1" 
let "i++"  
let i=i+1  
let i+=1   
let i++    
```

ivotron:
@akashrajr1 good catch!

ivotron:
thanks a lot!

PR #548, opened 2019-03-18
Fixes #540 Implemented git cmds with GitPython library

### LINKED ISSUE #540: cli: use gitpython library
Make use of https://github.com/gitpython-developers/GitPython instead of interacting directly with the `git` command. This would result in getting rid of the [`scm.py`](https://github.com/systemslab/popper/blob/master/cli/popper/scm.py) module.

JayjeetAtGithub:
Makes use of [GitPython](https://pypi.org/project/GitPython/) package instead of directly interacting with `git`.


cli/popper/gha.py
```diff
@@ -215,8 +219,17 @@ def download_actions(self):
                 pu.info('[popper] cloning actions from repositories\n')
                 infoed = True
 
-            scm.clone(user, repo, repo_parent_dir, version, debug=self.debug)
-
+            repo_dir = os.path.join(repo_parent_dir, repo)
+            if os.path.exists(repo_dir):
+                pu.exec_cmd('rm -rf {}'.format(repo_dir))
```
ivotron: I'm not familiar with `GitPython`, do you know if it is possible to avoid re-cloning a repo if it is already cloned? This would allow us to avoid having to `rm -rf` which is always scary.

cli/popper/commands/cmd_init.py
```diff
@@ -10,7 +10,12 @@
 def cli(ctx):
     """Initializes a repository by creating the .popper.yml file.
     """
-    project_root = scm.get_root_folder()
+    try:
+        repo = git.Repo(search_parent_directories=True)
+    except git.exc.InvalidGitRepositoryError:
+        pu.fail('Must be a git repository ! \n')
+
+    project_root = repo.git.rev_parse('--show-toplevel')
```
ivotron: can we move this functionality to `scm`? In this way this can be shared with other commands that might want to do this (like `scaffold`).

cli/popper/scm.py
```diff
@@ -1,51 +1,40 @@
...
+        return infer_repo_name_from_root_folder(repo)
 
     if url.endswith('.git'):
         url = url[:-4]
 
     return os.path.basename(url)
 
 
-def get_user(debug):
+def get_user(repo):
```
ivotron: can this be done via the GitPython library?

cli/popper/gha.py
```diff
@@ -2,8 +2,9 @@
 from builtins import dict, str
 import hcl
 import os
-import popper.scm as scm
 import popper.utils as pu
+import popper.scm as scm
+import git
```
ivotron: would it be possible to hide `git` within the `scm` module? Or alternatively, completely get rid of the `scm.py` file?

JayjeetAtGithub: @ivotron Getting rid of `scm.py` file completely is not possible i guess, as some functions in it like `get_user()` and `get_name()` etc. does not have direct API's in GitPython. So, one thing we can do is that keep the `scm.py` file and replace raw git commands with GitPython API's in there.

cli/popper/gha.py
JayjeetAtGithub: @ivotron what we can do is check if the repo is present locally , we avoid cloning . Thus getting rid of `rm -rf`

ivotron: yes, hopefully that can be handled by gitpython. The main issue here is, how do we ensure that the local folder corresponds to the repo we want to clone? E.g. two repos might have the same name but they are distinct. So we might need to create a full `/tmp/actions/github/org/repo` path (as opposed to what we do know which is to put it in `/tmp/actions/repo`.

cli/popper/scm.py
JayjeetAtGithub: @ivotron I couldn't find any direct API's for this

cli/popper/commands/cmd_init.py
JayjeetAtGithub: @ivotron Ok

cli/popper/gha.py
ivotron: hey @JayjeetAtGithub, thanks a lot for working on this. Looks like my comment above was not addressed, right? No biggie if not, we can open another issue for that. I just wanted to double-check that it will not be part of this PR, so I can proceed to merge it. Thanks!

JayjeetAtGithub: @ivotron I am working on this part currently . Updating soon

JayjeetAtGithub: @ivotron is `/github/` important in the path or we can do `/tmp/actions/org/repo` ?

JayjeetAtGithub: @ivotron I guess the url can be `https://github.com` or `https://gitlab.com` etc. So if this is the case, can the path be like `/tmp/actions/gitlab.com/org/repo` ?

JayjeetAtGithub: Can this be done by comparing the remote origins of the repos? if they are same i avoid cloning else i remove the exisiting one and clone

cli/popper/scm.py
```diff
@@ -1,30 +1,35 @@
 import os
 import popper.utils as pu
+import git
+import shutil
 
+try:
+    repo = git.Repo(search_parent_directories=True)
+except git.exc.InvalidGitRepositoryError:
+    pass
```
ivotron: Ignoring this exception might cause unexpected results. Since our assumption is that popper runs on git repos, I'd vote for failing if this is catched. In this way we also can avoid doing this `Repo` instantiation again

JayjeetAtGithub:
@ivotron Please review ! I have made the changes

cli/popper/scm.py
JayjeetAtGithub: Fixed it

cli/popper/utils.py
```diff
@@ -91,12 +92,12 @@ def get_project_root():
     Returns:
         project_root (str): The fully qualified path to the root of project.
     """
-    base, _ = exec_cmd('git rev-parse --show-toplevel', ignore_error=True)
-
-    if not base:
+    try:
```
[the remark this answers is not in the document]
JayjeetAtGithub: @ivotron Removed this function

cli/popper/scm.py
```diff
@@ -51,24 +61,30 @@ def get_user(debug):
...
     """Runs git rev-parse --short HEAD and returns result"""
-    return pu.exec_cmd('git rev-parse --short HEAD', debug=debug)[0]
+    try:
+        return repo.git.rev_parse(repo.head.object.hexsha, short=True)
+    except ValueError:
+        click.secho(
+            'ERROR: Needed a single revision',
+            fg='red', bold=True, err=True, nl=False
+        )
+        sys.exit(1)
```
ivotron: we can use the `utils.fail()` method

cli/popper/scm.py
```diff
@@ -1,30 +1,40 @@
...
+import sys
+
+try:
+    repo = git.Repo(search_parent_directories=True)
+except git.exc.InvalidGitRepositoryError:
+    click.secho(
+        'ERROR: Unable to find root of project. Initialize repository first.',
+        fg='red', bold=True, err=True, nl=False
+    )
+    sys.exit(1)
```
ivotron: let's use the `utils.fail()` function instead of directly calling `click`

ivotron:
not sure why github closed [this thread](https://github.com/systemslab/popper/pull/548#discussion_r267559416) where we were discussing dealing the possibility of avoid cloning repos that already exist. This is what you asked last:

> @ivotron is `/github/` important in the path or we can do `/tmp/actions/org/repo`?

The `github/` part is important because we can clone from any other URL, not just from github. So for example, this two repos might not be containing the same project, even though they have similar `user/repo`: `github.com/foo/bar` and `gitlab.com/foo/bar`.

> Can this be done by comparing the remote origins of the repos? if they are same i avoid cloning else i remove the exisiting one and clone

yeah, that can work. We look at what their `origin` remote points to and compare based on that.

cli/popper/scm.py
JayjeetAtGithub: This is leading to circular imports

cli/popper/scm.py
JayjeetAtGithub: Same

cli/popper/scm.py
ivotron: I see. This can be resolved by having the import of `utils` within the `except` context, as explained [here](https://stackoverflow.com/a/1556429)

ivotron: I see. This can be resolved by having the import of `utils` within the `except` context, as explained [here](https://stackoverflow.com/a/1556429)

cli/popper/scm.py
JayjeetAtGithub: @ivotron  Here if i am doing 
```
try:
    repo = git.Repo(search_parent_directories=True)
except git.exc.InvalidGitRepositoryError:
    from popper.utils import fail
    fail('Unable to find root of project. Initialize repository first.')
```
When i am running `popper` in a directory where git is not initialized error comes up

ivotron: Which error? Circular dependencies?

cli/popper/utils.py
```diff
@@ -3,7 +3,8 @@
 import sys
 import time
 import yaml
-
+import git
```
ivotron: I believe we don't need this here anymore

cli/popper/utils.py
```diff
@@ -83,25 +84,9 @@ def get_items(dict_object):
...
-    if not base:
-        fail("Unable to find root of project. Initialize repository first.")
-
-    return base
-
-
 def write_config(config):
     """Writes config to .popper.yml file."""
-    config_filename = os.path.join(get_project_root(), '.popper.yml')
+    config_filename = os.path.join(scm.get_root_folder(), '.popper.yml')
```
ivotron: We can also take the folder where the config is to be written as an argument, so that we don't need to import scm

cli/popper/scm.py
JayjeetAtGithub: yes

cli/popper/utils.py
JayjeetAtGithub: yup , removing

JayjeetAtGithub:
@ivotron  Fixed all the issues. Please have a look !

cli/popper/utils.py
ivotron: @JayjeetAtGithub we can do this (see above) so that we don't need to import `scm` in `utils.py`. In this way, we don't need to import `utils` from within the `exception` block.

cli/popper/scm.py
JayjeetAtGithub: @ivotron Solved it

cli/popper/utils.py
JayjeetAtGithub: @ivotron But then how do we get root project directory in utils.py ?

ivotron: whenever the `write_config()` and `is_popperized()` functions are called. So these two functions need to receive the folder now as an argument. In this way we don't need to make import `scm` from within the `utils.py` file

JayjeetAtGithub: @ivotron Sorry for that , i understood it . Please review

cli/popper/commands/cmd_init.py
```diff
@@ -10,13 +10,13 @@
 def cli(ctx):
     """Initializes a repository by creating the .popper.yml file.
     """
-    if pu.is_popperized():
+    if pu.is_popperized(scm.get_root_folder()):
```
ivotron: there's a `project_root` variable below that can be used to avoid having to call this function multiple times

JayjeetAtGithub: @ivotron  Please have a look

cli/popper/gha.py
JayjeetAtGithub: @ivotron i feel it would be better if we can solve this `/github/org/repo` issue in another PR as this PR has already lots of changes. I would be better to have a clean PR for this. Thanks a lot !

ivotron: sure. Thanks!

ivotron:
thanks a lot for your patience @JayjeetAtGithub !

JayjeetAtGithub:
Thanks a lot ! @ivotron

PR #554, opened 2019-03-23
Add some dry-run test fixes

akashrajr1:
Address Issue #556 
Hi @ivotron, so I was actually checking the dry-run test. So I found that `dry-run` gave different outputs according to different scenarios. So I integrated those in the test.
Scenario 1: Which was occurring most frequently. It is also appeared as a build fail when I rebased #530 [here](https://travis-ci.org/systemslab/popper/jobs/510243060#L294-L303). It happens when the `HostRunner` is called instead of `DockerRunner` class.
(https://travis-ci.org/systemslab/popper/jobs/510243060#L294-L303) https://github.com/systemslab/popper/blob/f4eb8078c23b497faba6de9746984ad526f1861b/ci/test/dry-run#L49-L59
Scenario 2: It happened only once on my local. Thereafter I couldn't reproduce it, but I included it for good measure
https://github.com/systemslab/popper/blob/f4eb8078c23b497faba6de9746984ad526f1861b/ci/test/dry-run#L35-L43

### DIFF
--- a/ci/test/dry-run
+++ b/ci/test/dry-run
@@ -78,12 +78,11 @@ check_count=`cat check | wc -m`
 
 if [ $count1 -eq $check_count ]; then
 diff 1 check
-fi
-
-if [ $count2 -eq $check_count ]; then
+elif [ $count2 -eq $check_count ]; then
 diff 2 check
-fi
-
-if [ $count3 -eq $check_count ]; then
+elif [ $count3 -eq $check_count ]; then
 diff 3 check
+else
+echo ""> empty
+diff empty check  #Fail scenarrio
 fi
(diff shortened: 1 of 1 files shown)


akashrajr1:
@ivotron now I have changed the logic to look at the number of lines, before deciding which scenario it is being checked for. 
Can you let me know if there are any more scenario which might be included in this test?

JayjeetAtGithub:
@akashrajr1 Can you please check the [travis build](https://travis-ci.org/systemslab/popper/jobs/510393099#L296-L303) of #555 . Is the build failing due the issue you address through this PR ?

akashrajr1:
Yes @JayjeetAtGithub, hopefully this PR will fix the issue at hand. Sorry about any problems it may have caused.

JayjeetAtGithub:
@akashrajr1 It's fine ! I just wanted to crosscheck. Thanks a lot :+1:

ivotron:
good catch @akashrajr1 ! thanks for fixing 😄

PR #563, opened 2019-03-26
use docker-py

### LINKED ISSUE #539: cli: use docker-py library
Use the [`docker-py`](https://github.com/docker/docker-py) python library instead of interacting directly with the `docker` command.

barbaraDorame:
resolves #539


ivotron:
this seems to be afected by #568

cli/popper/gha.py
```diff
@@ -406,71 +407,78 @@ def run(self, reuse):
...
         for e, v in self.env.items():
             env_vars.update({e: v})
-
         env_vars.update({'HOME': os.environ['HOME']})
+        volumes = [self.workspace, os.environ['HOME'], '/var/run/docker.sock']
+        self.container = self.docker_client.containers.create(
+            image=img, command=self.action.get('args', None),
+            name=self.cid, volumes={v: {'bind': v} for v in volumes},
+            working_dir=self.workspace, environment=env_vars,
+            entrypoint=self.action.get('runs', None))
```
ivotron: when `self.debug` is `True`, we can show a message saying that we're about to invoke `create()`. And also another one below when we're executing `start()`

barbaraDorame:
hey ivo, the `test interrupt` action is failling on travis, but it runs locally without problems. Any idea what could be the problem?

barbaraDorame:
Ok, i fixed the error above, but now i can't run the `test dry-run` it says `+ test 15 -eq 11`, so i checked out to master and it fails this test too, same error. Any ideas?

ivotron:
thanks a lot @barbaraDorame! I made some changes to the dry run test, but looks like this is failing some time. Can you please update the test with the following?

```bash
#!/bin/bash
set -ex

source ./common
init_test_repo
cd $test_repo_path

git clone https://github.com/cplee/github-actions-demo
cd github-actions-demo
git checkout c824a5a

export PHONY_SECRET=foo

# delete any containers that might exist
docker rm $(docker ps -aq) || true

dry_num_lines=$(popper run --dry-run | grep 'DRYRUN' | wc -l)

# 5 actions, at least one line per action
test $dry_num_lines -gt 5

# test that no containers are running
test -z "$(docker ps -aq)"
```

this checks that there is at least one `DRYRUN` line for each action, as opposed to expecting a particular number per action.

barbaraDorame:
ok, i added it, thanks a lot

barbaraDorame:
geez, I thought i fixed the `test interrupt`, but it keeps failling in travis, and it works correctly in local. Any thoughts?

barbaraDorame:
i tested it on master and it fails on local :thinking:

ivotron:
@barbaraDorame we also need to modify the code that stops containers when ctrl-c is received. This is located in the [`cli.py`](https://github.com/systemslab/popper/blob/master/cli/popper/cli.py#L95) file.

In addition, some workflows that are defined in our tests are not correctly setting the `runs` attribute. This is what's happening with the `interrupt` test, like [here](https://github.com/systemslab/popper/blob/master/ci/test/interrupt#L52) script

barbaraDorame:
Done! thanks for the help >:D

cli/popper/cli.py
```diff
@@ -92,8 +92,8 @@ def signal_handler(sig, frame):
         except OSError:
             pass
 
-    for cid in docker_list:
-        pu.info("Stopping container '{}'\n".format(cid))
-        pu.exec_cmd('docker stop -t 1 {}'.format(cid))
+    for container in docker_list:
+        pu.info("Stopping container '{}'\n".format(container.name))
+        pu.exec_cmd('docker stop -t 1 {}'.format(container.name))
```
ivotron: Can we use `container.remove(force=True)` instead?

barbaraDorame:
Done! :+1:

ivotron:
thanks a lot for your efforts @barbaraDorame!

PR #570, opened 2019-03-27
Fixes #564 Implements popper add

### LINKED ISSUE #564: cli: implement a 'popper add' command
Implement a `popper add` command that allows users to import a workflow to the current project. The usage:

```bash
popper add <repo>[/folder[/wf.workflow]]
```

If `wf.workflow` is omitted, assume `main.workflow`. If `/folder` is omitted, assume `main.workflow` on root of the project or `.github/main.workflow`.

JayjeetAtGithub:
Implements the `popper add` command. On doing `popper add cplee/github-actions-demo` , the repository is cloned, the `main.workflow` file searched, scanned, copied to `project_root` and then the toplevel folder of local actions are also copied which are identified by parsing the workflow.

### DIFF
--- a/ci/test/add
+++ b/ci/test/add
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -ex
+
+source ./common
+init_test_repo
+cd $test_repo_path
+
+popper init
+
+popper add cplee/github-actions-demo
+test -f /tmp/mypaper/workflows/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/workflows/.github/main.workflow
+rm -rf workflows
+
+popper add cplee/github-actions-demo/.github/
+test -f /tmp/mypaper/workflows/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/workflows/.github/main.workflow
+rm -rf workflows
+
+popper add cplee/github-actions-demo/.github/main.workflow
+test -f /tmp/mypaper/workflows/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/workflows/.github/main.workflow
+rm -rf workflows
+
+echo "Popper add command ran successfully !"
--- a/ci/test/add
+++ b/ci/test/add
@@ -8,18 +8,15 @@ cd $test_repo_path
 popper init
 
 popper add cplee/github-actions-demo
-test -f /tmp/mypaper/workflows/.github/actions/jshint/Dockerfile
-test -f /tmp/mypaper/workflows/.github/main.workflow
-rm -rf workflows
+test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/main.workflow
 
 popper add cplee/github-actions-demo/.github/
-test -f /tmp/mypaper/workflows/.github/actions/jshint/Dockerfile
-test -f /tmp/mypaper/workflows/.github/main.workflow
-rm -rf workflows
+test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/main.workflow
 
 popper add cplee/github-actions-demo/.github/main.workflow
-test -f /tmp/mypaper/workflows/.github/actions/jshint/Dockerfile
-test -f /tmp/mypaper/workflows/.github/main.workflow
-rm -rf workflows
+test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/main.workflow
 
 echo "Popper add command ran successfully !"
--- a/ci/test/add
+++ b/ci/test/add
@@ -7,16 +7,29 @@ cd $test_repo_path
 
 popper init
 
-popper add cplee/github-actions-demo
+popper add github.com/cplee/github-actions-demo
 test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
 test -f /tmp/mypaper/main.workflow
 
-popper add cplee/github-actions-demo/.github/
+popper add github.com/cplee/github-actions-demo/.github/
 test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
 test -f /tmp/mypaper/main.workflow
 
-popper add cplee/github-actions-demo/.github/main.workflow
+popper add github.com/cplee/github-actions-demo/.github/main.workflow
 test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
 test -f /tmp/mypaper/main.workflow
+
+popper add github.com/JayjeetAtGithub/github-actions-demo@develop
+test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/main.workflow
+
+popper add github.com/JayjeetAtGithub/github-actions-demo/.github@develop
+test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/main.workflow
+
+popper add github.com/JayjeetAtGithub/github-actions-demo/.github/main.workflow@develop
+test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/main.workflow
+
 
 echo "Popper add command ran successfully !"
--- a/ci/test/add
+++ b/ci/test/add
@@ -7,6 +7,18 @@ cd $test_repo_path
 
 popper init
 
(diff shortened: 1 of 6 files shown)


JayjeetAtGithub:
@ivotron Please have a look !

JayjeetAtGithub:
@ivotron Got it ! Will be updating soon.

JayjeetAtGithub:
@ivotron 
When we can refer to reuse using `./test` in `uses`.
```
'action "test reuse" {
  uses = "./test"
  runs = "reuse"
}
```
So, why do we use 
```
'action "test reuse" {
  uses = "./ci/test"
  runs = "reuse"
}
```
This ought to be had asked before, but somehow my eyes couldn't catch it ? 
Can you please help me understand this ?

ivotron:
sure. References in a `uses` attribute of a `.workflow` file are relative to the root of the project. So if a workflow file is in `/project/path/containing/main.workflow` and there is an action in `/project/another/path/to/an/action` that the workflow needs, it will reference it by doing `./another/path/to/an/action`. In the example you wrote, the action is defined in `./ci/test`, which is a container-less action executed by the `HostRunner` class.

The `runs` attribute of an action block are references to commands within the action. In the example you gave, it is referring to the `reuse` command (defined in `./ci/test/reuse`).

ihth

JayjeetAtGithub:
@ivotron This is ready for review.

cli/popper/commands/cmd_add.py
```diff
@@ -0,0 +1,76 @@
...
+    '--branch',
+    help='Specifies the branch to download from.',
+    required=False,
+    default='master'
+)
+@pass_context
+def cli(ctx, path, branch):
+    """Imports a workflow from a remote project to the current project
+    directory. For now it can import workflows only from repos hosted at
+    github.com
```
ivotron: it is possible to download from arbitrary git URLs.

cli/popper/commands/cmd_add.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    project_root = scm.get_root_folder()
+    if not pu.is_popperized(project_root):
+        pu.fail('Repository has not been popperized.')
+        return
+
+    parts = path.split('/')
+    parts = list(filter(None, parts))
+    if len(parts) < 2:
+        pu.fail('Remote url format should be  <repo>[/folder[/wf.workflow]]')
```
ivotron: we could extend the format to be:

```
<url>/<repo>/<folder>/wf.workflow
```

we might be able to refactor the logic that is used [here](https://github.com/systemslab/popper/blob/1d4fbe5/cli/popper/gha.py#L199-L.235) so that we can reuse it for `popper add` as well.

cli/popper/commands/cmd_add.py
```diff
@@ -0,0 +1,76 @@
...
+import popper.utils as pu
+import popper.scm as scm
+from popper.cli import pass_context
+from distutils.dir_util import copy_tree
+
+
+@click.command('add', short_help='Import workflow from remote repo.')
+@click.argument('path', required=True)
+@click.option(
+    '--branch',
```
ivotron: for this functionality, we can use the `@<rev>` suffix so that we are consistent with the way revisions are provided in workflows. So instead of having a `--branch` flag, we would do something like

```
popper add https://github.com/org/repo/action@master
```

JayjeetAtGithub:
@ivotron Should i move the reusable function to parse the url that is both used in `download_actions()` and 
`cmd_add.py` to `utils.py` or keep it in `gha.py` ?

JayjeetAtGithub:
@ivotron I have made the changes. Please have a look.

cli/popper/commands/cmd_add.py
```diff
@@ -0,0 +1,74 @@
...
+    directory.
+    """
+    project_root = scm.get_root_folder()
+    if not pu.is_popperized(project_root):
+        pu.fail('Repository has not been popperized.')
+        return
+
+    parts = path.split('/')
+    parts = list(filter(None, parts))
+    if len(parts) < 3:
```
ivotron: similar to how `uses` is treated in a workflow, we can relax the requirement so that when the URL is omitted, it defaults to `github.com`. This is already handled by the `parse` function

cli/popper/commands/cmd_add.py
```diff
@@ -0,0 +1,74 @@
...
+        pu.fail(
+            'Required url format <url>/<user>/<repo>[/folder[/wf.workflow]]')
+
+    url, service, user, repo, _ = pu.parse('https://{}'.format(path))
+    if '@' in path:
```
ivotron: this can also be handled by the generic function

cli/popper/commands/cmd_add.py
```diff
@@ -0,0 +1,74 @@
...
+        branch = path.split('@')[-1]
+    else:
+        branch = 'master'
```
ivotron: we cannot make the assumption that the default branch is `master`. Instead, we should make `branch=None` and clone/checkout whatever is the default branch

cli/popper/utils.py
```diff
@@ -342,3 +342,39 @@ def find_recursive_wfile():
                 wfile = os.path.abspath(wfile)
                 wfile_list.append(wfile)
     return wfile_list
+
+
+def parse(url):
```
ivotron: this can also take care of extracting the version, and a `path` that would represent anything between the `<action>` and the `@` sign.

cli/popper/commands/cmd_add.py
```diff
@@ -0,0 +1,74 @@
...
+
+    cloned_project_dir = os.path.join("/tmp", service, user, repo)
+    scm.clone(url, user, repo, os.path.dirname(
+        cloned_project_dir), branch
+    )
+
+    if len(parts) == 3:
```
ivotron: would it be possible to have this logic be part of the `Workflow` class? We can create a static `Workflow.import_from_repo` or something like that. Or alternatively a function in the `gha` module that implements this logic. The idea being that we can keep a separation between the logic of Popper and what is done in the CLI implementation.

cli/popper/commands/cmd_add.py
JayjeetAtGithub: ok

cli/popper/commands/cmd_add.py
JayjeetAtGithub: @ivotron Can you please explain me a bit why do we need a `Workflow.import_from_repo` when the cli logic is already seperated in `cmd_add.py` ?

ivotron: sure. The logic of adding a workflow to the current project is currently implemented in [this block](https://github.com/systemslab/popper/blob/0ec35ffc4b2754a4a2da1d70a47df766f48074c0/cli/popper/commands/cmd_add.py#L35-L74). If, for example, we implement a web service that exposes popper functionality via a REST interface, we would not be able to use this functionality unless we opened a shell and executed `popper add`. Instead, it is preferable to encapsulate all the logic of what Popper does in the `gha.py` module, as this can be used in contexts different than via a terminal.

JayjeetAtGithub:
@ivotron Updated it. Please have a look !

JayjeetAtGithub:
@ivotron `test interrupt` is failing sometimes and only for `python3.7`. Can you please check it once.

ivotron:
@JayjeetAtGithub can you please rebase master? I made some changes to ctrl-c and `test interrupt`

JayjeetAtGithub:
@ivotron rebased

JayjeetAtGithub:
@ivotron some work is left...will be doing soon

ci/test/add
```diff
@@ -0,0 +1,71 @@
...
+source ./common
+init_test_repo
+cd $test_repo_path
+
+popper init
+
+popper add cplee/github-actions-demo@master
+test -f /tmp/mypaper/.github/actions/jshint/Dockerfile
+test -f /tmp/mypaper/main.workflow
+
```
ivotron: might be preferable to remove the workflow file and `.github` folder after each test. Or create a new subfolder for each distinct test so that we guard against the possibility of `test -f ...` being executed against the previous execution of `popper add`.

Also, we could test with other workflows, e.g. the example ones in https://github.com/actions

JayjeetAtGithub: @ivotron Updated the tests. Thanks

ivotron:
thanks!

Issue #549, opened 2019-03-20
cli: add shellcheck action to check test bash scripts


ivotron:
Add an action to the [`cli.workflow`](https://github.com/systemslab/popper/blob/master/ci/cli.workflow) workflow that runs [`shellcheck`](https://github.com/actions/bin/tree/master/shellcheck) on all the scripts in the [`ci/test/`](https://github.com/systemslab/popper/tree/master/ci/test) folder.

PR #572, opened 2019-03-29
shellcheck added and scripts validated

arshul:
Added the shell-check action, and modified the scripts to pass the shell-check. ( #549 )


arshul:
@ivotron this is ready to be reviewed.

ivotron:
thanks @arshul !

PR #573, opened 2019-03-29
Prompt for entering non defined secret

### LINKED ISSUE #566: cli: interactively define secrets
If a workflow defines secrets, and they are not available in the environment, prompt the user for their values. This needs to take care of the case when `popper run` is executed by a CI service, in which case the prompt is skipped

arshul:
It will use TravisCI's Env variable `CI` to detect whether to prompt or not to prompt ( resolves #566 )

### DIFF
--- a/cli/popper/gha.py
+++ b/cli/popper/gha.py
@@ -1,5 +1,5 @@
 from __future__ import unicode_literals
-from builtins import dict, str
+from builtins import dict, str, input
 from concurrent.futures import ThreadPoolExecutor, as_completed
 import multiprocessing as mp
 import hcl
@@ -15,7 +15,8 @@ class Workflow(object):
     """A GHA workflow.
     """
 
-    def __init__(self, wfile, workspace, quiet, debug, dry_run, reuse, parallel):
+    def __init__(self, wfile, workspace, quiet, debug,
+                 dry_run, reuse, parallel):
         wfile = pu.find_default_wfile(wfile)
 
         with open(wfile, 'r') as fp:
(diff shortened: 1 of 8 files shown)


arshul:
@ivotron while the internal tests run, they are not using `--no-prompt` flag. What I can do is change the occurrences of all `popper run` commands in the `test/` folder  to use `no-prompt` flag or we can change the functionality, i.e pass a flag `prompt` which when passed shows the prompt and not by default.
What are your thoughts on this?

arshul:
@ivotron any idea why tests do not run on new commits now??

cli/popper/gha.py
```diff
@@ -175,7 +176,11 @@ def check_secrets(self):
         for _, a in self.wf['action'].items():
             for s in a.get('secrets', []):
                 if s not in os.environ:
-                    pu.fail('Secret {} not defined\n.'.format(s))
+                    if os.environ.get("CI") == "true":
+                        pu.fail('Secret {} not defined\n.'.format(s))
+                    else:
```
ivotron: when `--dry-run` is given (`dry_run=True`) the prompt should be skipped as well, as users don't want to actually run the pipeline

arshul: In the case of `dry-run` instead of letting the action fail, I have set the value for an undefined secret as an empty string. This should keep the action from failing and the user won't have to enter the value for the secret. 
Works?

ivotron: I think that would work. Although the empty value is not read after that. An alternative is to do:

```python
if self.dry_run:
    return
```

at the beginning of the `check_secrets` function

ivotron:
thanks!

PR #577, opened 2019-04-02
Fixes#571 Adds returncode for CalledProcessError

JayjeetAtGithub:
Fixes the return code returning mechanism of `utils.exec_cmd`.

### DIFF
--- a/cli/popper/utils.py
+++ b/cli/popper/utils.py
@@ -150,6 +150,7 @@ def exec_cmd(cmd, verbose=False, debug=False, ignore_error=False,
             return t.decode('utf-8')
         return t
 
+    ecode = None
     # quick shortcut for 1) above
     if not verbose and not log_file:
         out = ""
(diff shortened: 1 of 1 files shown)


JayjeetAtGithub:
@ivotron Can you please tell me why this import is failing in travis while behaving correctly in local ?
```

[test cli_test] ./cli_test
+python ../../cli/popper/tests.py
Traceback (most recent call last):
  File "../../cli/popper/tests.py", line 1, in <module>
    import popper.utils as pu
ModuleNotFoundError: No module named 'popper'
```

cli/popper/utils.py
```diff
@@ -158,12 +158,14 @@ def b(t):
         try:
             out = check_output(cmd, shell=True, stderr=PIPE,
                                universal_newlines=True)
+            returncode = 0
```
ivotron: we can use `ecode` below here. That is, we can move the declaration of the `ecode` variable before this block of code and work with it instead of defining `returncode`

cli/popper/utils.py
```diff
@@ -158,12 +158,14 @@ def b(t):
...
         except CalledProcessError as ex:
+            returncode = ex.returncode
```
ivotron: same here, we can use `ecode`

cli/popper/utils.py
```diff
@@ -158,12 +158,14 @@ def b(t):
...
             if debug:
                 info('DEBUG: Catched exception: {}\n'.format(ex))
             if not ignore_error:
                 fail("Command '{}' failed: {}\n".format(cmd, ex))
-        return b(out).strip(), 0
+        return b(out).strip(), returncode
```
ivotron: and return `ecode`

JayjeetAtGithub:
@ivotron Please have a look !

ivotron:
thanks!

Issue #579, opened 2019-04-05
cli: improve message for uninitialized git repo


ivotron:
when executing `popper run` on an uninitialized git repo, popper reports the following:

```
ERROR: Needed a single revision.
```

this could be improved so that users know that the repo needs to have at least one commit.

ulshell:
I would like to work on @ivotron

ivotron:
thanks @ulshell , this is already being addressed in #581

PR #581, opened 2019-04-05
Change fail message

akashrajr1:
Addresses #579

PR #582, opened 2019-04-07
Claim for issue #579

ulshell:
Correction in fail message.


ivotron:
thanks for working on this @ulshell! This is already being addressed in #581

Issue #589, opened 2019-04-09
cli: ctrl-c for 'popper --parallel' not working on 2.7


ivotron:
It seems like the issue is due to [pythonfutures](https://github.com/agronholm/pythonfutures) as it doesn't handle KeyBoard Interruprts. It is evident from this PR, agronholm/pythonfutures#50 . It is not specific to ProcessPool / ThreadPool, but looks like an issue of pythonfutures module.

JayjeetAtGithub:
#594 build fails due to this issue...Looking into it.

JayjeetAtGithub:
@ivotron Can you please give me some insight about this issue ? #595 is failing due to this.

JayjeetAtGithub:
This seems like an issue in pythonfutures module

ivotron:
Since `--parallel` is not a critical feature, I would vote for implementing a workaround for now: if `--parallel` is given, inspect the workflow and, if a `HostRunner` action is part of the workflow, then show a message and exit. We then comment this on the corresponding test, make a reference to this issue and we keep it open

JayjeetAtGithub:
@ivotron Ok !

JayjeetAtGithub:
@ivotron The `SIGINT` is not caught in either of the runners. All of the runners are having similar behaviour like `HostRunner`, when executed in python2.7

ivotron:
I see. In this case then let's disable `--parallel` for 2.7 altogether, so we can investigate what's going on later

JayjeetAtGithub:
@ivotron i will disable interrupt tests for `--parallel`

ivotron:
closed as byproduct of closing #742

PR #595, opened 2019-04-10
Fixes#588 popper --parallel for HostRunner

JayjeetAtGithub:
Fixes `popper run --parallel` for `HostRunner` .

### DIFF
--- a/ci/test/interrupt
+++ b/ci/test/interrupt
@@ -104,6 +104,7 @@ fi
 
 #################
 # hostrunner actions
+if [[ $PY_VERSION != *"Python 2.7"* ]]; then
 
 mkdir -p actions/sleep
 
@@ -202,3 +203,6 @@ sleep 5
 ps aux > running_procs
 test $(cat running_procs | grep sleep | wc -l) -eq 0
 rm running_procs
+
+# }
+fi
--- a/ci/test/interrupt
+++ b/ci/test/interrupt
@@ -137,72 +137,68 @@ kill -2 $pid
 sleep 5
 
 ps aux > running_procs
-! cat running_procs | grep sleep
+test $(cat running_procs | grep sleep | wc -l) -eq 0
 rm running_procs
 
 #################
 # --parallel run on hostrunner
 
-# TODO: uncomment to reproduce #587 and #588
-# {
-#
-#mkdir -p actions/sleep
-#
-#cat <<EOF > main.workflow
-#workflow "test" {
-#  resolves = ["sleep"]
-#}
-#
-#action "sleep1" {
-#  uses = "./actions/sleep"
-#  args = "360"
-#}
-#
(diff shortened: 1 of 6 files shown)


JayjeetAtGithub:
@ivotron Please have a look

JayjeetAtGithub:
@ivotron Please have a look.

ci/test/interrupt
```diff
@@ -146,63 +146,63 @@ rm running_procs
 # TODO: uncomment to reproduce #587 and #588
 # {
 #
```
ivotron: we need to remove this comment block (and the commented closing braces at the bottom)

ci/test/interrupt
```diff
@@ -146,63 +146,63 @@ rm running_procs
...
+# give popper some time to start processes
+sleep 15
+
+# -2 is for SIGINT or Ctrl-C
+kill -2 $pid
+
+sleep 5
+
+ps aux > running_procs
+! cat running_procs | grep sleep
```
ivotron: @JayjeetAtGithub I tested on my machine and, while the test does not fail, the behavior is still the same. Not sure why the test is not failing, but this should make it fail accordingly:

```bash
test $(cat running_procs | grep sleep | wc -l) -eq 0
```

JayjeetAtGithub: @ivotron Can you please tell me what behaviour you are talking about ?

ivotron: yes, sorry, I should have explained better. The test has to fail, but it is not failing because line 206 is not catching the error. You can look at the log on travis, specifically [here](https://travis-ci.org/systemslab/popper/jobs/518407405#L1694). There should not be `sleep` processes running. This also fails when `--parallel` is not given, so it seems to be an issue with the way we're killing `HostRunner` processes

ivotron: if you replace line 206 with what I pasted above, the test should fail on your machine

JayjeetAtGithub: Thanks @ivotron ! Let me see if can find out whats going wrong.

JayjeetAtGithub:
@ivotron Fixed the bug for `python3.7` but couldn't understand what to do for `python2.7`. #589 has to be solved before.

JayjeetAtGithub:
@ivotron Modified `interrupt` tests for `--parallel` to run in only python3.7 and the build passes now. Please check the approach in fixing the `--parallel` for `HostRunner`.

ivotron:
awesome, thanks a lot! Can we add a check for the version of python in `cmd_run` and, if `--parallel` is passed, we show a message like `--parallel is only supported on Python 3` and exit?

JayjeetAtGithub:
ya sure

JayjeetAtGithub:
@ivotron Made the changes

ivotron:

Issue #586, opened 2019-04-09
cli: use click-log


ivotron:
make use of https://github.com/click-contrib/click-log to improve treatment of debug, verbose, quiet modes.

akashrajr1:
@ivotron I wanted some clarification about this issue.
1. Do we need to also change the implementation of `info` and `fail` in `utils` to use `click_log`? With `setLevel` we can filter out the level of logs we want.  `fail` being equivalent to `ERROR` level, `debug` equal to `DEBUG`. No changes for `quiet`. So without changing `fail` we won't be able to exercise the full power of `click_log`
2. And is `info` equivalent to `INFO` logging level? Shouldn't `info` be always printed out to the stdout irrespective of the level set by `setLevel`?
3. Are we condensing the functionality of `--debug` into the inbuilt `--verbosity` option. Because in the click log [docs](https://click-log.readthedocs.io/en/stable/), `--debug` is implemented as a `setLevel`
> The simple_verbosity_option() decorator adds a --verbosity option that takes a (case-insensitive) value of DEBUG, INFO, WARNING, ERROR, or CRITICAL, and calls setLevel on the given logger accordingly.

ivotron:
thanks a lot for working on this @akashrajr1 ! please see my answers inline


> 1. Do we need to also change the implementation of `info` and `fail` in `utils` to use `click_log`? With `setLevel` we can filter out the level of logs we want.  `fail` being equivalent to `ERROR` level, `debug` equal to `DEBUG`. No changes for `quiet`. So without changing `fail` we won't be able to exercise the full power of `click_log`

yes, that would be ideal. There might also be a way of disabling/re-enabling logging, in which case click-log would also work for implementing the `--quiet` functionality

> 2. And is `info` equivalent to `INFO` logging level? Shouldn't `info` be always printed out to the stdout irrespective of the level set by `setLevel`?

yes, it is equivalent. In our case, `info` is the default logging level

> 3. Are we condensing the functionality of `--debug` into the inbuilt `--verbosity` option. Because in the click log [docs](https://click-log.readthedocs.io/en/stable/), `--debug` is implemented as a `setLevel`

not sure I understand this question. do you mean allowing users to specify a `--verbosity` option? In our case, popper users will not be aware of logging levels. Instead, the semantics we have is: actions are verbose by default and they're silenced when `--quiet` is given. If `--debug` is given, then we print everything that popper is doing.

akashrajr1:
[...]

PR #598, opened 2019-04-11
Adds logging levels

akashrajr1:
Addresses #586


akashrajr1:
@ivotron can you help me with this [build error](https://travis-ci.org/systemslab/popper/jobs/519746037#L311) ? It is passing the test [here](https://travis-ci.org/systemslab/popper/jobs/519746037#L2577).
I decided to use `colorlog` which is essentially the `logging` module with color support; which I did include in `setup.py`.

ivotron:
Sorry for the delay! PR #563 will remove the pip-less method for installing popper.

akashrajr1:
@ivotron ready for review!
The log colors are controlled using
```
log_colors = {
    'DEBUG': 'bold_purple',
    'INFO': 'reset',
    'WARNING': 'bold_yellow',
    'ERROR': 'bold_red',
    'CRITICAL': 'bold_red,bg_white'
}
```

akashrajr1:
@ivotron can you point me towards where [this](https://travis-ci.org/systemslab/popper/jobs/520555705#L520) debug info is generated? It has debug color scheme but not the 'DEBUG' prefix

cli/popper/commands/cmd_version.py
```diff
@@ -8,4 +8,5 @@
 @pass_context
 def cli(ctx):
     """Displays version of Popper and exit."""
+    pu.setup_logging()
```
ivotron: can we put this in [`cli.py`](https://github.com/systemslab/popper/blob/master/cli/popper/cli.py)? Then that would setup for the entire app

cli/popper/utils.py
```diff
@@ -104,19 +116,112 @@ def is_popperized(rootfolder):
...
+                log('')
+                remove_level()
+                rest = ''.join(msg_lines[index:])
+                log(rest+"\n")
+        else:
+            log(msg)
+    return wrapper
+
+
+@multiline
```
ivotron: since we are now using `logging`, it would be nice if we can use `log.fail`, `log.warn`, etc... directly on the code without having to wrap those calls. In the case of multiline output, we could leave that to the code that is invoking the logging call. This means that the code doing:

```python
log.debug('this is a multi-line output \n that can be put in multiple logging calls')
```

can be transformed to this:

```python
log.debug('this is a multi-line output')
log.debug(' .   that can be put in multiple')
log.debug(' .   logging calls')
```

in this way we don't need to worry and can make use of `logging` directly.

cli/popper/utils.py
```diff
@@ -104,19 +116,112 @@ def is_popperized(rootfolder):
...
 def fail(msg):
     """Prints the error message on the terminal."""
-    click.secho('ERROR: ' + msg, fg='red', bold=True, err=True, nl=False)
+    log.error(msg[:-1])
```
ivotron: If we remove the `fail`, `warn`, `info` wrappers, we can assume that the code that will call `logging` functions will **not** be ending messages new lines

cli/popper/utils.py
```diff
@@ -221,15 +319,14 @@ def b(t):
                 num_times_point_at_current_sleep_time += 1
 
                 if debug:
-                    info('DEBUG: sleeping for {}\n'.format(sleep_time))
+                    debug_log('sleeping for {}\n'.format(sleep_time))
```
ivotron: we can remove the conditional here as well

cli/popper/gha.py
```diff
@@ -491,7 +491,7 @@ def docker_create(self, img):
         env_vars.update({'HOME': os.environ['HOME']})
         volumes = [self.workspace, os.environ['HOME'], '/var/run/docker.sock']
         if self.debug:
-            pu.info('DEBUG: Invoking docker_create() method\n')
+            pu.debug_log('Invoking docker_create() method\n')
```
ivotron: now that we are using `logging`, we can now remove the conditionals and do `log.debug` directly (assuming we will remove the wrappers in `utils.py` as mentioned below)

cli/popper/utils.py
```diff
@@ -175,32 +278,27 @@ def b(t):
 
     if log_file:
         if verbose:
-            if debug:
-                info('\nDEBUG: Creating file for combined stdout/stderr\n')
+            debug_log('\nCreating file for combined stdout/stderr')
```
ivotron: We can also use this opportunity to have `logging` deal with log files. Instead of us having to create the log files, we can instruct `logging` to create them. This will simplify things greatly, as we only need to do `log.info()`.

If we do the above, then we need to check the value of `log_file` and setup `logging` accordingly so that it also creates log files.

ivotron: This means that the logic described at the beginning of the function (where we describe the four cases), can be handled entirely by making use of `logging`, and we only create a single stream of output that we pass to `log.info()`.

cli/popper/utils.py
ivotron: In the case of `FAIL`, we can implement a custom handler so that when `log.fail` is invoked, we show a message and raise an exception.

cli/setup.py
```diff
@@ -18,13 +18,14 @@
         'future; python_version == "2.7"',
         'spython',
         'click',
-        'requests',
+        'requests == 2.14.2',
```
ivotron: this is a leftover dependency from v1.x. We can remove this as well as `lark-parser` and `pyaes`

cli/popper/utils.py
akashrajr1: So I was using `@multiline` because we have logging levels also implemented. If we use:
```
log.debug('this is a multi-line output')
log.debug('    that can be put in multiple')
log.debug('    logging calls')
```
it will result in 
```
DEBUG: this is a multi-line output
DEBUG:    that can be put in multiple
DEBUG:    logging calls
```

ivotron: Having one DEBUG per line would be OK.

cli/setup.py
akashrajr1: Was just checking the dependencies, seems like `docker-py` depends on `requests==2.20.0`. From this [requirements](https://github.com/docker/docker-py/blob/master/requirements.txt)

ivotron: yeah, that is handled by `pip` though. So we only need to declare `docker-py` as a dependency here, and pip will handle docker-py's dependencies

cli/popper/utils.py
akashrajr1: we need wrappers for `fail` because we need to call `sys.exit(1)` 
and for `info` as well because i am wrapping it inside `remove_level()` and `add_level()`
Without utilizing those functions, if we retain `levelname` for `info` it will be like [this](https://travis-ci.org/systemslab/popper/jobs/518905794#L658)

ivotron: Sorry @akashrajr1 , I briefly mentioned it in one of the earlier comments but I should have explained better what I meant.

The logging module has facilities for implementing custom handlers that change the default behavior. This helps us so that we don't need to wrap log calls, rather we provide a custom handler that the logging module makes use of.

With this, we can for example set the behavior in such a way that for INFO level messages, no prefix is shown; and that for FAIL, a message is shown and then an exception is raised (or sys.exit(1) is invoked).

akashrajr1: Thanks @ivotron, so the way forward is we implement our own custom handlers for `sys.exit(1)` and no `levelname` for `info`?
I was also trying a different approach by created a `log.py` module so that we can use `log.info()` directly. I still don't understand what might be the cons of using wrapping log calls.

cli/popper/utils.py
akashrajr1: Can you explain a bit about how we are generating logs? Specifically where in the code are we writing to the files? Thanks

ivotron: sure. We open the file(s) for writing in [this block](https://github.com/systemslab/popper/blob/master/cli/popper/utils.py#L176-L185), and write to them [here](https://github.com/systemslab/popper/blob/master/cli/popper/utils.py#L213).

However, now that we are using `logging`, we can simplify things:

  * expose a `--log` flag to `popper run` that allows users to specify where to write a single log file to. We can have it disabled by default (empty string or `None`).
  * when `--log` is not empty, we pass it to the `setup_logging` function so that we configure the `logging` custom handler in a way that we create a log file in the path specified by the value of the flag
  * then, the runners (host, docker and singularity) only need to write output using the `logging` API and do not need to deal with files

ihth

cli/popper/utils.py
ivotron: yeah, the handler facility from `logging` is the way a python codebase defines custom logging behavior. Wrapping log calls has the potential of introducing performance overheads, e.g. calling the functions that change levels every time we log a message might not be negligible.

ivotron:
@akashrajr1 thinking a bit more about this feature: we can define a new level on a custom handler that corresponds to messages logged by actions. I'll reference [the table with default logging levels from the official doc](https://docs.python.org/3/library/logging.html#logging-levels) (reproducing below for convenience):


Level | Numeric value
-- | --
CRITICAL | 50
ERROR | 40
WARNING | 30
INFO | 20
DEBUG | 10
NOTSET | 0

We can add a new `ACTION_INFO` level between `DEBUG` and `INFO`. This would be the level that runners use to log messages that are coming from actions, whereas the `INFO` level is used by popper to log messages that come from popper itself. For example, take the following sample output:

```
$ popper run
[popper] cloning actions from repositories
[install] docker pull node:11.6.0
[install] docker create node:11.6.0 npm install
[install] docker start
audited 195 packages in 2.498s
found 0 vulnerabilities

[test] docker pull node:11.6.0
[test] docker create node:11.6.0 npm test
[test] docker start

> github-actions-demo@1.0.0 test /private/tmp/mypaper/github-actions-demo
> mocha ./tests --recursive



Listening on port 8080
  GET /
    ✓ should respond with hello world (48ms)


  1 passing (74ms)

[lint] docker build -t popper/cplee/github-actions-demo/jshint:c824a5a /private/tmp/mypaper/github-action
s-demo/./.github/actions/jshint
[lint] docker create popper/cplee/github-actions-demo/jshint:c824a5a
[lint] docker start
[branch-filter] docker build -t actions/bin /tmp/actions/github.com/actions/bin/./filter
[branch-filter] docker create actions/bin branch master
[branch-filter] docker start
refs/heads/master matches refs/heads/master
[deploy] docker build -t actions/bin /tmp/actions/github.com/actions/bin/./sh
[deploy] docker create actions/bin env
[deploy] docker start
Running 'env'...
POPPER_EVENT_NAME=push
HOSTNAME=1d490843e5ff
PHONY_SECRET=foo
HOME=/Users/ivo
POPPER_WORKSPACE=/private/tmp/mypaper/github-actions-demo
POPPER_SHA=c824a5a
POPPER_ACTOR=popper
POPPER_REF=refs/heads/master
Successfully ran 'env'

Workflow finished successfully.
```

messages between squared brackets are regular `INFO`, whereas the ones coming from the action are `ACTION_INFO` messages. With this, we can turn off messages that come from actions (when `--quiet` is given), but at the same time show popper messages

cli/popper/utils.py
akashrajr1: Thanks that clear things up

akashrajr1:
Sounds good @ivotron , will incorporate it

akashrajr1:
@ivotron the update is almost ready, but I can't figure from where the runners are generating their logs. Can you help me?

cli/popper/utils.py
akashrajr1: So I wanted to clarify, in this Logic mentioned
```
    # the main logic is the following:
    #
    # 1) verbose=False and log_file=None
    #      ==> don't write anything to stdout/log
    # 2) verbose=True and log_file=None
    #      ==> combine stdout/stderr in the same stream and print it to stdout
    # 3) verbose=False and log_file not None
    #      ==> write two files, one .out and one .err
    # 4) verbose=True and log_file not None
    #      ==> combine stdout/stderr, write to stdout and to a SINGLE log file
```
*  Regarding point 1 and 2, are these conditions ever meet? Because the only call to `exec_cmd` [here](https://github.com/systemslab/popper/blob/d719f9276dbbc7c4eb56af4a78b05d12c7b68668/cli/popper/gha.py#L710-L713) is taking `log_file=self.log_filename` and `self.log_filename` is always defined in the `ActionRunner`, so the condition `log_file=None` is never met.
* For point 3 I was thinking I will use the current level being used ( default: `ACTION_INFO` , `--debug`: `DEBUG` and `--quiet`: `INFO`) for `.out` file and the level `ERROR` for `.err` file.
* For point 4, the level `DEBUG` for `.log` files.

ivotron: > @ivotron just wanted to confirm, is [this](https://github.com/systemslab/popper/blob/d719f9276dbbc7c4eb56af4a78b05d12c7b68668/cli/popper/gha.py#L530) `pu.info(b(l))` where we print our action info?

yeah, that's for the dockerrunner. In the case of the singularity runner, it is [here](https://github.com/systemslab/popper/blob/d719f9276dbbc7c4eb56af4a78b05d12c7b68668/cli/popper/gha.py#L672).

For the docker runner, I already took out the code that was dealing with writing to log files. This can be simplified further by removing the conditional for when `quiet` and `debug` are given.

For singularity, we can simplify the logic of the code significantly, as we can remove conditionals that setup log files, as well as those that deal with `quiet`, `verbose` and `debug`.

ivotron:
> @ivotron just wanted to confirm, is [this](https://github.com/systemslab/popper/blob/d719f9276dbbc7c4eb56af4a78b05d12c7b68668/cli/popper/gha.py#L530) `pu.info(b(l))` where we print our action info?

yeah, that's for the dockerrunner. In the case of the singularity runner, it is [here](https://github.com/systemslab/popper/blob/d719f9276dbbc7c4eb56af4a78b05d12c7b68668/cli/popper/gha.py#L672).

For the docker runner, I already took out the code that was dealing with writing to log files. This can be simplified further by removing the conditional for when `quiet` and `debug` are given.

For singularity, we can simplify the logic of the code significantly, as we can remove conditionals that setup log files, as well as those that deal with `quiet`, `verbose` and `debug`.

cli/popper/utils.py
ivotron: couple of things that can help simplify the code:

  * as this function is now only used by the `HostRunner`, we can move it from `utils.py` to `gha.py` 
  * drop the `verbose` print to `log.debug`, `log.info` and `log.actioninfo` (assuming that's the name of the new level)
  * drop `log_file` as we will now deal with files via the custom `logging` handler (i.e. there's no need to create files here in this function, as that'll be done by the handler)

ihth

akashrajr1: So are we still generating two separate files `.err` and `.out`/ or one `.log` file? And are we mirroring the output thrown to the stdout in the logs or will it be different?

ivotron: in general, the use of `logging` means two things:
 1. the entire popper codebase will **only** use `log.debug`, `log.actioninfo`, `log.info`, etc.. to log messages. This means that we don't need to add special logic (the conditionals that check for the values of `debug` or `quiet`) to determine when these get logged, nor the code base needs to directly create files. The main goal of using a logging library is to delegate those tasks to the it.
 2. we can configure the behavior of the `logging` handler so that it knows which messages are logged (the logging level) and to where they get sent to (stdout/stderr, files, a database, etc.). This configuration can be done based on the flags that users passed to `popper run`:

      * if `--log` is given, then instruct the `logging` custom handler to create a **single** file.
      * if `--quiet` is given, then increase the level so that `ACTION_INFO` messages are turned off (as `ACTION_INFO` would be the default level).
      * if `--debug` is given, then decrease the level so that `DEBUG` messages are also logged.

akashrajr1:
> This can be simplified further by removing the conditionals for when `quiet` and `debug` are given.

Already took care of this. 

**Fyi**: The PR is updated with `ACTION_INFO` level and I am using `log.action_info()` for that.

And is possible to remove [this](https://github.com/systemslab/popper/blob/634c14ca3a47c9df0c2b4d3e71649681ba9325f1/cli/popper/gha.py#L512) condition? If we can then we don't actually need to propagate the `quiet` in Runners. I think `debug` flag after the last commit isn't used for any condition checking. So I will clean up the code by removing it. Essentialy `debug` and `quiet` will be used only once at the beginning to set the logging level.

ivotron:
> And is possible to remove [this](https://github.com/systemslab/popper/blob/634c14ca3a47c9df0c2b4d3e71649681ba9325f1/cli/popper/gha.py#L512) condition?

yes, that can be removed

> If we can then we don't actually need to propagate the `quiet` in Runners. I think `debug` flag after the last commit isn't used for any condition checking. So I will clean up the code by removing it. Essentialy `debug` and `quiet` will be used only once at the beginning to set the logging level.

yeah, that's the ideal, to cleanup the codebase so that the code just uses `log.<level>` and delegate that to the logging handler

cli/popper/utils.py
akashrajr1: So I added the `--log-file` flag  which takes and generates a logfile, and it has the same level as the `stdout` , so that whatever is shown on the console is written to a logfile. 
If the flag is not present I create a logfile in the path `popper_logs\` and a file with the name `log` + `timestamp` in that directory.

akashrajr1:
Was the previous logfile code somehow connected to travis ci build logs? By removing it now I am not getting any output for test actions on the log. 
I thought Travis CI was represented as `stdout`, but maybe it is `stderr`?

cli/popper/utils.py
```diff
@@ -2,147 +2,65 @@
...
 
         if add_to_process_list:
             popper.cli.process_list.append(p.pid)
 
-        if debug:
-            info('DEBUG: Reading process output\n')
+        log.debug('Reading process output')
 
         while ecode is None:
+            # when we are not writing output to stdout, print dot progress
```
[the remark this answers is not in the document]
ivotron: @akashrajr1 BTW, this is also the reason why there's no output showing on TravisCI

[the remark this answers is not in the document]
akashrajr1: Thanks a lot!! I will fix it

cli/popper/log.py
```diff
@@ -0,0 +1,161 @@
...
+        return result
+
+
+class PopperLogger(logging.Logger):
+    """
+    A Logger so that we can add popper fail and action_info log methods
+    """
+
+    @staticmethod
+    def remove(msg):
```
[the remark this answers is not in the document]
akashrajr1: Previous commits were reflecting that, but it seems that 10000 is the limit for Travis Build Log so the build was like [this](https://travis-ci.org/systemslab/popper/jobs/522838826). One has to look at the [raw logs](https://api.travis-ci.org/v3/job/522838826/log.txt), which don't properly visualize ANSI color codes.
But you are right the overhead is costly, and now that I know how we are interacting with Travis, I think using the `remove()` over there should solve the issue.

cli/popper/utils.py
```diff
@@ -2,147 +2,67 @@
...
-            else:
-                # when we are not writing output to stdout, print dot progress
-                if sleep_time < 30 \
-                        and num_times_point_at_current_sleep_time == 5:
-                    sleep_time *= 2
-                    num_times_point_at_current_sleep_time = 0
+        for line in iter(p.stdout.readline, ''):
+            line_decoded = b(line)
+            log.info(line_decoded[:-1])
+        # when we are not writing output to stdout, print dot progress
```
ivotron: since the above loop consumes all the output, the block below (the one that used to sleep before; lines 43-53) are not needed anymore. As a consequence, variables `sleep_time` and `num_times_point_at_current_sleep_time` are also not needed.

Also, since this function is only used by the `HostRunner`, we can move it to `gha.py` as

akashrajr1: @ivotron Updated it

ivotron:
thanks you so much for your patience @akashrajr1 !

PR #605, opened 2019-04-19
Fixes#604 Enhance code quality

JayjeetAtGithub:
Enhances code quality according to reports from Codacy test


cli/popper/exceptions.py
```diff
@@ -1,6 +1,6 @@
 import click
-from click._compat import get_text_stderr
 from click import secho
+from click._compat import get_text_stderr
```
ivotron: another thing we can use to simplify the code is to remove this file (`exceptions.py`) as only `cli.py` uses it to throw a `UsageError` exception, which can be changed for a `ClickException`.

JayjeetAtGithub: @ivotron Sure ! Also this PR is still not ready. I will fix some more issues soon.

JayjeetAtGithub:
@ivotron You may have a look. Thanks.

ivotron:
thanks a lot!

Issue #627, opened 2019-04-28
cli: show colorized output from steps


ivotron:
currently, if a step generates colored output, `popper run` is not showing it

akashrajr1:
@ivotron how do I reproduce the required colors for the example? 
And if the goal is to use colors, we can also set them through `popper` so that even if an example doesn't produce colors, `popper` can do that on its own.

ivotron:
thanks a lot for looking into this @akashrajr1! To produce colors out of a container:

```bash
docker run --rm -t -e "TERM=xterm-256color" alpine:3.9 ls -l
```

so it might be a matter of enabling `tty` when `popper run` creates containers.

PR #635, opened 2019-05-03
Show colorized output from actions

akashrajr1:
Addresses #627

### DIFF
--- a/ci/test/parallel
+++ b/ci/test/parallel
@@ -84,52 +84,52 @@ action "ls cmds" {
 
 action "curl github" {
   uses = "actions/bin/curl@master"
-  args = ["github.com"]
+  args = ["-L", "github.com"]
 }
 
 action "curl google" {
   uses = "actions/bin/curl@master"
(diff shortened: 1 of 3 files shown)


akashrajr1:
@ivotron I am get `UnicodeDecodeError` for some bytes, so I added a `try` block for now to bypass it. But this is resulting in loss of information. Seems like we need to change `utf-8` to something else.

The [example](https://github.com/cplee/github-actions-demo.git) I ran, choked on these bytes `b'\xe2`', `b'\x9c`' and `b'\x93`'

ivotron:
thanks a lot for working on this @akashrajr1! One thing that occurs to me is to take a look at how [`docker-compose`](https://github.com/docker/compose) does it, as it is also using `docker-py`. Seems that their [`utils.py`](https://github.com/docker/compose/blob/master/compose/utils.py) file is where this is implemented.

ivotron:
will be worked on a different PR

Issue #619, opened 2019-04-28
cli: add ability of running a command without implementing an action


ivotron:
Sometimes, we want to run a command on the local machine without having to implement an action. To achieve this, we can introduce a special `sh` action name that indicates popper to run the action using the `HostRunner`. For example:

```hcl
action "foo" {
  uses = "sh"
  runs = ["ls", "-l"]
}
```

the above allows to execute arbitrary commands on the local machine without having to create an action.

PR #639, opened 2019-05-06
Adds sh action

JayjeetAtGithub:
Addresses #619

### DIFF
--- a/cli/popper/gha.py
+++ b/cli/popper/gha.py
@@ -73,7 +73,7 @@ class WorkflowRunner(object):
 
         for _, a in self.wf.actions:
             if ('docker://' in a['uses'] or 'shub://' in a['uses'] or
-                    './' in a['uses']):
+                    './' in a['uses'] or a['uses'] == 'sh'):
                 continue
 
             url, service, usr, repo, action_dir, version = scm.parse(a['uses'])
@@ -105,6 +105,12 @@ class WorkflowRunner(object):
     def instantiate_runners(self):
         """Factory of ActionRunner instances, one for each action"""
         for _, a in self.wf.actions:
+            if a['uses'] == 'sh':
+                a['runner'] = HostRunner(
+                    a, self.workspace, self.env,
+                    self.dry_run)
+                continue
+
             if 'docker://' in a['uses']:
                 a['runner'] = DockerRunner(
                     a, self.workspace, self.env,
@@ -166,7 +172,7 @@ class WorkflowRunner(object):
             'GITHUB_ACTOR': 'popper',
             'GITHUB_REPOSITORY': repo_id,
             'GITHUB_EVENT_NAME': self.wf.on,
-            'GITHUB_EVENT_PATH': '/{}/{}'.format(self.workspace,
+            'GITHUB_EVENT_PATH': '{}/{}'.format(self.workspace,
                                                  'workflow/event.json'),
             'GITHUB_SHA': scm.get_sha(),
             'GITHUB_REF': scm.get_ref()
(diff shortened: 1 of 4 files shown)


JayjeetAtGithub:
@ivotron You may have a look.

cli/popper/gha.py
```diff
@@ -537,18 +543,21 @@ def __init__(self, action, workspace, env, dry):
...
         cwd = self.cwd
-        if not self.dry_run:
-            if 'repo_dir' in self.action:
-                os.chdir(self.action['repo_dir'])
-                cmd[0] = os.path.join(self.action['repo_dir'], cmd[0])
-            else:
-                os.chdir(os.path.join(cwd, self.action['uses']))
-                cmd[0] = os.path.join(cwd, self.action['uses'], cmd[0])
+        if self.action['uses'] == 'sh':
+            cmd = self.action.get('runs')
```
ivotron: here we also need to consider `args`, e.g.:

```hcl
action "foo" {
  uses = "sh"
  runs = "./mycommand"
  args = ["some", "args"]
}
```

ivotron: In addition, paths are specified with respect to the root of the project, so we need to `cd` to the root of the project as it is done below. More details are available in the [docs](https://popper.readthedocs.io/en/latest/sections/extensions.html#host).

JayjeetAtGithub:
@ivotron Please review !

cli/popper/gha.py
```diff
@@ -537,18 +543,27 @@ def __init__(self, action, workspace, env, dry):
...
+            cmd.extend(self.action.get('args', []))
+
+            if not self.dry_run:
+                os.chdir(cwd)
```
ivotron: depending on where `popper run` is executed, `cwd` might not be the root of the project folder. So we can instead run `pu.get_git_project_root()` here instead.

cli/popper/gha.py
```diff
@@ -537,18 +543,27 @@ def __init__(self, action, workspace, env, dry):
...
+                if cmd[0].startswith('./'):
```
ivotron: in the case of `uses = "sh"`, this does not apply, since we are running from the root of the project

cli/popper/gha.py
JayjeetAtGithub: @ivotron Yes, I was only thinking about `main.workflow` in the root.

cli/popper/gha.py
JayjeetAtGithub: If we need to execute a bash script located in the root, then how to reference it ?

ivotron: do you mean something like the following?

```hcl
action "run on host" {
  uses = "sh"
  runs = "./scriptatroot"
}
```

ivotron: can also be:

action "run on host" {
  uses = "sh"
  runs = "scriptatroot"
}

JayjeetAtGithub: @ivotron Thanks! Got it.

JayjeetAtGithub: @ivotron So can we set the default runs to `sh -c`  i.e when `runs` attribute not provided ?

ivotron: yeah, that's a good idea!

JayjeetAtGithub:
@ivotron Please check ! Thanks

ivotron:
thanks @JayjeetAtGithub . Do you think you can add the changes to the docs in this PR as well? What I mentioned [above](https://github.com/systemslab/popper/pull/639#pullrequestreview-235718628).

JayjeetAtGithub:
@ivotron Updated the docs.

ivotron:

Issue #172, opened 2017-10-15
cli: add support for running a workflow in vagrant


ivotron:
Add vagrant as a runtime available for running workflows. We can have a box with Docker inside to which we share the workspace.

JayjeetAtGithub:
@ivotron [`this`](https://pypi.org/project/python-vagrant/) package is a python api wrapper for `vagrant-cli`, but the problem is that it is not being maintained for 2 years now and also lacks proper documentation. Please have a look. Is there any alternative vagrant-cli wrapper? If this package is really unusable then we might have to proceed by using `utils.exec_cmd` to process raw commands. What do you say ? Thanks :smile:

JayjeetAtGithub:
@ivotron Also what are providers that will be used with `VagrantRunner` in our case, like from VirtualBox, docker, VMWare, etc?

JayjeetAtGithub:
@ivotron If we would use raw commands, then how to get back the output in `utils.exec_cmd()` as a byte string?

ivotron:
thanks a lot for looking into this @JayjeetAtGithub! it seems that there's quite a significant amount of effort put into the [`python-vagrant`](https://github.com/todddeluca/python-vagrant) project. I'd vote for using that

JayjeetAtGithub:
@ivotron Sure ! Thanks

JayjeetAtGithub:
@ivotron Can we use a `vagrant://` prefix with box names to refer boxes from [here](https://app.vagrantup.com/) in the `uses` attribute of an action?

ivotron:
yeah, that's a great idea!

Issue #550, opened 2019-03-20
cli: add support for running multiple workflows concurrently


ivotron:
Currently, if two or more instances of `popper run` are executed concurrently on the same (or similar) workflow, unexpected issues might arise. For example, the `/tmp/actions/` folder is shared across all instances and repos are cloned there and checked out to the version specified in a workflow. Another issue is that containers are not uniquely named and, as a consequence, two workflows with the same actions will potentially clash when creating/running/deleting containers.

One way of taking care of this is to generate a `popper run` ID that might be based on the path to `.workflow` file, plus `datetime.datetime.now()`. This then can be used to:

 1. uniquely identify containers, e.g. name a container based on the action name and append the `popper run` ID.
 2. create an actions cache folder unique for the `popper id`.

Other things might need to be taken care of. So we need to analyze the overall workflow execution flow to identify more issues like this.

akashrajr1:
@ivotron correct me if I didn't understand it properly. We append an id generated using the path to the workflow and timestamp. This is used to uniquely identify
* docker containers
* actions cache folder

For my Gsoc proposal I was thinking of expanding it. A new feature `popper multirun` which would parallelly execute all workflows using the underlying mechanism of  this feature: all docker container  and actions cache folder having unique ids (in case of a duplicate workflow)  
PS: I think `popper multirun` might be equivalent to `popper run --recursive --parallel-workflows`

ivotron:
the `multirun` command seems like a useful feature. It is difficult to know in advance which workflows to run though, as users might be executing workflows from multiple projects. So for this issue we can focus on the core functionality, which is to assume that a user can invoke `popper run` in ad-hoc ways (e.g. running workflows from multiple projects). For that to work, the `popper run` logic needs to consider that the user might have started another instance of `popper run`, without us knowing anything about it.

akashrajr1:
Won't `popper id` solve this issue of having multiple `popper run` instances running? 
Also for `multirun` maybe we can ask confirmation from the user beforehand as to which workflows he would like to execute.
Eg. `popper multirun`
`Found 3 workflow`
`1. a.workflow`
`2. b.workflow`
`3. c.workflow`
Then the user enters id corresponding to the workflows he wants to run?
`Enter workflow ids to run: 2 3`

ivotron:
that assumes that workflows all pertain to the same project. But users might be running workflows from different projects. A `multirun` feature would still need the functionality described in the OP. So it might be best to treat that separately.

PR #644, opened 2019-05-11
Adds vagrant runner

JayjeetAtGithub:
Addresses #172

### DIFF
--- a/cli/popper/gha.py
+++ b/cli/popper/gha.py
@@ -11,6 +11,7 @@ from subprocess import CalledProcessError, PIPE, Popen, STDOUT
 import hcl
 import yaml
 import docker
+import vagrant
 from spython.main import Client as sclient
 
 import popper.cli
@@ -73,7 +74,8 @@ class WorkflowRunner(object):
 
         for _, a in self.wf.actions:
             if ('docker://' in a['uses'] or 'shub://' in a['uses'] or
-                    './' in a['uses'] or a['uses'] == 'sh'):
+                './' in a['uses'] or 'vagrant://' in a['uses'] or
+                    a['uses'] == 'sh'):
                 continue
 
             url, service, usr, repo, action_dir, version = scm.parse(a['uses'])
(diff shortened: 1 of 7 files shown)


JayjeetAtGithub:
@ivotron While adding `VagrantRunner`,  in implementing the `--reuse` feature, we need to check the existence of a VM. For this, we need a vagrant cli functionality `vagrant global-status` which is absent in `python-vagrant`. I forked `python-vagrant` and implemented the feature and then used it to build and test `VagrantRunner` temporarily. Also, this is due to which the builds are failing. So, what shall we do now, continue with the forked lib or is there any other way around? Please have a look. Thanks.

JayjeetAtGithub:
@ivotron `virtualbox` installation is failing in Travis CI even after several different attempts. Some answers say that we can't run `virtualbox` in Travis CI. Can you please help me out with this?

ivotron:
hey @JayjeetAtGithub, thanks a lot for working on this. We have also tried to install virtualbox on Travis in the past without success. If this gets too much of your time, then I'd say we leave it out for now.

Regarding the changes done to the `python-vagrant` package, we can open a PR on the upstream repo. We can point to the fork in the `setup.py` file in the meantime (hopefully they'll merge the PR).

JayjeetAtGithub:
@ivotron Does that mean we won't have a `test vagrant` in the codebase and `VagrantRunner` would depend on local machine tests?

ivotron:
@JayjeetAtGithub , yes, that would mean we have to test manually. Alternatively, we could also see if we can use the docker provider for vagrant. In this way we test the popper/vagrant interaction.

JayjeetAtGithub:
@ivotron Can we clone the remote actions within the runners ?. This is needed to reuse the environment in case of vagrant which otherwise gets destroyed during erasing the prev clone and recloning. We would like to check whether the prev clone is present and if present we would skip cloning.

JayjeetAtGithub:
@ivotron Can you also give me some idea how to set the env vars inside the vagrant VM's ?
I was doing `vagrant ssh --command "echo export VAR=value >> ~/.bashrc"`, but this was'nt working.

ivotron:
> @ivotron Can you also give me some idea how to set the env vars inside the vagrant VM's ?
> I was doing `vagrant ssh --command "echo export VAR=value >> ~/.bashrc"`, but this was'nt working.

I believe this can be done by modifying the `Vagrantfile`, so that we add a shell provisioner and define the environment variables there.

ivotron:
> @ivotron Can we clone the remote actions within the runners ? This is needed to reuse the environment in case of vagrant which otherwise gets destroyed during erasing the prev clone and recloning. We would like to check whether the prev clone is present and if present we would skip cloning when the `--reuse` flag is passed. In this way, we can preserve the vagrant environment for reusing.

This is related to #550 and #652, in the sense that if those two were already solved, the problem you described would not arise.

JayjeetAtGithub:
@ivotron Please review this. The proper functioning of this feature depends on #653. Thanks.

cli/popper/gha.py
```diff
@@ -537,10 +559,114 @@ def singularity_start(self):
...
+            log.fail('Action {} failed!'.format(self.action['name']))
+
+        cmd = []
+        if self.action.get('runs', None):
+            cmd.extend(self.action['runs'])
+        if self.action.get('args', None):
+            cmd.extend(self.action['args'])
+        self.vagrant_ssh(' '.join(cmd))
+
+    def setenv(self, k, v):
```
ivotron: looks like this is not used, right? if it isn't, can we remove it?

JayjeetAtGithub: > Also, we need to deal with sharing $HOME and $GITHUB_WORKSPACE folders.

Can you please help me on how to pass these env vars inside the vagrant vm automatically? I was thinking of making a template Vagrantfile and populating it during `vagrant_init()`. Is there any other way round ?

ivotron: I do not think there is a workaround it. We will need to dynamically inject this into the Vagrantfile.

JayjeetAtGithub: Yeah, thanks

Issue #642, opened 2019-05-07
cli: add singularity installation as part of CI config file generation


ivotron:
Currently, config files generated for CI services do not include instructions for installing singularity. We can add a `--with-singularity` flag to the `popper ci` command to include this

PR #645, opened 2019-05-13
Adds singularity install support to ci

JayjeetAtGithub:
Addresses #642

### DIFF
--- a/cli/popper/commands/cmd_ci.py
+++ b/cli/popper/commands/cmd_ci.py
@@ -123,8 +123,9 @@ def cli(ctx, service, with_singularity):
         if with_singularity:
             if service in ['jenkins', 'gitlab']:
                 log.fail(
-                    'Scaffolding of singularity install script not supported for '
-                    'Jenkins and Gitlab CI. Please do it manually.')
+                    'Scaffolding of Singularity install script is not '
+                    'supported for Jenkins and Gitlab CI. Include it '
+                    'manually depending upon the CI\'s OS.')
             scripts.append(install_scripts['singularity'])
 
         if scripts:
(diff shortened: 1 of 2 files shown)


cli/popper/commands/cmd_ci.py
```diff
@@ -5,6 +5,11 @@
 from popper import scm, utils as pu
 from popper.cli import pass_context, log
 
+install_scripts = {
+    'singularity': """wget http://neurodeb.pirsquared.org/\
+pool/main/s/singularity-container/\
+singularity-container_2.6.1-2~nd16.04+1_amd64.deb && \
+sudo apt-get -f install ./singularity-container_2.6.1-2~nd16.04+1_amd64.deb"""}
```
ivotron: this is only going to work on Travis and probably CircleCI, as those are the only CI workers we know it has ubuntu/debian installed. In the other cases, we either don't know (Jenkins) or the OS is not debian-based (Gitlab). So what we can do is only enable this for Circle/Travis and show an error message on Jenkins/Gitlab.

cli/popper/commands/cmd_ci.py
```diff
@@ -106,8 +120,18 @@ def cli(ctx, service):
...
 
+        scripts = []
+        # customize content
+        if with_singularity:
+            scripts.append(install_scripts['singularity'])
+
+        if scripts:
+            scripts = ' && '.join(scripts)
+        else:
+            scripts = 'echo Nothing to install !'
```
ivotron: this might confuse users. I'd leave the else block out

JayjeetAtGithub:
@ivotron Please review ! Thanks

Issue #625, opened 2019-04-28
cli: add option of running an action and all of its dependencies


ivotron:
Currently, when `popper run` is given an action argument, it will only run the specified action. We can add a `--with-dependencies` flag that instructs popper to executes all the dependencies of the given action as well.

PR #650, opened 2019-05-15
Adds --with-dependencies flag

JayjeetAtGithub:
Addresses #625

### DIFF
--- a/ci/test/actions-demo
+++ b/ci/test/actions-demo
@@ -38,9 +38,38 @@ cat <<EOF > main.workflow
 workflow "clone" {
   resolves = "test"
 }
+
+action "dependency four" {
+  uses = "docker://busybox"
+  runs = ["ls", "-ltr"]
+}
+
+action "dependency three" {
+  uses = "docker://busybox"
+  runs = ["ls", "-ltr"]
+}
+
+action "dependency two" {
+  needs = ["dependency three", "dependency four"]
+  uses = "docker://busybox"
+  runs = ["ls", "-ltr"]
+}
+
+action "dependency one" {
+  needs = ["dependency two"]
+  uses = "docker://busybox"
+  runs = ["ls", "-ltr"]
+}
+
 action "test" {
+  needs = ["dependency one"]
   uses = "https://gitlab.com/barbaragd/action-test@master"
 }
 EOF
 
-popper run
+# popper run
+
+# test --with-dependencies flag
+popper run test
+popper run --with-dependencies test
+popper run --with-dependencies 'dependency one'
--- a/ci/test/actions-demo
+++ b/ci/test/actions-demo
@@ -115,3 +115,45 @@ popper run --with-dependencies a2
 popper run --with-dependencies b
 popper run --with-dependencies c
 popper run --with-dependencies d
+
+cat <<EOF > main.workflow
+workflow "wf" {
+  resolves = "d"
+}
+
+action "a1" {
+  uses = "sh"
+  args = "ls"
+}
+
+action "a2" {
+  uses = "sh"
+  args = "ls"
+}
+
+action "b" {
+  needs = ["a1", "a2"]
+  uses = "sh"
+  args = "ls"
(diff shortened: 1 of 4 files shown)


JayjeetAtGithub:
This PR is not ready for review.

JayjeetAtGithub:
@ivotron Please review. Thanks !

ci/test/actions-demo
```diff
@@ -38,9 +38,38 @@ cat <<EOF > main.workflow
...
+  needs = ["dependency one"]
   uses = "https://gitlab.com/barbaragd/action-test@master"
 }
 EOF
 
 popper run
+
+# test --with-dependencies flag
+popper run test
+popper run --with-dependencies test
```
ivotron: can we check the output in order to ensure that only dependent actions get executed? Also, we can skip the previous `popper run test` as that functionality is being checked somewhere else

cli/popper/gha.py
```diff
@@ -157,7 +157,7 @@ def instantiate_runners(self):
                     a, self.workspace, self.env,
                     self.dry_run)
 
-    def run(self, action_name=None, reuse=False, parallel=False):
+    def run(self, action_name=None, with_dependencies=False):
```
ivotron: we can simplify the implementation of this feature by extending the [`Workfow.get_stages()`](https://github.com/systemslab/popper/blob/master/cli/popper/parser.py#L67) function so that it can be given an action where to start generating stages. In this way, the logic of running a workflow stays the same (i.e. no modifications would be needed).

ci/test/actions-demo
JayjeetAtGithub: @ivotron Are we required to match the output with a hardcoded output ? If thats the case, we might use `--dry-run` flag.

JayjeetAtGithub:
@ivotron Please have a look at the approach.

ci/test/actions-demo
ivotron: yeah, using `--dry-run` would speedup the execution. No need to compare to hardcoded strings though. We could run with and without `--with-dependencies` and ensure that the latter is including dependent actions in the execution

JayjeetAtGithub: @ivotron On running without `--with-dependencies`, the actions defined in the `needs` attribute of the referenced action will get executed ?

ivotron: no, only the given action

cli/popper/gha.py
```diff
@@ -184,17 +185,21 @@ def run(self, action_name=None, reuse=False, parallel=False):
         self.download_actions()
         self.instantiate_runners()
 
-        if action_name:
-            self.wf.get_runner(action_name).run(reuse)
+        if self.action_name:
+            if self.with_dependencies:
```
ivotron: looks like this condition is doing the same as the outer `else` below (lines 195-196), right? Can we instead remove it?

cli/popper/parser.py
```diff
@@ -79,25 +79,35 @@ def get_stages(self):
                         'next', set()))
             current_stage = next_stage
 
-    def complete_graph(self):
+    def complete_graph(self, action_name=None):
```
ivotron: can we update the docstring so it explains the purpose of this new parameter? Is this new parameter restricting how deep the completion goes? I feel like the code is a bit difficult to parse. 

Conceptually, it might be more readable to let the `complete_graph` work on the complete graph and then implement any sort of filtering or ad-hoc graph traversal separately. So in this case, we could leave the `complete_graph` function as it was before, and instead add an `action` parameter to the `get_stages` method that indicates that we want to traverse the graph from the root to the given action (as opposed to going from the root to the end of the graph), ignoring any nodes that are not a direct dependency to this given action

JayjeetAtGithub: Okay

JayjeetAtGithub:
@ivotron Please have a look !

ivotron:
thanks a lot!

Issue #626, opened 2019-04-28
cli: add option of skipping actions


ivotron:
Add ability in `popper run` of skipping one or more actions (via a `--skip` flag). This will run all the workflow, except the given list of actions. If, after removing all the skipped actions, the remaining workflow contains a "gap" (i.e. an action that depends on one of the skipped actions), then `popper run` needs to throw an error notifying users about it.

JayjeetAtGithub:
@ivotron Are we required to pass a `,` separated list of action names to the `--skip` flag?
Ex : `popper run --skip a,b,c`

ivotron:
yeah, that would be useful. The other alternative would be to allow `--skip` multiple times, but I'm not sure if click supports that (I believe it does).

Issue #622, opened 2019-04-28
cli: show message when no action will get executed


ivotron:
To reproduce:

```hcl
workflow "test" {
  on = "push"
  resolves = "foo"
}

action "bar" {
  uses = "actions/bin/sh@master"
  runs = ["ls", "-l"]
}
```

the above does not execute any action. However, `popper run` still goes through it. Instead, we would like to notify users that no action will get executed.

Issue #621, opened 2019-04-28
cli: check existence of referenced actions


ivotron:
We need to check that actions referenced in a `uses` or `resolves` attributes are defined. Currently, when an action is referenced and it is not defined, a python stacktrace is shown. This workflow reproduces this behavior:

```hcl
workflow "test" {
  on = "push"
  resolves = "foo"
}

action "bar" {
  uses = "actions/bin/sh@master"
  runs = ["ls", "-l"]
}

action "bar" {
  needs = "undefined"
  uses = "actions/bin/sh@master"
  runs = ["ls", "-l"]
}
```

we can show a friendlier message.

PR #654, opened 2019-05-20
Adds --skip flag

JayjeetAtGithub:
Addresses #626 , #621, #622

### DIFF
--- a/ci/test/skip
+++ b/ci/test/skip
@@ -1,0 +1,127 @@
+#!/bin/bash
+set -ex
+# shellcheck source=./ci/test/common
+source ./common
+init_test_repo
+cd "$test_repo_path"
+
+cat <<EOF > main.workflow
+workflow "wf" {
+  resolves = "d"
+}
+
+action "a1" {
+  uses = "sh"
+  args = "ls"
+}
+
+action "a2" {
+  uses = "sh"
+  args = "ls"
+}
+
+action "b" {
+  needs = ["a1", "a2"]
+  uses = "sh"
+  args = "ls"
+}
+
+action "c" {
+  needs = ["a1", "a2"]
+  uses = "sh"
+  args = "ls"
+}
+
+action "d" {
+  needs = ["b", "c"]
+  uses = "sh"
+  args = "ls"
+}
+EOF
+
+popper run a1
+popper run a2
+popper run b
+popper run c
+popper run d
+
+! popper run --skip e1
+! popper run --skip a1 --skip a2
+popper run --skip a1 --dry-run
+popper run --skip a2 --dry-run
+! popper run --skip b --skip c
+popper run --skip b --dry-run
+popper run --skip c --dry-run
+popper run --skip d --dry-run
+
+cat <<EOF > main.workflow
+workflow "wf" {
+  resolves = "wrong"
+}
+
+action "a1" {
+  uses = "sh"
+  args = "ls"
+}
+
+action "a2" {
+  uses = "sh"
+  args = "ls"
+}
+
+action "b" {
+  needs = ["a1", "a2"]
+  uses = "sh"
+  args = "ls"
+}
+
+action "c" {
+  needs = ["a1", "a2"]
+  uses = "sh"
+  args = "ls"
+}
+
+action "d" {
+  needs = ["b", "c"]
+  uses = "sh"
+  args = "ls"
+}
+EOF
+
+! popper run
+
+cat <<EOF > main.workflow
+workflow "wf" {
+  resolves = "d"
+}
+
+action "a1" {
+  uses = "sh"
+  args = "ls"
+}
+
+action "a2" {
+  uses = "sh"
+  args = "ls"
+}
+
+action "b" {
+  needs = ["a1", "a2"]
+  uses = "sh"
+  args = "ls"
+}
+
+action "c" {
+  needs = ["a1", "a2"]
+  uses = "sh"
+  args = "ls"
+}
+
+action "d" {
+  needs = ["b1", "c"]
+  uses = "sh"
+  args = "ls"
+}
+EOF
+
+! popper run
--- a/cli/popper/gha.py
+++ b/cli/popper/gha.py
@@ -3,6 +3,7 @@ import os
 import shutil
 import subprocess
 import multiprocessing as mp
+from copy import deepcopy
 from builtins import dict, input, str
 from distutils.dir_util import copy_tree
 from distutils.spawn import find_executable
@@ -27,26 +28,14 @@ class WorkflowRunner(object):
     """A GHA workflow runner.
     """
 
-    def __init__(self, wfile, workspace, dry_run,
-                 reuse, parallel, skip_secrets_prompt=False):
+    def __init__(self, workflow):
 
-        wfile = pu.find_default_wfile(wfile)
-
-        self.workspace = workspace
-        self.dry_run = dry_run
-        self.reuse = reuse
-        self.parallel = parallel
-        self.skip_secrets_prompt = skip_secrets_prompt
         self.actions_cache_path = os.path.join('/', 'tmp', 'actions')
-
-        # Initialize a Worklow. During initialization all the validation
-        # takes place automatically.
-        self.wf = Workflow(wfile)
-        self.check_secrets()
+        self.wf = workflow
         log.debug('workflow:\n{}'.format(
             yaml.dump(self.wf, default_flow_style=False, default_style='')))
 
-    def check_secrets(self):
+    def check_secrets(self, wf, dry_run, skip_secrets_prompt):
         """Checks whether the secrets defined in the action block is
         set in the execution environment or not.
 
@@ -56,9 +45,9 @@ class WorkflowRunner(object):
             else it prompts the user to enter the environment vars
             during the time of execution itself.
         """
-        if self.dry_run or self.skip_secrets_prompt:
+        if dry_run or skip_secrets_prompt:
             return
-        for _, a in self.wf.actions:
+        for _, a in wf.actions.items():
             for s in a.get('secrets', []):
                 if s not in os.environ:
                     if os.environ.get('CI') == "true":
@@ -67,12 +56,12 @@ class WorkflowRunner(object):
                         val = input("Enter the value for {0}:\n".format(s))
                         os.environ[s] = val
 
-    def download_actions(self):
+    def download_actions(self, wf, dry_run):
         """Clone actions that reference a repository."""
         cloned = set()
         infoed = False
 
-        for _, a in self.wf.actions:
+        for _, a in wf.actions.items():
             if ('docker://' in a['uses'] or 'shub://' in a['uses'] or
                     './' in a['uses'] or a['uses'] == 'sh'):
                 continue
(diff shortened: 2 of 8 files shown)


JayjeetAtGithub:
@ivotron Please review this.

cli/popper/parser.py
```diff
@@ -92,9 +94,14 @@ def complete_graph(self):
...
 
             for n in a_block.get('needs', []):
-                if not self.workflow['action'][n].get('next', None):
-                    self.workflow['action'][n]['next'] = set()
-                self.workflow['action'][n]['next'].add(name)
+                try:
+                    if not self.workflow['action'][n].get('next', None):
+                        self.workflow['action'][n]['next'] = set()
+                    self.workflow['action'][n]['next'].add(name)
+                except KeyError:
```
ivotron: we can avoid catching the exception by doing the removal of skipped tasks to the end of the constructor (i.e. invoke `remove_skipped_actions` after `complete_graph`). When we do that, we can check whether the removal is leaving the graph in an "unrunnable" state and raise the exception notifying about it

JayjeetAtGithub:
@ivotron  Please have a look. I have made the changes.

ci/test/skip
```diff
@@ -0,0 +1,74 @@
...
+        "curl bitbucket",
+        "curl medium",
+        "curl youtube",
+        "curl twitter",
+    ]
+}
+EOF
+
+! popper run --skip 'curl xyz'
+! popper run --skip 'curl google' --skip 'curl facebook'
```
ivotron: Sorry @jayjeetatgithub, I think I did not describe the definition of a "gap" properly. A gap is present when an action depends on a previous stage (one or more actions), but no actions for that stage are available in the workflow. So for example, for this workflow:

```hcl
workflow "wf" {
  resolves = "d"
}

action "a" {
  uses = "sh"
}

action "b" {
  needs = "a"
  uses = "sh"
}

action "c" {
  needs = "a"
  uses = "sh"
}

action "d" {
  needs = ["b", "c"]
  uses = "sh"
}
```

If we say `popper run --skip b` the workflow should run OK. In this case, we remove `b` from the workflow, and remove it as a dependency from `d` (and as a `next` item in `a`'s list of forward edges). This also should work if we do `popper run --skip c`.

If we skip `b` and `c` though, an error will be thrown because `a` cannot reach `d` (because there is a gap).

JayjeetAtGithub: Thanks a lot. @ivotron.

JayjeetAtGithub: @ivotron Can we place the `remove_skip_actions()` just before `complete_graph()` ? In this way we can leverage `complete_graph()`'s capability to find out unreachable actions. Does this sound correct ?

ivotron: the goal of the `complete_graph` function is to add forward-going edges and the root. We can add another `check_for_unreachable_nodes` function, or something alike. The goal being to have a function do a single thing to improve readability and maintainability (see [my comment](https://github.com/systemslab/popper/pull/650#discussion_r286936204) on #650).

JayjeetAtGithub: Thanks @ivotron.

JayjeetAtGithub: @ivotron  Can the order of calling the functions be like this ? 
```
self.remove_skip_actions()
self.complete_graph()
self.check_for_unreachable_nodes()`
```
In this way, we first remove the actions to be skipped, then complete the graph, and finally check for gaps? Otherwise if `remove_skip_actions()` is after `complete_graph()`, we have to again manipulate the graph.

ivotron: Sorry @JayjeetAtGithub, I should have expressed better what I have in mind.

The idea of running `complete_graph` first is to have an in-memory representation of what the `.workflow` file contains. It would simplify the code if we can have the `Workflow` object be immutable after it is constructed. Then we can apply distinct operations to the graph that return new workflow objects, for example:

 1. filter all actions but one. This is what we do when we are given an action argument.
 2. get the path from the root to a particular action (used when `--with-dependencies` is given).
 3. remove nodes (when `--skip` is given)

Each of the above can be implemented in such a way that they return a new workflow object that is the result of applying the operation to the complete workflow. This would be a functional-ish approach to manipulating the graph, as a way of making it more readable/maintainable. Then we can add other functions that do things like checking for unreachable nodes, node generators (to support distinct ways of traversing the graph), etc. 

The idea is to make the `Workflow` object a generic abstraction that can potentially be reused in multiple scenarios. For example, running the workflow multiple times (but with distinct objectives each time) in a single `popper run` invocation; e.g. run just an action first, then a particular path, then the entire workflow, etc. If we write things in such a way that the original workflow data is lost (e.g. by permanently removing nodes out of the graph), then we won't be able to implement this. Another scenario would be when popper is used as a library (e.g. in a web service), where it would be undesirable to have to read a `.workflow` file multiple times given that the file has not been modified.

So in summary, we can refactor the code so that:

  * the `Workflow`'s constructor only builds the graph, as given in the `.workflow` file.
  * the above means we need to refactor `complete_graph` so that it only completes the graph and does not do any check for unreachable nodes. 
  * add operations for the three things mentioned above
  * add a `check_for_unreachable_nodes`
  * any other thing I'm missing

what do you think?

JayjeetAtGithub: @ivotron Sorry for giving you all the trouble. I now understand what needs to be done completely.
I would merge #621, #622 in this pr. Thanks a lot !

JayjeetAtGithub: @ivotron Does removal of `a` block in the above workflow, leads to unreachable node `b`, `c`,`d` ?

JayjeetAtGithub: @ivotron Also, currently we show a warning for unreachable actions. So would that  case be changed to a fail, and occur through `check_unreachable_nodes()`?

JayjeetAtGithub: @ivotron I have done some refactoring according to your review, and solved issues #621, #626, #622 through this pr. Please review. Thanks.

cli/popper/parser.py
```diff
@@ -198,53 +203,110 @@ def normalize(self):
...
-        for line in self.workflow_content:
+        for line in self._workflow_content:
             line = line.strip()
             if line.startswith('action '):
                 acount += 1
         if parsed_acount != acount:
             log.fail('Duplicate action identifiers found.')
+
+
+def validate_workflow(workflow):
```
ivotron: this can also be a member function, or is there a particular reason this is not a member function? Also, `validate_workflow` might be a bit ambiguous. We could rename this to `find_unreachable_actions` or `check_for_unreachable_actions`

ci/test/skip
```diff
@@ -0,0 +1,121 @@
...
+}
+
+action "d" {
+  needs = ["b1", "c"]
+  uses = "sh"
+  args = "ls"
+}
+EOF
+
+! popper run
```
ivotron: we could `grep` the output to see if the command is failing due to the reasons we expect it should fail for.

ci/test/skip
ivotron: awesome, thanks a lot!! I've left a review

ivotron: > @ivotron Also would the new functions like `remove_skip_actions()`, `check_unreachable_nodes()` etc that would be implemented be part of `Workflow` class, or seperate functions ? In the first case, the function call would look something like `new_wf = old_wf.remove_skip_actions()` and in the second one it will be like `new_wf = remove_skip_actions(old_wf)`.
> 
> Which approach would be better ?

@JayjeetAtGithub sorry, I missed this question. I'd vote for the latter (i.e. as static members of the class).

cli/popper/parser.py
```diff
@@ -198,53 +203,110 @@ def normalize(self):
...
+    _traverse(workflow.root, visited, workflow)
+
+    unreachable = actions - visited - skipped
+    if unreachable:
+        log.warn('Actions {} are unreachable.'.format(', '.join(unreachable)))
+
+
+def skip_actions(workflow, skip_list):
+    """Removes the actions to be skipped from the workflow graph and
+    return a new `Workflow` object.
```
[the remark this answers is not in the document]
JayjeetAtGithub: Currently it looks like the workflow was getting updated from debugging when `popper run --skip b` is ran.
```
Before:  dict_items([('a1', {'uses': 'sh', 'args': ['ls'], 'name': 'a1', 'next': {'b', 'c'}}), ('a2', {'uses': 'sh', 'args': ['ls'], 'name': 'a2', 'next': {'b', 'c'}}), ('b', {'needs': ['a1', 'a2'], 'uses': 'sh', 'args': ['ls'], 'name': 'b', 'next': {'d'}}), ('c', {'needs': ['a1', 'a2'], 'uses': 'sh', 'args': ['ls'], 'name': 'c', 'next': {'d'}}), ('d', {'needs': ['b', 'c'], 'uses': 'sh', 'args': ['ls'], 'name': 'd'})])
After :  dict_items([('a1', {'uses': 'sh', 'args': ['ls'], 'name': 'a1', 'next': {'c'}}), ('a2', {'uses': 'sh', 'args': ['ls'], 'name': 'a2', 'next': {'c'}}), ('b', {'needs': ['a1', 'a2'], 'uses': 'sh', 'args': ['ls'], 'name': 'b', 'next': {'d'}}), ('c', {'needs': ['a1', 'a2'], 'uses': 'sh', 'args': ['ls'], 'name': 'c', 'next': {'d'}}), ('d', {'needs': ['b', 'c'], 'uses': 'sh', 'args': ['ls'], 'name': 'd'})])
```
@ivotron Please verify.
But now when we move it in the `Workflow` class we need to use `deepcopy()`.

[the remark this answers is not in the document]
ivotron: yes, we need to operate on a copy, otherwise the internal dict gets modified when remove skipped actions. Is that what you mean?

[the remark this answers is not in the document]
JayjeetAtGithub: @ivotron  Yeah, we need to stop the original workflow from mutated.

cli/popper/gha.py
```diff
@@ -26,11 +26,10 @@ class WorkflowRunner(object):
     """A GHA workflow runner.
     """
 
-    def __init__(self, wfile, workspace, dry_run,
+    def __init__(self, wfile, skip, workspace, dry_run,
```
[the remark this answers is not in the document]
JayjeetAtGithub: @ivotron So would it be something like this ? Like in `cmd_run.py`,

```
wf = Workflow(wfile) // Initial graph generation and syntax validation
wf_runner = WorkflowRunner(wf) // All workflow manipulations happen here
wf_runner.run() // Run the workflow
```
One doubt, Where to pass all the arguments like `parallel`,  `dry_run`, `reuse` (In step 2 or step 3)?
Thanks!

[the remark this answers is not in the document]
ivotron: I'd vote for passing that info to the WorkflowRunner (2)

ci/test/skip
JayjeetAtGithub: @ivotron Should we warn / fail for unreachable nodes ?

ivotron: fail when `--skip` is passed (as this is likely a result of removing the skipped actions from the workflow); warn in the remaining scenarios.

JayjeetAtGithub:
@ivotron Please have a look !

ivotron:

Issue #623, opened 2019-04-28
cli: regex-based implementation of scm.parse


ivotron:
The logic of the [`scm.parse`](https://github.com/systemslab/popper/blob/450d4e9/cli/popper/scm.py#L137-L193) function is getting a bit too complex to debug. We can implement the same functionality using the [`re`](https://docs.python.org/3/library/re.html) library instead.

PR #660, opened 2019-05-28
Fixes splitting of Git URLs

### LINKED ISSUE #658: cli: unable to run action defined in 3rd level of a repo directory
An action defined in `user/repo/dir@master` runs but if an action is defined as in `user/repo/dir/subdir@master` fails to run.
Popper is only able to run actions from up to the 2nd level directory and not more than that.

ivotron:
fixes #658


ivotron:
closing as this will be solved as part of the work for #623

Issue #657, opened 2019-05-26
cli: improve message for syntax check


ivotron:
Currently, `parser.py` checks for invalid attributes and shows the following message:

```
$ popper run --wfile .ci/main.workflow
ERROR: Invalid attrs found
```

We can improve the message by explicitly showing which attributes are the ones that are not expected. And spell `attrs` as `attributes` to minimize confusion.

PR #662, opened 2019-05-29
Improved message for syntax check

JayjeetAtGithub:
Addresses #657.

### DIFF
--- a/cli/popper/parser.py
+++ b/cli/popper/parser.py
@@ -155,7 +155,8 @@ class Workflow(object):
             wf_block = list(self._workflow['workflow'].values())[0]
             for key in wf_block.keys():
                 if key not in VALID_WORKFLOW_ATTRS:
-                    log.fail('Invalid workflow attribute {} was found.'.format(key))
+                    log.fail(
+                        'Invalid workflow attribute {} was found.'.format(key))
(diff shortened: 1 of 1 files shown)


cli/popper/parser.py
```diff
@@ -155,7 +155,8 @@ def _validate_syntax(self):
             wf_block = list(self._workflow['workflow'].values())[0]
             for key in wf_block.keys():
                 if key not in VALID_WORKFLOW_ATTRS:
-                    log.fail('Invalid attrs found.')
+                    log.fail(
+                        'Invalid workflow attribute {} was found.'.format(key))
```
ivotron: can we single quote the value, i.e. `"Invalid workflow attribute '{}' was found"`?

cli/popper/parser.py
```diff
@@ -168,7 +169,9 @@ def _validate_syntax(self):
             for _, a_block in self._workflow['action'].items():
                 for key in a_block.keys():
                     if key not in VALID_ACTION_ATTRS:
-                        log.fail('Invalid attrs found.')
+                        log.fail(
+                            'Invalid action attribute {} was '
+                            'found.'.format(key))
```
ivotron: same as above

Issue #647, opened 2019-05-14
cli: allow to execute 'popper run' without having a git repo


ivotron:
Based on feedback from @vsoch (discussed in the [singularity mailing list](https://groups.google.com/a/lbl.gov/d/topic/singularity/MznKsRpvvIk/discussion))

JayjeetAtGithub:
@ivotron I was thinking if we allow popper to run without a `.git` folder and take `$PWD` as root, then how to implement the functions like `get_sha()`, `get_ref()`, etc in `scm.py` ?

vsoch:
I don't think you can - if the user chooses to run without git, it's sort of at their own risk, for their own use case. The ref / sha, and other git related bits would most logically return as None.

vsoch:
And you can give the user a nasty warning about how terribly they are for not having a reproducible workflow :P

JayjeetAtGithub:
@vsoch Thanks a lot.

PR #664, opened 2019-05-30
Makes popper independent of .git

JayjeetAtGithub:
Addresses #665, #647

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
 - "3.7"
 services: docker
 before_install:
+- git rebase HEAD $TRAVIS_BRANCH
 - sudo apt-get update
 - wget http://neurodeb.pirsquared.org/pool/main/s/singularity-container/singularity-container_2.6.1-2~nd16.04+1_amd64.deb
 - sudo apt-get -f install ./singularity-container_2.6.1-2~nd16.04+1_amd64.deb
@@ -13,4 +14,7 @@ before_install:
 install:
 - pip install cli/
 script:
-- popper run --recursive
+- export POPPER_TEST_MODE=1
+- popper run
+- export POPPER_TEST_MODE=2
+- popper run
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ python:
 - "3.7"
 services: docker
 before_install:
-- git rebase HEAD $TRAVIS_BRANCH
 - sudo apt-get update
 - wget http://neurodeb.pirsquared.org/pool/main/s/singularity-container/singularity-container_2.6.1-2~nd16.04+1_amd64.deb
 - sudo apt-get -f install ./singularity-container_2.6.1-2~nd16.04+1_amd64.deb
@@ -14,7 +13,7 @@ before_install:
 install:
 - pip install cli/
 script:
-- export POPPER_TEST_MODE=1
+- export POPPER_TEST_MODE=with-git
 - popper run
-- export POPPER_TEST_MODE=2
+- export POPPER_TEST_MODE=without-git
 - popper run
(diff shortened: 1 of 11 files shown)


JayjeetAtGithub:
@ivotron When we return `""` from `scm.get_sha()` method, then in this line in [`gha.py`](https://github.com/systemslab/popper/blob/master/cli/popper/gha.py#L317), the `tag` variable in docker runner remains incomplete and throws error due to `self.env['GITHUB_SHA']` being empty. With what should we replace the return value to handle this situation ?

cli/popper/gha.py
```diff
@@ -313,8 +313,11 @@ def run(self, reuse=False):
 
                 if action_dir:
                     repo_id += '/'
-
-            tag = repo_id + action_dir + ':' + self.env['GITHUB_SHA']
+            
+            if self.env['GITHUB_SHA']:
```
ivotron: the conditional can be removed by doing:

```python
tag = repo_id + action_dir + ':' + self.env.get('GITHUB_SHA', 'na')
```

where `na` is the value assigned with `GITHUB_SHA` is not defined. We could also use `unknown`

JayjeetAtGithub: Oh yes! Thanks.

JayjeetAtGithub:
@ivotron Can you please tell me what would be the best way to run all the tests with and without `.git` ?

JayjeetAtGithub:
@ivotron  Should we fail or warn on finding a detached head? Means I want to know whether it is ok to have a detached head. Currently, I check for a detached head in `init_repo_object` in `scm.py` and fail if one is found. [This](https://github.com/systemslab/popper/blob/master/ci/test/dry-run#L12) line puts the head in detached state and build fails. How to deal with this ? Thanks

.travis.yml
```diff
@@ -6,11 +6,15 @@ python:
...
+- git rebase HEAD $TRAVIS_BRANCH
 - sudo apt-get update
 - wget http://neurodeb.pirsquared.org/pool/main/s/singularity-container/singularity-container_2.6.1-2~nd16.04+1_amd64.deb
 - sudo apt-get -f install ./singularity-container_2.6.1-2~nd16.04+1_amd64.deb
 - pip install git+https://github.com/virtuald/pyhcl.git@0.3.12#egg=pyhcl
 install:
 - pip install cli/
 script:
-- popper run --recursive
+- export POPPER_TEST_MODE=1
```
ivotron: can we rename the values assigned to this, e.g. `gitless` (or `without-git-repository`) instead of 1

ci/test/dry-run
```diff
@@ -9,7 +9,7 @@ cd "$test_repo_path"
 
 git clone https://github.com/cplee/github-actions-demo
 cd github-actions-demo
-git checkout c824a5a
+# git checkout c824a5a
```
ivotron: we can change this create a branch:

```bash
git checkout -b pinned-commit c824a5a
```

cli/popper/scm.py
```diff
@@ -15,14 +15,23 @@ def init_repo_object():
     try:
         repo = git.Repo(search_parent_directories=True)
     except git.exc.InvalidGitRepositoryError:
-        log.fail('Unable to find root of a Git repository.')
+        log.warn('Unable to find root of a Git repository.')
+
+    if repo:
+        if repo.head.is_detached:
+            log.fail('HEAD is in detached state.')
```
ivotron: this would impede popper to run at all on any repo in detached `HEAD` state, right? Instead, we could move this check to another function (e.g. `check_if_detached_head`) that is invoked only when `CI` set is enabled (from `popper run`).

JayjeetAtGithub:
@ivotron Made the changes. I called `check_if_detached_head` from `scm.py` in CI env because if I would call it through `gha.py`, I would naturally skip the check when using the scm functions separately, like in the `get_head_commit()`, etc and run into error.

ivotron:
thanks a lot!

Issue #567, opened 2019-03-26
cli: add 'popper info' command


ivotron:
After #485 is done, implement a `popper info <action>` that displays the readme of the action. The info is read from the metadata cache that is created when `popper search` runs

Issue #495, opened 2019-02-16
cli: invert update behavior for 'popper search'


ivotron:
Since pipelines and stages don't change that frequently, we can update on demand. So now `--skip-update` is `--update-cache` for the `popper search` command.

Issue #485, opened 2019-02-12
cli: add support for actions in `popper search`


ivotron:
Bring back the functionality of the [`popper search`](https://github.com/systemslab/popper/blob/v1.x/cli/popper/commands/cmd_search.py) command so that it is able to index and search actions contained in a pre-defined list of repositories (e.g. all repositories in the [`popperized`](https://github.com/popperized) organization). This list of repositories containing actions is written in the `.popper.yml` file and can be modified by the user

JayjeetAtGithub:
@ivotron Will the `popper search` functionality in popper 2.0 be the same as that of popper 1.x ?

ivotron:
Yes, the idea is to allow users to do `popper search <keyword>` so that they can search for actions by matching the given keyword to the name of an action. An `--include-readme` flag then instructs popper to search also by looking at the content of the README file.

A list of action repos is defined in a `search_sources` attribute of the `.popper.yml` file. For example:

```yaml
search_sources:
- github.com/popperized/zenodo
- github.com/popperized/cloudlab
- github.com/actions/aws
- github.com/actions/gcloud
..
```

We can have a list of repos specified by default. An alternative would be to have that list be defined in a repo, and retrieve it everytime we do a `popper search --update-sources`. This latter can be left out of the scope of this issue (we might want to create a separate issue for that).

PR #673, opened 2019-06-06
Implements search and info subcommand

JayjeetAtGithub:
Addresses #485 , #495 , #567

### DIFF
--- a/cli/popper/commands/cmd_info.py
+++ b/cli/popper/commands/cmd_info.py
@@ -1,0 +1,52 @@
+import sys
+
+import click
+
+from popper import utils as pu
+from popper.cli import log, pass_context
+
+
+@click.argument('action', required=True)
+@click.option(
+    '--full',
+    help='Show full content of the README.',
+    is_flag=True
+)
+@click.option(
+    '--update-cache',
+    help='Update the action metadata cache before looking for info.',
+    is_flag=True
+)
+@click.command('info', short_help='Show details about an action.')
+@pass_context
+def cli(ctx, action, full, update_cache):
+
+    try:
+        org, repo = action.split('/')
+    except ValueError:
+        log.fail('Bad action name. Required format {org}/{repo}')
+
+    metadata = pu.fetch_metadata(update_cache)
+
+    repo_metadata = None
+    if metadata.get(org, None):
+        if metadata[org].get(repo, None):
+            repo_metadata = metadata[org][repo]
+
+    if not repo_metadata:
+        log.fail('Unable to find metadata for given pipeline.')
+
+    if not repo_metadata.get('repo_readme', None):
+        log.fail('This repository does not have a README.md file.')
+
+    if full:
+        log.info(repo_metadata['repo_readme'])
+    else:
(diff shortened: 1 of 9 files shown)


JayjeetAtGithub:
@ivotron Can you please add a secret environment variable `POPPER_GITHUB_API_TOKEN` ?. The previously set token was not working correctly. So, I set mine one but still not working.

ivotron:
done! I've added the token to the travis settings

cli/popper/commands/cmd_search.py
```diff
@@ -0,0 +1,56 @@
...
+from popper.cli import log, pass_context
+
+
+@click.argument(
+    'keywords',
+    required=True
+)
+@click.option(
+    '--update-cache',
+    help=('Skip updating the metadata and search on the local cache.'),
```
ivotron: The message should be something like `Update the metadata cache before searching`

cli/popper/commands/cmd_info.py
```diff
@@ -0,0 +1,52 @@
...
+    is_flag=True
+)
+@click.command('info', short_help='Show details about an action.')
+@pass_context
+def cli(ctx, action, full, update_cache):
+
+    try:
+        org, repo = action.split('/')
+    except ValueError:
+        log.fail('Bad action name. Required format {org}/{repo}')
```
ivotron: here we can follow the same syntax as in the `uses` keyword. This with the goal of being consistent with how actions are referenced in `.workflow` files. This also means that we can support things like: `org/repo/path/to/action`.

And later we could support actions in arbitrary git repos, but we can leave that for later

cli/popper/commands/cmd_info.py
```diff
@@ -0,0 +1,52 @@
...
+
+import click
+
+from popper import utils as pu
+from popper.cli import log, pass_context
+
+
+@click.argument('action', required=True)
+@click.option(
+    '--full',
```
ivotron: we can show the entire README always

docs/sections/cli_features.md
```diff
@@ -98,6 +98,46 @@ Workflow docker-data-science has been added successfully.
...
+
+Additionally, when searching for an action, you may choose to include
+the contents of the readme in your search by using the `--include-readme`
+flag.
+
+Once `popper search` runs, it caches all the metadata related to the search.
+So, to get the latest releases of the actions, you might want to update the
+cache using the `--update-cache` flag.
+
+By default, popper searches for actions from [popperized](https://github.com/popperized)
```
ivotron: we can link to the file here, as the list of default orgs/repos might change over time

cli/popper/commands/cmd_info.py
JayjeetAtGithub: @ivotron We can utilize the `scm.parse()` for this. Since it is not implemented properly now, should I implement this now or fix it with #623 in a separate pr?

ivotron: that'd be awesome

JayjeetAtGithub: Thanks ! I will send the PR for #623 soon along with this fix.

PR #674, opened 2019-06-07
Implements popper search and info subcommands

JayjeetAtGithub:


cli/popper/resources/.search_sources.yml
```diff
@@ -0,0 +1,6 @@
+organizations:
+- popperized
+- actions
+repositories:
```
ivotron: we can add the repos that are currently listed [here](https://github.com/boyney123/github-actions/tree/master/src/actions).

JayjeetAtGithub:
@ivotron For accessing the `search_sources.yml` from code, we might need to push [this](https://github.com/systemslab/popper/pull/674/files#diff-9060895597f65f12558c13f62bc67c7e) to master before. Can i go ahead with this?

PR #676, opened 2019-06-08
Changes the way 'popper add' copies files

### LINKED ISSUE #675: cli: popper add not downloading dependencies
For example:

```
popper add popperized/popper-examples/workflows/cloudlab-iperf
```

only copies `main.workflow` file, but not the rest of files. In general, all the files/folders that contained in the folder where the `.workflow` file is contained should be downloaded

ivotron:
fixes #675


JayjeetAtGithub:
Thanks, @ivotron. Now the copy part is more general and everything looks clean.

Issue #683, opened 2019-06-14
cli: add ability to run without network connectivity


ivotron:
Add an `--offline` flag to `popper run` that instructs popper to skip cloning repositories (assume that they have been cloned previously).

JayjeetAtGithub:
@ivotron Some commands like `popper add` , `popper search` will not work in offline condition, so besides showing an warning, do we need to handle the connection failure errors on these commands ?

JayjeetAtGithub:
Also,
> skip cloning repositories

This also applies to skip pulling docker images if not found locally i guess ?

ivotron:
> @ivotron Some commands like `popper add` , `popper search` will not work in offline condition, so besides showing an warning, do we need to handle the connection failure errors on these commands ?

these can be left as they are, that is, they'll raise networking-related exceptions.

> > skip cloning repositories
> 
> This also applies to skip pulling docker images if not found locally i guess ?

yeah, that also applies in that case. Based on your comment, I was thinking that `--offline` might be misleading. So maybe we could instead add `--no-clone`/`--no-pull` (or `--skip-clone`/`--skip-pull`), what do you think?

JayjeetAtGithub:
@ivotron Yeah, this sounds great, as while using this command, user becomes self aware whether the dependencies are present offline or not.

PR #685, opened 2019-06-15
Add --skip-clone and --skip-pull flag

JayjeetAtGithub:
Addresses #683 .

### DIFF
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -174,12 +174,12 @@ def run_pipeline(action, wfile, skip_clone, skip_pull, skip, workspace, reuse,
                  "present locally.")
 
     try:
-        pipeline.run(action, skip_clone, skip_pull, skip, workspace, reuse, dry_run,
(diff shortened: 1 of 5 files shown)


cli/popper/commands/cmd_run.py
```diff
@@ -153,13 +165,21 @@ def run_pipeline(action, wfile, skip, workspace, reuse,
         log.warn("Using --parallel may result in interleaved output. "
                  "You may use --quiet flag to avoid confusion.")
 
+    if skip_clone:
+        log.warn("Please make sure that all the required action repositories "
+                 "are available locally.")
+
+    if skip_pull:
+        log.warn("Please make sure that all the required images are "
+                 "present locally.")
```
ivotron: these might get too repetitive for someone running with `--skip-pull` or `--skip-clone` over and over again. I'd vote for leaving these warning messages out; you have implemented proper messages that are shown when an image doesn't exist and `--skip-pull` is used, which is informative enough

PR #695, opened 2019-06-22
Fixes #689 and other fixes

### LINKED ISSUE #689: cli: skip downloading dependencies, checking secrets, instantiating runners for unreachable actions.
Consider the following workflow,
```
workflow "sample" {
    resolves = ["reachable"]
}

action "unreachable" {
    uses = "actions/bin/curl@master"
}

action "reachable" {
    uses = "popperized/cmake@master"
}
```

Here popper clones both `popperized/cmake` and `actions/bin/curl` whereas it should only clone `popperized/cmake` as the action using the other one is unreachable and will not be executed. In the same way, the other functions like instantiating runners, checking secrets should be skipped for unreachable actions.

Also, upon passing an unreachable action to `--on-failure` flag, it gets through the 3 functions twice which is not desirable.

JayjeetAtGithub:


ci/test/actions-demo
```diff
@@ -297,3 +297,23 @@ EOF
...
+  args = ["verssion"]
+}
+
+action "cleanup" {
+  uses = "actions/bin/sh@master"
+  args = ["echo Cleaning up..."]
+}
+EOF
+
+popper run --on-failure cleanup
```
ivotron: can we grep the output to see if `Cleaning up...` is being printed? And similarly, in the workflow above where `--on-failure` should also be triggered?

PR #698, opened 2019-06-27
Replace most invalid docker name characters with underscores

malept:
This fixes the case where GitHub Action names have, for example, parentheses in them.


ivotron:
thanks a lot!

Issue #697, opened 2019-06-25
cli: Add support for choosing between runtimes


JayjeetAtGithub:
Add a `--runtime` flag which would take values like docker, singularity to specify the runtime in which to execute the workflow.

Issue #694, opened 2019-06-20
cli: allow reusing a step, even if its attributes change


ivotron:
currently, when `--reuse` is given to `popper run`, the previously instantiated container is just restarted, which means that any change to the step attributes (`args`, `command`, `env`, etc.) are ignored since they are used at creation time (when `docker create` is issued) and not at execution time (when `docker run` is invoked). We can add the ability of reusing the state of a container, even in the case where `step` attributes have changed, by committing the container first, generating a new image, and instantiating a new container out of this image.

Issue #670, opened 2019-06-03
cli: bindings not correctly set in docker


ivotron:
building an action that uses docker internally results in:

```
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
ERROR: Action 'docker build' failed!
```

ivotron:
Running an action that bind-mounts folders (e.g. an action using docker-compose) fails. We need to rework the way we configure the bindings a bit differently:

  * bind `$HOME:$HOME` and `$HOME:/github/home`
  * define `$HOME` variable with `$HOME` as its value
  * bind `$PWD:$PWD` and `$PWD:/github/workspace`
  * define `$GITHUB_WORKSPACE` variable with `$PWD` as its value

the above supports both (1) actions that need to bind-mount the workspace (or home folder) in another container (by referring to it as `$HOME` or `$GITHUB_WORKSPACE); and (2) we create the actual paths for `/github/workspace` and `/github/home`

JayjeetAtGithub:
> bind $HOME:$HOME and $HOME:/github/home

@ivotron Does this mean that 2 directories `$HOME` and `/github/home` will be pointed mounted to `$HOME` on the host?

ivotron:
yes

JayjeetAtGithub:
@ivotron Sorry, forgot onething, can you tell me how to  reproduce the error ?

ivotron:
with the change in 40d1abf, we don't experiment errors now. The main issue is that `/github/home` and `/github/workspace` are not currently available inside containers. We need to figure out how to bindmount them using `docker-py`

JayjeetAtGithub:
@ivotron Thanks

PR #699, opened 2019-06-27
Adds support to run actions in specified runtime.

JayjeetAtGithub:
Addresses #697, #670, #694 , #678


ci/test/add
```diff
@@ -3,7 +3,7 @@ set -ex
 
 function cleanup {
   cd /tmp
-  rm -rf test_folder/
+  sudo rm -rf test_folder/
```
ivotron: instead of running with sudo, we can delete via docker (i.e. bind-mount /tmp, then run `rm -r` inside the container), as it is done in the [`test/common.sh`](https://github.com/systemslab/popper/blob/master/ci/test/common#L6) script. BTW, we can refactor that and reuse that function

ci/cli.workflow
```diff
@@ -118,17 +118,17 @@ action "end" {
     "test ci",
     "test dot",
     "test dry-run",
-    "test interrupt",
+    // "test interrupt",
```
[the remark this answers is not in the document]
JayjeetAtGithub: This is failing for singularity in python2.7 some times, So I commented it out to address it later.

ci/cli.workflow
```diff
@@ -118,17 +118,17 @@ action "end" {
...
     "test quiet",
     "test recursive",
     "test reuse",
     "test scaffold",
-    "test singularity",
     "test validate",
     "test skip",
     "test sh",
     "test search",
+    "test runtime",
```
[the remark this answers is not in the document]
JayjeetAtGithub: I added this test to test for the spack workflow. We can add other complex workflows to it to test real life workflows.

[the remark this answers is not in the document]
ivotron: I see. Then we could rename the name of the script to avoid confusion on what the test is checking

.travis.yml
```diff
@@ -4,16 +4,16 @@ language: python
 python:
 - "2.7"
 - "3.7"
+env:
+- RUNTIME=singularity POPPER_TEST_MODE=with-git
+- RUNTIME=singularity POPPER_TEST_MODE=without-git
+- RUNTIME=docker POPPER_TEST_MODE=with-git
+- RUNTIME=docker POPPER_TEST_MODE=without-git
```
ivotron: since `with-git` and `without-git` test the same functionality regardless of the runtime, we can drop lines 9-10

.travis.yml
```diff
@@ -4,16 +4,16 @@ language: python
...
 services: docker
 before_install:
-- sudo apt-get update
-- wget http://neurodeb.pirsquared.org/pool/main/s/singularity-container/singularity-container_2.6.1-2~nd16.04+1_amd64.deb
-- sudo apt-get -f install ./singularity-container_2.6.1-2~nd16.04+1_amd64.deb
+- scripts/install_singularity.sh
```
ivotron: can we move this to `ci/scripts`?

scripts/install_singularity.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -ex
+
```
ivotron: we can also check that if `RUNTIME` is not `singularity`, we don't need to do anything

cli/popper/utils.py
```diff
@@ -266,3 +272,41 @@ def fetch_readme_for_repo(user, repo, path_to_action, version):
                        user, repo, version, path_to_action, 'README.md')
     r = make_gh_request(url, err=False)
     return r.text
+
+
+def convert_to_singularityfile(dockerfile, singularityfile):
```
ivotron: since these utility functions are only used by the `SingularityRunner` class, can we move them to be static methods of it?

ivotron:
thanks a lot!

Issue #696, opened 2019-06-25
cli: allow flags in commit message-driven execution


ivotron:
currently, the keyword that allows to specify what workflows to run (in a commit message) has the form:

```
popper:whitelist[...]
```

Now that there are multiple ways in which we can control the execution of a workflow, we can generalize this so that we can take all the same flags that can be given to `popper run`. For example:

```
popper:run[--on-failure cleanup --wfile path/to/main.workflow]
```

we can take more than one of this.

ivotron:
we can also take this opportunity to deprecate the `--wfile` flag, and instead take it as an argument, something like:

```
popper run <options> [action|wfile]
```

JayjeetAtGithub:
```
popper:[--on-failure cleanup --wfile path/to/main.workflow]
```
Does this arrangement apply to only those options which take a workflow/action as value ?

ivotron:
the idea is to allow passing any flag that can be passed to the `popper run` command. We could instantiate another click context by calling the [`make_context()`](https://click.palletsprojects.com/en/7.x/api/?highlight=parse#click.BaseCommand.make_context) method of the `cmd_run` command, so that we don't need to parse this ourselves

JayjeetAtGithub:
@ivotron Now that we deprecated the `popper:whitelist` and `popper:skip`, so if an user want to whitelist 1 workflow he would ideally pass it as an argument to `popper run` by doing `popper run /path/to/workflow`. But if they want to whitelist more than 1 workflow, what should they do ?  For this case, we can allow `skip` to take workflow inputs as well, so that the workflows not required can be skipped. What do you think ?

ivotron:
yeah, that's a good idea

PR #703, opened 2019-07-04
Allow flags for commit message driven execution

JayjeetAtGithub:
Addresses #696

### DIFF
--- a/cli/popper/commands/cmd_run.py
+++ b/cli/popper/commands/cmd_run.py
@@ -118,8 +118,8 @@ from popper import log as logging
 def cli(ctx, **kwargs):
     """Executes one or more pipelines and reports on their status.
     """
-    # Prepare the workflow file list.
     if os.environ.get('CI') == 'true':
+        # When CI is set,
         log.info('Running in CI environment...')
         occurences = parse_commit_message()
         if occurences:
@@ -127,28 +127,29 @@ def cli(ctx, **kwargs):
                 kwargs.update(params)
                 if kwargs['recursive']:
                     log.warn('When CI is set, --recursive is ignored.')
+                    kwargs['recursive'] = False
                 prepare_pipeline(**kwargs)
         else:
-            for workflow in pu.find_recursive_wfile():
-                kwargs['action_wfile'] = workflow
-                prepare_pipeline(**kwargs)
-    else:
-        if kwargs['recursive']:
-            if kwargs['action_wfile']:
-                log.fail(
-                    "An 'action/workflow' argument and --recursive flag "
-                    "cannot be given together.")
-
-            for workflow in pu.find_recursive_wfile():
-                kwargs['action_wfile'] = workflow
-                prepare_pipeline(**kwargs)
-        else:
+            # If no special keyword is found, we run all the workflows,
+            # recursively.
+            kwargs['recursive'] = True
             prepare_pipeline(**kwargs)
+    else:
+        # When CI is not set,
+        prepare_pipeline(**kwargs)
 
 
 def prepare_pipeline(**kwargs):
     """Set parameters for the workflow execution.
     """
+
+    def is_workflow(argument):
+        """Determine whether the action_wfile argument is
+        a workflow ref or an action ref."""
+        if argument:
+            return os.path.isfile(argument) and argument.endswith('.workflow')
+        return False
+
     # Set the logging levels.
     level = 'ACTION_INFO'
     if kwargs['quiet']:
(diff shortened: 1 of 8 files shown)


JayjeetAtGithub:
@ivotron I was using action argument to take both action and workflows, but on moving ahead with that I found that it increased the cyclomatic complexity of the code very much, because of a large number of type checkings, if else, etc. Same was the case with `--skip` taking workflows.

So I was thinking if we could keep the `--wfile` flag intact and just make it  `multiple=True`. Also we can change the action argument to a flag `--action` instead. It would be very helpful, if you could review the current PR. Thanks.

ivotron:
I see. The main problem I see with allowing multiple `--wfile` is that all other arguments (except `--recursive`) assume there's only one workflow being executed. For example `--skip` or `--on-failure`, etc.. 

One alternative we have is to go with the original plan of supporting multiple occurrences of the `popper:run[...]` keyword, and run as many times as this is shown. If we do this, we could deprecate `--wfile` and parse the positional argument with the following criteria: if there's a file with that name, then assume it's a workflow, otherwise assume it's an action.

cli/popper/commands/cmd_run.py
```diff
@@ -15,7 +15,9 @@
 @click.command(
     'run', short_help='Run a workflow or action.')
 @click.argument(
-    'action', required=False)
+    'action_wfile',
```
ivotron: since this is shown in the help, let's name it `target` and then in the help description of the command we explain what this is

cli/popper/commands/cmd_run.py
```diff
@@ -120,124 +112,148 @@
     default=popper.scm.get_git_root_folder()
 )
 @pass_context
-def cli(ctx, action, wfile, skip_clone, skip_pull, skip, reuse,
-        recursive, quiet, debug, dry_run, parallel, log_file,
-        with_dependencies, on_failure, runtime, workspace):
-    """Execute github actions workflows.
+def cli(ctx, **kwargs):
+    """Executes one or more pipelines and reports on their status.
```
ivotron: we can explain here a bit more of what the command does, making references to what the flags do, at a high-level

cli/popper/commands/cmd_run.py
```diff
@@ -120,124 +112,148 @@
...
+            # If no special keyword is found, we run all the workflows,
+            # recursively.
+            kwargs['recursive'] = True
+            prepare_pipeline(**kwargs)
+    else:
+        # When CI is not set,
+        prepare_pipeline(**kwargs)
+
+
+def prepare_pipeline(**kwargs):
```
ivotron: we might want to rename this to `prepare_workflow_execution` to be consistent with terminology

cli/popper/commands/cmd_run.py
```diff
@@ -120,124 +112,148 @@
...
-        log.info("Found and running workflow at " + wfile)
-        run_pipeline(action, wfile, skip_clone, skip_pull, skip, workspace,
-                     reuse, dry_run, parallel, with_dependencies, on_failure,
-                     runtime)
+        run_pipeline(wfile, action, **kwargs)
 
 
-def run_pipeline(action, wfile, skip_clone, skip_pull, skip, workspace, reuse,
-                 dry_run, parallel, with_dependencies, on_failure, runtime):
+def run_pipeline(wfile, action, **kwargs):
```
ivotron: rename to `run_workflow`

cli/popper/commands/cmd_run.py
```diff
@@ -120,124 +112,148 @@
...
 
+    log.info('Found and running workflow at ' + wfile)
     # Initialize a Worklow. During initialization all the validation
     # takes place automatically.
     wf = Workflow(wfile)
     pipeline = WorkflowRunner(wf)
```
ivotron: and this can be renamed to `wf_runner`

cli/popper/commands/cmd_run.py
```diff
@@ -120,124 +112,148 @@
...
-        with_dependencies, on_failure, runtime, workspace):
-    """Execute github actions workflows.
+def cli(ctx, **kwargs):
+    """Executes one or more pipelines and reports on their status.
     """
-    popper.scm.get_git_root_folder()
+    if os.environ.get('CI') == 'true':
+        # When CI is set,
+        log.info('Running in CI environment...')
+        occurences = parse_commit_message()
```
ivotron: can we rename this to `popper_run_occurrences` or `popper_run_instances`, as a way of making it more explicit what this refers to?

cli/popper/commands/cmd_run.py
```diff
@@ -120,124 +112,148 @@
...
+        if occurences:
+            for params in get_params(occurences):
```
ivotron: here we could use `args` instead of `params`

cli/popper/commands/cmd_run.py
```diff
@@ -120,124 +112,148 @@
...
+    else:
+        # When CI is not set,
+        prepare_pipeline(**kwargs)
+
+
+def prepare_pipeline(**kwargs):
+    """Set parameters for the workflow execution.
+    """
+
+    def is_workflow(argument):
```
ivotron: would it be possible to extend this logic so that we also accept a folder name as an argument, so that we can do:

```
popper run workflows/mywf
```

and if there's a `main.workflow` or `.github/main.workflow` file in that folder, then we run it?

JayjeetAtGithub: @ivotron In the case of `github-actions-demo` workflow, if we do `popper run github-actions-demo`, the workflow fails, since we are outside of project root. I guess this is not good. Also .github/main.workflow is present in the root always. So what should be done?

ivotron: oh, I see. In that case then I think it's best to not support passing a folder name

JayjeetAtGithub: Yeah.

PR #706, opened 2019-07-08
Paused reuse for singularity and other modifications.

JayjeetAtGithub:


ci/test/reuse
```diff
@@ -51,7 +51,7 @@ action "show env" {
 }
 EOF
 
-popper run --runtime "$RUNTIME" > output
+popper run > output
```
ivotron: looking forward to when we add other runtimes, we could instead add a conditional to check whether `RUNTIME=singularity` is true and skip accordingly

docs/sections/extensions.md
```diff
@@ -53,27 +53,21 @@ of the `uses` attribute of action blocks.
 > **NOTE**: As part of our roadmap, we plan to add support for Vagrant 
 > and Conda runtimes. Open a [new 
 > issue](https://github.com/systemslab/popper/issues/new) to request 
-> another runtime you would Popper to support.
+> another runtime you would want Popper to support.
```
ivotron: Other couple of changes above are:

  * line 49, the sentence `Popper adds the ability .... ` is not true anymore. Instead, we can say that `Popper adds the ability of running actions in other runtimes by providing a `--runtime` flag to the `popper run` command.

  * on line 54, we need to remove Conda and add Podman instead. And add links to vagrant and podman homepages

PR #711, opened 2019-07-12
Remove recursive and other changes.

JayjeetAtGithub:
Remove `--recursive` option for Popper CLI


cli/popper/commands/cmd_run.py
```diff
@@ -115,35 +115,29 @@
...
-    [TARGET] : It can be either path to a workflow file or an action name.
-    If TARGET is a workflow, the workflow is executed.
-    If TARGET is an action, the specified action from the default workflow
-    will be executed.
+    ACTION : The action to execute from a workflow.
 
     Examples:
 
-    1. When no TARGET argument is passed, Popper will search for the
+    1. When no action argument is passed, Popper will search for the
```
ivotron: strictly speaking these are not examples but a description of what `popper run` does, so we could instead write these in narrative form instead of using a numbered list. Removing `Examples:` and then start a paragraph with

> When no `--wfile` is given, Popper will search for the... If an ACTION is passed, the specified action....

Also, since we don't do `--recursive`, the first sentence can be `Executes a github action workflow.`

cli/popper/commands/cmd_run.py
```diff
@@ -113,37 +113,29 @@
...
-    If TARGET is a workflow, the workflow is executed.
-    If TARGET is an action, the specified action from the default workflow
-    will be executed.
+    """Runs a Github Action Workflow.
 
-    Examples:
+    ACTION : The action to execute from a workflow.
 
-    1. When no TARGET argument is passed, Popper will search for the
+    When no action argument is passed, Popper will search for the
```
ivotron: this should read:

> By default, Popper searches for a workflow in `.github/main.workflow` or `./main.workflow` and it executes it if found.

PR #712, opened 2019-07-15
inject pre/post action

arshul:
Actions can be injected via `--inject-actions-from path/to/workflow` with the workflow having actions named `pre` and/or `post`.


cli/popper/commands/cmd_run.py
```diff
@@ -21,7 +21,7 @@
 @click.option(
     '--debug',
     help=(
-        'Generate detailed messages of what popper does (overrides --quiet)'),
+            'Generate detailed messages of what popper does (overrides --quiet)'),
```
ivotron: can you revert this, so we make pep8 happy?

cli/popper/commands/cmd_run.py
```diff
@@ -56,8 +56,8 @@
 @click.option(
     '--recursive',
     help=(
-        'Run any .workflow file found recursively from current path. '
-        'Ignores flags --on-failure, --skip and --with-dependencies.'
+            'Run any .workflow file found recursively from current path. '
+            'Ignores flags --on-failure, --skip and --with-dependencies.'
```
ivotron: same as above

PR #713, opened 2019-07-15
Add unit tests and bug fixes for scm.py

JayjeetAtGithub:
Unit test for `scm.py` module and other fixes.

### DIFF
--- a/cli/test/test_scm.py
+++ b/cli/test/test_scm.py
@@ -1,0 +1,273 @@
+import os
+import sys
+import shutil
+import unittest
+
+import git
+
+from popper import scm
+
+
+class TestScmWithGit(unittest.TestCase):
+    """Unit tests for popper.scm module in git
+    environment."""
+
+    def setUp(self):
+        if os.path.exists('/tmp/test_folder'):
+            shutil.rmtree('/tmp/test_folder')
+        os.makedirs('/tmp/test_folder')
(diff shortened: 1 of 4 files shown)


JayjeetAtGithub:
@ivotron Can we make the default version returned from `scm.parse` as `master` instead of `None` ?

ivotron:
git repos have a default branch that is pointed to by `refs/remotes/origin/HEAD`, so we could get that when no revision is given

cli/test/test_scm.py
```diff
@@ -0,0 +1,278 @@
...
+            shutil.rmtree('/tmp/test_folder')
+        os.makedirs('/tmp/test_folder')
+        os.chdir('/tmp/test_folder')
+        scm.clone(
+            'https://github.com',
+            'cplee',
+            'github-actions-demo',
+            os.path.join(os.getcwd(), 'github-actions-demo')
+        )
+        shutil.rmtree('/tmp/test_folder/github-actions-demo/.git')
```
ivotron: we can make this depend on `os.environ.get('POPPER_TEST_MODE')`. In this way we can have a single `TestScm` class

JayjeetAtGithub: Yeah, Done

ivotron:
thanks @JayjeetAtGithub !

PR #718, opened 2019-07-17
Fixes #714

### LINKED ISSUE #714: cli: hostrunner not setting up GITHUB_WORKSPACE correctly

JayjeetAtGithub:


cli/popper/gha.py
```diff
@@ -251,6 +251,30 @@ def __init__(self, action, workspace, env, dry_run, skip_pull):
...
+            env.update({s: os.environ[s]})
+
+        for e, v in self.env.items():
+            env.update({e: v})
+
+        env['GITHUB_ACTION'] = self.action['name']
+        env['POPPER_ACTION'] = self.action['name']
+
+        if set_env:
+            for k, v in env.items():
```
ivotron: there's a `self.env` variable (initialized in the constructor) that it is not referenced in this `prepare_environment` function, which I believe contains all the `GITHUB_*` variables. BTW, would it be possible to add a test to verify that this is working as it should?

JayjeetAtGithub: Sure

JayjeetAtGithub: @ivotron [This](https://github.com/systemslab/popper/blob/jayjeet/issue714/cli/popper/gha.py#L260) line refers `self.env` variables.

ivotron: you're right! I missed it

PR #735, opened 2019-08-18
Fixes #707 Adds support for exit code 78

### LINKED ISSUE #707: cli: add support for exit code 78
An action exiting with code 78 interrupts execution of the workflow but does not represent a failure.

JayjeetAtGithub:
Addresses #707

### DIFF
--- a/ci/test/actions-demo
+++ b/ci/test/actions-demo
@@ -409,23 +409,13 @@ workflow "sample" {
   resolves = "end"
 }
 
-action "a" {
-  uses = "docker://debian:buster-slim"
-  args = "whoami"
-}
-
 action "b" {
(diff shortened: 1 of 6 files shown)


ci/test/actions-demo
```diff
@@ -403,4 +403,36 @@ EOF
...
+  uses = "actions/bin/sh@master"
+  args = "whoami"
+}
+
+action "end" {
+  needs = ["a", "b", "c"]
+  uses = "sh"
+  args = "whoami"
+}
+EOF
```
ivotron: this can be simplified if we remove action `a` and `c`. Or is there anything in particular you would like to test with those two?

JayjeetAtGithub: Yeah We can remove them. I actually included them while doing local manual tests.

PR #736, opened 2019-09-01
Fixed dot test

JayjeetAtGithub:
`ci/test/dot` test was failing as the example-aws repo being used by it was deleted. So we are now using cplee/github-actions-demo instead.

### DIFF
--- a/ci/test/dot
+++ b/ci/test/dot
@@ -19,10 +19,6 @@ grep "\"lint\" -> \"branch-filter\";" dot_output
 grep "\"test-and-deploy\" -> \"install\";" dot_output
 grep "\"test\" -> \"branch-filter\";" dot_output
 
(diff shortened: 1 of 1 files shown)


JayjeetAtGithub:
@ivotron Can you please help me figure out the cause of [this](https://travis-ci.org/systemslab/popper/jobs/579377928#L1998) error? This error does not occur in my local machine. Thanks

ivotron:
I can reproduce this on my Linux box. This seems to be related to https://github.com/spack/spack/issues/12485, so this is really an issue that needs to be fixed in https://github.com/popperized/spack (I opened https://github.com/popperized/spack/issues/5).

ivotron:
thanks @JayjeetAtGithub !

PR #740, opened 2019-10-11
Fix failing CI tests

JayjeetAtGithub:
Addresses #741 and adds some more action references to `.search_sources.yml`


ivotron:
thanks a lot !

Issue #738, opened 2019-09-25
cli: add support for running workflows within git submodules


qwofford:
In Popper v1.0, it was possible to launch pipelines which were added to a parent repo as submodules. Popper 2.0+ breaks this. In Popper 2.0, scripts referenced by main.workflow are resolve relative to the parent ".git" directory.

Resolution for this must either assume that the "nearest" `.git` directory above the main.workflow file is the base directory, or the cli must be extended to take a submodule argument. I prefer for former, personally.

qwofford:
https://github.com/systemslab/popper/blob/af9b3d31412192c86bf88bf02e8ad8e6dc5bbfda/cli/popper/scm.py#L14

What about just changing this to False?

qwofford:
https://github.com/systemslab/popper/blob/6448aa33de1ff39d2b9e06fa198c2fd7adff7041/cli/popper/scm.py#L57-L58

Perhaps this is the issue? When I descend into my popper submodule and call git.Repo I see:
```
import git
git.Repo()
<git.Repo "parent_repo_path/.git/modules/experiments/submodule_path">
```

When I call the same command from _above_ the submodule, I get:
```
git.exc.InvalidGitRepositoryError: /autofs/nccs-svm1_home1/qwofford/Projects/tmp/ECP-Artifacts/experiments
git.Repo(search_parent_directories=True)
<git.Repo "parent_repo_path/.git">
```

I think this `get_name()` is snipping the path prematurely. Although `get_name()` references a url, I think it is also used for setting a workflow environment see:

https://github.com/systemslab/popper/blob/442df30129105118e4f09f0b7ded1e8a1b9dc11b/cli/popper/gha.py#L148-L150

qwofford:
@ivotron curious for your thoughts.

ivotron:
thanks a lot for looking into this @qwofford! The issue is that when the `Repo` object is initialized:

https://github.com/systemslab/popper/blob/6448aa33de1ff39d2b9e06fa198c2fd7adff7041/cli/popper/scm.py#L14

as you correctly point out, if we're inside a submodule, the path to the git root is in the `.git/submodule` folder. This is assigned in

https://github.com/systemslab/popper/blob/6448aa33de1ff39d2b9e06fa198c2fd7adff7041/cli/popper/scm.py#L36

What we can do instead is return the value of `repo.working_tree_dir`, as that will return the correct path to both parent and submodule repos

PR #743, opened 2019-10-18
allow popper to run workflows from submodules. issue#728

qwofford:
Resolution for issue https://github.com/systemslab/popper/issues/738


ivotron:
thanks Quincy! can you please rebase master?

PR #744, opened 2019-10-23
colorize logs done and tested for dry-run

sarthak0906:
Logs colourized for the popper run command and tested it with the popper run --dry-run command


sarthak0906:
Hello Everyone, 
I have a doubt regarding the failing tests,
I am using the same colour codes as used in log.py file so as no clashes happen.
Can anyone guide me on how to fix this issue?

ivotron:
hey @sarthak0906 thanks a lot for working on this! overall, I think this feature should be implemented in `log.py`, that is, the changes should not touch `gha.py` (or any other script) and instead leave this to be configured via the logging framework.

Also, we can turn on coloring by default, allowing users to turn it off via a `--no-color` flag, and turn coloring off when the environment variable `CI=true` is defined.  This would also make it easier to test, as we would disable coloring for all tests, except for the test that checks that we're colorizing the output appropriately.

ihth

ivotron:
you can create a new `info` function:

```python
function info(message, action_name="", dryrun=False):
```

in this way we can handle all the coloring stuff inside the function.

ivotron:
actually, the function signature of a `Logger` can handle kwargs, so we can pass this with this attributes without having to define a new function. We only need to query for those key/value pairs in the implementation of the logger functions

sarthak0906:
Hello everyone.
I once again have a few doubts on the failing test on the run command.
The only thing I changed was to add kwargs for msg_prefix and action to add colour.
It would be really helpful if someone could explain what was my error in this commit and how to correct it.

cli/popper/log.py
```diff
@@ -5,6 +5,12 @@
 ACTION_INFO = 15
 logging.addLevelName(ACTION_INFO, 'ACTION_INFO')
 
+# Log Colors
+BOLD_CYAN = '[01;36m'
+RESET = '[0m'
+BOLD_YELLOW = '[01;33m'
+BOLD_RED = '[01;31m'
+
```
sarthak0906: Had to use colors in more than one classes so made them global.

JayjeetAtGithub:
### Please read it in edit mode.

@sarthak0906 The raw output that you are getting from the actions is 
```
Action [[01;33msh[0m] : ran successfully !
```

of this format. But you are comparing this with 

```
Action 'sh' ran successfully !
```

in the bash tests. That's why the tests are failing. You might want to try out comparing the colored logs with the log written in the file `output` like this,  

```
grep "Action \[\[01;33msh\[0m\] : ran successfully !" output
```

I hope this makes some sense. Also once you implement the `--no-color` flag, this will not be a problem anymore. You can use `--no-color` flag while testing.

cli/popper/log.py
```diff
@@ -100,7 +126,21 @@ def info(self, msg='', *args, **kwargs):
         """
         Logs a message with severity 'INFO'
         """
-        super(PopperLogger, self).info(msg, *args, **kwargs)
+        m = ""
```
JayjeetAtGithub: Change the name of variable `m`  to something meaningful.

cli/popper/log.py
```diff
@@ -69,6 +70,31 @@ class PopperLogger(logging.Logger):
     A Logger so that we can add popper fail and action_info log methods
     """
 
+    Color = True
```
JayjeetAtGithub: Use snake case for naming variables and functions.

cli/popper/log.py
```diff
@@ -69,6 +70,31 @@ class PopperLogger(logging.Logger):
...
+
+    def noColor(self):
+        self.Color = False
+        formatter = PopperFormatter(False)
+
+        # INFO/ACTION_INFO goes to stdout
+        h1 = logging.StreamHandler(sys.stdout)
```
JayjeetAtGithub: same goes here. Change `h1` so that it portrays the purpose of it.

cli/popper/log.py
```diff
@@ -69,6 +70,31 @@ class PopperLogger(logging.Logger):
...
+        h1.addFilter(LevelFilter([logging.INFO, ACTION_INFO], False))
+        h1.setFormatter(formatter)
+
+        # anything goes to stdout
+        h2 = logging.StreamHandler(sys.stderr)
```
JayjeetAtGithub: same

cli/popper/log.py
```diff
@@ -46,7 +47,7 @@ class PopperFormatter(logging.Formatter):
         'ACTION_INFO': '%(msg)s',
         'INFO': '%(msg)s',
         'WARNING': '%(levelname)s: %(msg)s',
-        'ERROR': '%(levelname)s: %(msg)s',
+        'ERROR': '%(levelname)s: %(message)s',
```
JayjeetAtGithub: Is there any specific reason of changing `msg` to `message` ?

cli/popper/log.py
```diff
@@ -100,7 +126,21 @@ def info(self, msg='', *args, **kwargs):
...
+                m = msg[0: msg.find("[") +1]
+                m += BOLD_CYAN
+                m += msg[msg.find("[")+1: msg.find("]")]
+                m += RESET
+                m += msg[msg.find("]"): ]
+            else:
+                m = msg
+        else:
+            m = msg
+
```
JayjeetAtGithub: Please check if so much of `else` statements can be avoided.

sarthak0906: done

ivotron:
@sarthak0906 the error with the linter should get resolved after @vipulchhabra99's #749 branch is merged

ivotron:
hi @sarthak0906, can you please rebase master? the issue with improper use of the `!` bash operator has been fixed and merged to master. thanks!

JayjeetAtGithub:
Hi, @sarthak0906. Any progress on this? Can you please get the PR rebased? Thanks.

ivotron:
closing due to inactivity

Issue #700, opened 2019-07-01
cli: add docstrings to Python classes and functions


JayjeetAtGithub:
Some of our code is need of having docstrings added to it. For this issue, we will do a pass to add docstrings to all functions and classes in our codebase

shivangisrivastava0408:
I would like to work on this issue.

vipulchhabra99:
Hi @JayjeetAtGithub I would also like to work on this issue if this issue is still pending.

JayjeetAtGithub:
@vipulchhabra99 Thanks for your interest in Popper. Please go ahead. We follow [`Google`](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) python style guide for documenting the code.

vipulchhabra99:
@JayjeetAtGithub Thanks, Sure I will take care of it.

PR #749, opened 2019-12-12
Added docstrings to classes and functions

vipulchhabra99:
Added the docstrings for the classes and functions.

This is the pr for the issue #700.

### DIFF
--- a/cli/popper/cli.py
+++ b/cli/popper/cli.py
@@ -21,8 +21,11 @@ popper_version = popper_version
 
 
 class PopperCLI(click.MultiCommand):
-
+    """Provides CLI interface for Popper.
+    """
     def list_commands(self, ctx):
+        """Returns the list of available commands in sorted order.
+        """
         rv = []
         for filename in os.listdir(cmd_folder):
             if filename.endswith('.py') and filename.startswith('cmd_'):
@@ -31,6 +34,9 @@ class PopperCLI(click.MultiCommand):
         return rv
 
     def get_command(self, ctx, name):
+        """Imports the command if available in commmands list and provides
+        with most similar commands if the command is not present in the list.
+        """
         try:
             if sys.version_info[0] == 2:
                 name = name.encode('ascii', 'replace')
@@ -69,7 +75,7 @@ flist = None
 
 
 def signal_handler(sig, frame):
-
+    """Handles the interrupt signal."""
     if interrupt_params.get('parallel', None) and flist:
         for future in flist:
             future.cancel()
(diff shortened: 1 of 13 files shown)


ivotron:
@vipulchhabra99 can you please rebase master? the issue with improper use of the `!` bash operator has been merged. thanks!

vipulchhabra99:
@ivotron Yeah, Sure.

ivotron:
many thanks for working on this PR @vipulchhabra99! this is a great start. In order to merge this PR, we need to apply the [Google Python Style convention for docstrings](http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) to all the code, meaning we have to document the arguments (and return behavior) for every method/function, as well as adding docstrings for all classes. For examples you can take a look [here](http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings).

vipulchhabra99:
Hi, @ivotron I have updated the pr with desirable changes, I have left a few variables since I was not able to understand their usage. If still there are more changes or anything that I could have missed please do inform me, I would surely try to complete them.

ivotron:
thank you very much @vipulchhabra99 this is awesome work! Answers to your queries made on Gitter:
  * `ctx` is defined by the click package. We can add a link to the click documentation, for example [this section](https://click.palletsprojects.com/en/7.x/commands/#nested-handling-and-contexts).
  * `wf` is an instance of the `Worfklow` class.
  * `frame` received in `cli.signal_handler()` is described [here](https://docs.python.org/3/library/signal.html#signal.signal) which points to [this link](https://docs.python.org/3/reference/datamodel.html#frame-objects), so we can point to either of these two links.
  * `mod` in `cli.PopperCLI.get_command()` is the `cli` function that any of the `cmd_*.py` files define.
  *  `wid` is a workflow ID produced by a `utils.get_id()`
  *  `r` in `scm.get_default_branch` is a [`git.Repo`](https://gitpython.readthedocs.io/en/stable/tutorial.html#meet-the-repo-type) object

vipulchhabra99:
Thanks a lot, @ivotron, I will update the PR and make desired changes.

ivotron:
the other thing we could add to the `cli.py` file is a link to [this page](https://click.palletsprojects.com/en/7.x/commands/#custom-multi-commands) of the Click documentation, which is where we based the initial CLI skeleton from. That page basically explains the objects defined and the main rationale behind lazily loading command modules.

vipulchhabra99:
Thanks for reviewing @ivotron, I have updated with original links but they are long enough and show pep8 styling errors.

ivotron:

vipulchhabra99:
Thanks @ivotron  for helping me out during this entire process. It was a nice learning experience.

Issue #753, opened 2019-12-19
cli: support for custom arguments flag(--arg) in popper run


realsdx:
Currently `popper run` command doesn't support custom arguments. Custom arguments cloud be really helpful for this type of use case :

```hcl
action "run" {
  uses = "docker://paddlepaddle/paddle:$_TAG"
  args = ["-some", "-args", "-for", "-paddle"]
}
```

and popper could be invoked like this:

```bash
popper run --arg _TAG=1.6.2
```

vipulchhabra99:
@ivotron I would like to work on this but will require your guidance to add this new feature.

ivotron:
thanks a lot for volunteering to work on this @vipulchhabra99 , please go ahead. The idea is to provide functionality to [Google Cloud Build's substitutions feature](https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions). In terms of implementation, this can be done by applying a post-parsing step that substitutes all the argument references for their values. Since this would be done _before_ the workflow gets executed, the workflow execution code shouldn't be affected by this.

PR #758, opened 2020-01-01
implemented --substitutions for custom args with popper run.

vipulchhabra99:
- Currently, the substitutions in uses are only supported.
- If multiple values are assigned to a variable, then the most recent value would be used.
- Multiple substitutions can be made by using different substitutions flag every time
Example:  `popper run --substitutions _variable=v1 --substitutions _variable2=v2`

Pr for #753


JayjeetAtGithub:
@vipulchhabra99 Thanks a lot for working on this. 
> popper run --substitutions _variable=v1,_variable2=v2

We can remove the support for passing multiple key-value pairs through one `--substitution` using `,`, as we want to keep any kind of argument parsing by ourselves to the minimum. We want `click` to handle as much of it as possible.

Also, It would be great if you can write the corresponding unit tests.

vipulchhabra99:
Yeah, sure @JayjeetAtGithub.

JayjeetAtGithub:
@vipulchhabra99 Sounds good. The test that you have written `ci/test/substitution` is an integration test (kind of testing the feature at a high level from a user point of view). The unit tests (where we test every function in an isolated manner) are situated [here](https://github.com/systemslab/popper/tree/master/cli/test). You can see how the unit tests are designed and test the new functions you have implemented in a similar manner. Does that make sense?

After you write the tests, you can test it by doing
```
$ cd popper/cli
$ python setup.py test
```

vipulchhabra99:
Yeah, that makes sense to me. Sorry for misunderstanding it, I will implement them and update the pr soon. Thanks a lot for guiding me :relaxed:

vipulchhabra99:
@ivotron Can you please review this pull request? So that If more changes are required I can make them and fix this issue.

ci/test/substitutions
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+set -ex
+# shellcheck source=./ci/test/common
+source ./ci/test/common
+init_test_repo
+cd "$test_repo_path"
+
+cat <<EOF > main.workflow
+workflow "wf" {
```
ivotron: we can remove a1, a2, c, and d actions from this and the two other workflows, as those are not relevant to this test. In addition, we can also check that substitutions are correctly changed in other parts of the workflow file ( in `env`, in `secrets`, etc..)

ci/test/substitutions
```diff
@@ -0,0 +1,124 @@
...
+  needs = ["b", "c"]
+  uses = "sh"
+  args = "ls"
+}
+EOF
+
+(! popper run --substitutions)
+popper run --dry-run --runtime "$RUNTIME" a1 | grep "a1"
+popper run --dry-run --runtime "$RUNTIME" a2 | grep "a2"
+popper run --dry-run --substitutions _X1=sh --runtime "$RUNTIME" b | grep "b"
```
ivotron: we need to add a check that the substitution was correctly done. This `grep` is only checking the existence of a `b` action. Similarly for the next to lines

cli/test/test_parser.py
```diff
@@ -799,3 +799,190 @@ def test_get_stages(self):
             {'h'},
             {'end'}
         ])
+
+    def test_substitutions(self):
+        self.create_workflow_file("""
+        workflow "example" {
```
ivotron: same comment applies here

cli/popper/parser.py
```diff
@@ -387,3 +387,27 @@ def find_root_recursively(workflow, action, required_actions):
...
+        workflow = deepcopy(wf)
+        substitution_dict = dict()
+
+        for args in substitutions:
+            item = args.split('=')
+            substitution_dict[item[0]] = item[1]
+
+        for key in workflow.parsed_workflow['action']:
+            for var in substitution_dict:
+                uses_str = workflow.parsed_workflow['action'][key]['uses']
```
ivotron: the goal is to support substitutions on all parts of the workflow, not only `uses` attribute of actions.

cli/popper/parser.py
```diff
@@ -387,3 +387,27 @@ def find_root_recursively(workflow, action, required_actions):
...
+                temp_str = uses_str.replace(var, substitution_dict[var])
+                workflow.parsed_workflow['action'][key]['uses'] = temp_str
+
+        for itr in range(len(workflow.workflow_content)):
+            for var in substitution_dict:
+                content_str = workflow.workflow_content[itr]
+                temp_str = content_str.replace(var, substitution_dict[var])
+                workflow.workflow_content[itr] = temp_str
+
```
ivotron: we can also check before returning if all the given substitutions were used to replace values. If some of theme were not used, we throw an error. We could add a `--allow-loose` option as in google cloud build to not throw an error when a substitution is not used.

ci/test/substitutions
vipulchhabra99: Ok, I can fix it by adding the feature to substitute for the whole workflow.

ci/test/substitutions
vipulchhabra99: I thought that if I would enter something that is wrong it will throw some error and that name of the action won't be printed. Rather I can check for the value substituted.

vipulchhabra99:
Hi @ivotron , Thanks a lot for reviewing in addition to the above features you mentioned in comments I have added one extra feature to verify if the substitution variable follows the special rules if they don't follow I have logged the error. Should I retain that in final commit or remove it?

cli/test/test_parser.py
```diff
@@ -799,3 +799,90 @@ def test_get_stages(self):
...
+            resolves = ["b"]
+        }
+
+        action "a" {
+            uses = "_VAR1"
```
ivotron: similarly here, substitutions are `$_VAR1`

ci/test/substitutions
```diff
@@ -0,0 +1,73 @@
...
+init_test_repo
+cd "$test_repo_path"
+
+cat <<EOF > main.workflow
+workflow "wf" {
+  resolves = "b"
+}
+
+action "b" {
+  uses = "_VAR1"
```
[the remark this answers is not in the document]
vipulchhabra99: Hi, @ivotron, Sorry for the delay, Actually there are two issues arising after using the dollar symbol as the starting symbol first the substitutions arguments need to be passed as a string second. Second when I used the $ symbol in sample workflows of integration tests then they are not being parsed in the workflow so I decided to remove $ symbol from the beginning of the variables.

Example to use the substitutions feature : 

```popper run --substitutions _variable=v1 --substitutions _variable2=v2```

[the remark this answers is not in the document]
Comment (ivotron) [reply ci/test/substitutions]: > after using the dollar symbol as the starting symbol first the substitutions arguments need to be passed as a string

can you please elaborate on the above? I am not sure I understand your question

File: cli/popper/parser.py [hunk 8]
```diff
@@ -283,6 +286,76 @@ def _traverse(entrypoint, reachable, actions):
         for a in unreachable:
             self.action.pop(a)
 
+    def parse_substitutions(self, substitutions, allow_loose):
```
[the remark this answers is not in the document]
ivotron: @vipulchhabra99 the above would change a bit:

```python
# create dict of substitutions with '$' in keys
subs = {}
for s in substitutions:
    kv = s.split('=', 1)
    if len(kv) != 0:
        raise Exception("Expecting '=' as separator")
    subs['$'+kv[0]] = kv[1]

# map to keep track which subs have been used
used = {}

for a_name, a_block in self.action.items():
    attr = a_block.get('needs', []):
    for i in range(len(attr)):
        for k, v in subs.items():
            if k in attr[i]:
                used[k] = 1
            attr[i] = attr[i].replace(k, v)
                

         # same for all action attributes (and 'resolves' attribute of workflow block)....

if not allow_loose and len(subs) != len(used):
    # throw exception 
```

ci/test/substitutions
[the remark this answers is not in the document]
vipulchhabra99: Actually, to show in a better way, I am attaching screenshots for the same home it helps out.
There are based on what I assumed from your previous comment.
If I use dollar in substitutions as well
Screenshot from 2020-01-22 06-44-40

The substitutions list becomes blank

Screenshot from 2020-01-22 06-44-25

But when I passed it in the form of string
Screenshot from 2020-01-22 06-47-37

Then the substitutions list got passed but then workflow variables turned blank.
Screenshot from 2020-01-22 06-45-55

But from your last comment If I use it in only workflow file then I would still obtain them blank during parsing.

cli/popper/parser.py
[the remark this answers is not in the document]
vipulchhabra99: Actually, since I am performing this step before the actions variable is declared, But I got the sense of how to use it in current situation.

ci/test/substitutions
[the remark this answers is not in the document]
ivotron: @vipulchhabra99, I see where the confusion is coming from. The `$` is only used in the `.workflow` file; it is **NOT** used when passed in the `--substitutions` flag. So this:

```hcl
workflow "foo" {
  resolves = "$_VAR"
}
```

is defined as:

```bash
popper run --substitution VAR=foo
```

btw, could we do `--substitution` (singular) instead of `--substitutions`? This is more appropriate since each flag defines a single substitution.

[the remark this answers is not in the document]
vipulchhabra99: Yeah actually substitution seems a better one. If i use $ in workflow the variable disappears after parsing. In the last screen shot I am printing parsed workflow. Please check that once.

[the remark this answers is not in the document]
ivotron: we need to escape the dollar sign if we're using it in a script (`\$`)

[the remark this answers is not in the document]
vipulchhabra99: Yeah, thanks it worked.

vipulchhabra99:
Hi @ivotron, There were 2 very minor errors in the previous commit, I have fixed those errors, please review it once and do let me know if more updates are required.

cli/popper/parser.py
```diff
@@ -367,6 +380,96 @@ def _traverse(entrypoint, reachable, actions):
...
+                for k, v in substitution_dict.items():
+                    if k in value:
+                        used[k] = 1
+                        temp_dict[key] = v
+
+            if(len(temp_dict) != 0):
+                wf_block['env'] = temp_dict
+
+        if not allow_loose and len(substitution_dict) != len(used):
+            log.fail("All the substitutions passed are not used !")
```
ivotron: maybe `"Not all given substitutions are used in the workflow file"`

cli/popper/commands/cmd_run.py
```diff
@@ -226,7 +240,8 @@ def run_workflow(**kwargs):
     # Initialize a Workflow. During initialization all the validation
     # takes place automatically.
     wf = Workflow(kwargs['wfile'])
-    wf_runner = WorkflowRunner(wf)
+    wf_runner = WorkflowRunner(wf, kwargs['substitutions'],
+                               kwargs['allow_loose'])
```
ivotron: since substitutions are relevant to workflow parsing (not execution), there is no need for the runner to be aware of these options. Instead, these can be arguments to the `Workflow` constructor (line 242 above)

cli/popper/gha.py
```diff
@@ -218,7 +218,8 @@ def get_workflow_env(wf, workspace):
 
     def run(self, action, skip_clone, skip_pull, skip, workspace,
             reuse, dry_run, parallel, with_dependencies, engine,
-            engine_conf, skip_secrets_prompt=False):
+            engine_conf, substitutions, allow_loose,
```
ivotron: these arguments are not used in this function, so there's no need to pass them

cli/popper/gha.py
```diff
@@ -32,9 +32,9 @@
 class WorkflowRunner(object):
     """A GHA workflow runner."""
 
-    def __init__(self, workflow):
+    def __init__(self, workflow, substitutions=None, allow_loose=False):
         self.wf = workflow
-        self.wf.parse()
+        self.wf.parse(substitutions, allow_loose)
```
ivotron: if we make these arguments be part of the `Workflow` constructor, then we can skip passing them to the `WorkflowRunner` constructor and we'd remove them from this function altogether

cli/popper/parser.py
vipulchhabra99: Thanks ivo, for reviewing the pr, Fixed this part.

cli/popper/commands/cmd_run.py
vipulchhabra99: Fixed.

cli/popper/gha.py
vipulchhabra99: Actually they can't be removed because from cmd_run file kwargs are being passed to run function and hence If I don't create a placeholder here it throws some error.

vipulchhabra99: Actually they can't be removed directly because the kwargs are being passed to this run function from cmd_run and due to which I created a placeholder in this function. To remove from here, I can try other methods like creating a temporary tuple and then putting all the parameters except these 2 and then modifying kwargs.

cli/popper/gha.py
vipulchhabra99: Fixed.

cli/popper/gha.py
ivotron: they need to be removed from the `kwargs` dictionary. I pushed this small change

vipulchhabra99: Ok Thanks a lot, From where I studied it was told that tuple and hence I thought we cannot modify it. I got to learn a new thing.

ivotron:
thanks a lot @vipulchhabra99

PR #759, opened 2020-01-08
Support for OAUTH Token for private repos

### LINKED ISSUE #728: cli: add support for actions in private repositories
Context:

- A container (or action) that checks out a repository
- Secrets are provided to the container at _runtime_, **not** at _build time_

In the given context, a popper workflow supports public repositories by providing secrets to the container from environment variables or interactively (prompt; if env variable not defined). However, this prevents the container image from being run directly (due to docker requiring build args be defined via [ARG instruction][docker-arg]).

However, when running a popper workflow without build args defined via the ARG instruction, the workflow still fails. The exact same instructions work when running the docker image directly (e.g. interactively).

The issue _appears_ to be that the repository being checked out is private. The secrets provided are a username and github token for checking out the private repository, via https.

For referene, the private repository can be found here (hopefully github security holds up):

    https://github.com/drin/xhca-private

[docker-arg]: https://docs.docker.com/engine/reference/builder/#arg

vipulchhabra99:
Fixes #728 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
+---
 dist: xenial
 language: python
 python:
-- '3.7'
+- "3.7"
 env:
-  - RUNTIME=singularity POPPER_TEST_MODE=with-git
-  - RUNTIME=docker POPPER_TEST_MODE=without-git
+- RUNTIME=singularity POPPER_TEST_MODE=with-git
+- RUNTIME=docker POPPER_TEST_MODE=without-git
 services: docker
 before_install:
 - ci/scripts/install_scripts.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,10 @@
----
 dist: xenial
 language: python
 python:
-- "3.7"
+- '3.7'
 env:
-- RUNTIME=singularity POPPER_TEST_MODE=with-git
-- RUNTIME=docker POPPER_TEST_MODE=without-git
+  - RUNTIME=singularity POPPER_TEST_MODE=with-git
+  - RUNTIME=docker POPPER_TEST_MODE=without-git
 services: docker
 before_install:
 - ci/scripts/install_scripts.sh
--- a/cli/popper/scm.py
+++ b/cli/popper/scm.py
@@ -236,17 +236,13 @@ def clone(url, org, repo, repo_dir, version=None):
         auth_token = os.getenv('GITHUB_API_TOKEN')
 
         if(auth_token is not None):
+            # To verify the link of github for private repo support.
             if('github' in url):
                 if('@' not in url):
                     # The authentication token has to be added after protocol
-                    # The length of protocol is 8 in case of https://,
-                    # 7 in case of http://
+                    # The length of protocol is 8 in case of https://
                     if('https' in url):
                         url = url[:8]+auth_token+'@'+url[8:]
-
-                    elif('http' in url):
-                        url = url[:7]+auth_token+'@'+url[7:]
-
 
         repo_url = '{}{}/{}'.format(url, org, repo)
         cloned_repo = git.Repo.clone_from(repo_url, repo_dir)
--- a/cli/popper/scm.py
+++ b/cli/popper/scm.py
@@ -232,6 +232,22 @@ def clone(url, org, repo, repo_dir, version=None):
         else:
             url += '/'
 
+        # To obtain the authentication token if set as environment variable.
+        auth_token = os.getenv('GITHUB_API_TOKEN')
+
+        if(auth_token is not None):
+            if('github' in url):
+                if('@' not in url):
+                    # The authentication token has to be added after protocol
+                    # The length of protocol is 8 in case of https://,
+                    # 7 in case of http://
+                    if('https' in url):
+                        url = url[:8]+auth_token+'@'+url[8:]
+
+                    elif('http' in url):
+                        url = url[:7]+auth_token+'@'+url[7:]
+
+
         repo_url = '{}{}/{}'.format(url, org, repo)
         cloned_repo = git.Repo.clone_from(repo_url, repo_dir)
 
--- a/cli/popper/scm.py
+++ b/cli/popper/scm.py
@@ -235,14 +235,11 @@ def clone(url, org, repo, repo_dir, version=None):
         # To obtain the authentication token if set as environment variable.
         auth_token = os.getenv('GITHUB_API_TOKEN')
 
-        if(auth_token is not None):
+        if(auth_token is not None and 'github' in url and '@' not in url):
             # To verify the link of github for private repo support.
-            if('github' in url):
-                if('@' not in url):
-                    # The authentication token has to be added after protocol
-                    # The length of protocol is 8 in case of https://
-                    if('https' in url):
-                        url = url[:8]+auth_token+'@'+url[8:]
+            # The authentication token has to be added after protocol
+            # The length of protocol is 8 in case of https://
+            url = url[:8]+auth_token+'@'+url[8:]
 
         repo_url = '{}{}/{}'.format(url, org, repo)
         cloned_repo = git.Repo.clone_from(repo_url, repo_dir)
--- a/docs/sections/gha_workflows.md
+++ b/docs/sections/gha_workflows.md
@@ -187,6 +187,14 @@ refer to an action on a public Git repository or Docker container registry:
 
 -->
 
+### Using Private repositories of `Github` in an action
+
+When creating an action block we can also use an action defined in Github private repository. To use private repository you need to create Personal access token from <a href="https://github.com/settings/tokens"> here </a> with permissions to access private repositories and set it as an environment variable using
+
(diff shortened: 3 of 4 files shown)


vipulchhabra99:
@ivotron I have added the support for it in the file specified by you in the issue but It doesn't seems feasible to add the test for it.

JayjeetAtGithub:
@vipulchhabra99 Can you please tell why it's not feasible to add test?

JayjeetAtGithub:
@vipulchhabra99 We don't want to involve `.travis.yml` in storing any API key/secret. Please look at [this](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings). This way we do not need to encrypt the keys ourselves and update the `.travis.yml` multiple times.

vipulchhabra99:
Hi @JayjeetAtGithub , This link points to defining the environment variables in Travis settings of the repo and I think I don't have the rights to add it in environment variables.

cli/popper/scm.py
```diff
@@ -183,6 +183,12 @@ def get_default_branch(r):
         else:
             url += '/'
 
+        auth_token = os.getenv('GITHUB_API_TOKEN')
+
+        if(auth_token is not None):
```
ivotron: can we add logic for checking that the URL is pointing to github? If the URL is pointing somewhere else it'll likely fail.

cli/popper/scm.py
```diff
@@ -183,6 +183,12 @@ def get_default_branch(r):
...
+            if('@' not in url):
```
ivotron: this conditional will result in not being able to work with URLs of the form `git@github.com`, which is OK but we need to add this to the documentation.

cli/popper/scm.py
```diff
@@ -183,6 +183,12 @@ def get_default_branch(r):
...
+                url = url[:8]+auth_token+'@'+url[8:]
```
ivotron: can you add a comment explaining why 8 characters before and after`url[:8]`?

.travis.yml
```diff
@@ -1,11 +1,13 @@
...
-- "3.7"
+- '3.7'
 env:
-- RUNTIME=singularity POPPER_TEST_MODE=with-git
-- RUNTIME=docker POPPER_TEST_MODE=without-git
+  matrix:
+  - RUNTIME=singularity POPPER_TEST_MODE=with-git
+  - RUNTIME=docker POPPER_TEST_MODE=without-git
+  global:
+  - secure: DXDmJmyxb89CZOI4DEoPQE4iI+tna3TDwf0UekgIt36GqJfqDpYjtaar6tZVWhDZVdswOxxe+ulnRnSLLb0l4EDuPnOPu6oiBo9i8q1t5qy1bsaiyyM9pkOnZbKcNEvF9cSv15a/A/m1y6/aOSWFJDaSnKWUu1u0sQrUGuY7XlZCy9lns1p8diAeatQMx9tBy5bcJi4L6xcrHwNMs6IJnBcYEAxKAKHkNMGUOPObVnOyGGLfo4T4hTzVouw7i/wZHfS1KIM/KPYOZuZfNVK/NgV3m4YKlV2Y+zj8T1fsQH/aEIkEoQwUGbZhPofvZIncJ7v2z7TC2qW76DfuwOslDPkxqZ7Lw2/Wx22T3NgKsFKKwFaseH1p+M/PhgS5APR6qU/jDH18Eo8SgdPYgy97zlMgg7qiXdQOK4kCKcGySpJ8qXi+jpBxzxkXpI170SkmWDMd/U5R16hC37rLkcycwgMhz4FcaIBQGTR+6YtG19avW7/Qz1ZlEahjB9ZMgBo4jLaFGGeAOeRMOQ5t0aniPmOWjEuZvR9YKHLKgH8SzpOELh5WD+lc1t73tFfkt50DKTUy2yUvHcU4qXoGdKnY4fiO1dQiAJJ0kqZ5Iw/KjsU+yTVXfiZl+UARLr34yhBv544lKDUiTfqJSw7JgrYTp48ImUIj5OnNGdzophUkJLo=
```
ivotron: this can be removed

ivotron:
@vipulchhabra99 I've added a `GITHUB_API_TOKEN` variable to travis associated to my account. If you create a private action/repo and give my github user access, our travis will be able to test these feature. Note that the test of the PR is not going to pick this env variable up, as workflows triggered by PRs cannot read environment variables. What we can do is that you can test locally on your machine, and once everything works on your end you can post here and I'll manually trigger on this PR. Would that work?

cli/popper/scm.py
vipulchhabra99: Yeah sure.

.travis.yml
vipulchhabra99: OK I can remove this

cli/popper/scm.py
vipulchhabra99: Yes, I can add this to documentation to make it clear to the user.

cli/popper/scm.py
vipulchhabra99: Yeah, sure I can add it.

cli/popper/scm.py
vipulchhabra99: Actually I just realized that there can be other protocols defined http:// and git@ although git@ would automatically wouldn't be considered for adding token, I can add support for it too.

vipulchhabra99: @ivotron Should I also add a check that protocol should be HTTPS only because Token works only with HTTPS protocol.

vipulchhabra99:
Hi, @ivotron I have updated the pr as per your review please check them once. Also, I have added you as a collaborator in that dummy repo from my end please accept invitation so that you can access that private repository and also I have one question that If we have somehow passed the integration tests for it but later on if someone else sends new pull request to the repo will this test would be triggered again? If triggered then I think then Travis would result for build failed for every pull request.

ivotron:
@vipulchhabra99 you're right. An alternative is to check for whether `GITHUB_TOKEN_API` variable is defined and skip the test if is not.

.travis.yml
```diff
@@ -1,11 +1,10 @@
----
 dist: xenial
 language: python
 python:
-- "3.7"
+- '3.7'
 env:
-- RUNTIME=singularity POPPER_TEST_MODE=with-git
-- RUNTIME=docker POPPER_TEST_MODE=without-git
+  - RUNTIME=singularity POPPER_TEST_MODE=with-git
```
ivotron: since these are cosmetic changes, we can leave them out of the PR

cli/popper/scm.py
```diff
@@ -232,6 +232,18 @@ def get_default_branch(r):
         else:
             url += '/'
 
+        # To obtain the authentication token if set as environment variable.
+        auth_token = os.getenv('GITHUB_API_TOKEN')
+
+        if(auth_token is not None):
+            # To verify the link of github for private repo support.
+            if('github' in url):
```
ivotron: this can be a single `if` statement (conjunction of the three `if` conditions)

cli/popper/scm.py
```diff
@@ -232,6 +232,18 @@ def get_default_branch(r):
...
+                if('@' not in url):
+                    # The authentication token has to be added after protocol
+                    # The length of protocol is 8 in case of https://
+                    if('https' in url):
```
ivotron: HTTP is the only protocol used by Popper, so `https` is always in the URL

docs/sections/gha_workflows.md
```diff
@@ -187,6 +187,14 @@ refer to an action on a public Git repository or Docker container registry:
 
 -->
 
+### Using Private repositories of `Github` in an action
```
ivotron: This can be `### Referencing private Github repositories in an action`

docs/sections/gha_workflows.md
```diff
@@ -187,6 +187,14 @@ refer to an action on a public Git repository or Docker container registry:
...
+
+When creating an action block we can also use an action defined in Github private repository. To use private repository you need to create Personal access token from <a href="https://github.com/settings/tokens"> here </a> with permissions to access private repositories and set it as an environment variable using
```
ivotron: couple of changes (note that readthedocs accepts markdown):

We can make use of actions located in private Github repositories by defining a `GITHUB_API_TOKEN` environment variable that the `popper run` command can inspect. To accomplish this, the repository referenced in the `uses` attribute is assumed to be private and, to access it, an API token from Github is needed (see instructions [here](https://github.com/settings/tokens)). The token needs to have permission to read the private repository in question. To define the `GITHUB_API_TOKEN` variable:

```
export GITHUB_API_TOKEN=access_token_here
popper run
```

If the access token doesn't have permissions to access private repositories, the `popper run` command will fail.

vipulchhabra99:
Hi @ivotron, Thanks a lot for reviewing the pr I have updated it as per your review. Please have a look once

ivotron:

vipulchhabra99:
Thanks, a lot @ivotron and @JayjeetAtGithub, Your guidance helped me a lot to solve the issue.

PR #760, opened 2020-01-09
Adds ability to specify custom options for container runtime 

### LINKED ISSUE #723: cli: add ability to specify custom options for container runtime
allow the `popper run` command to read a config file with custom options for the container runtime. This will allow users to define custom volumes, networking configuration, capabilities, etc.

JayjeetAtGithub:
Fix #723 (Implemeted only for docker currently)


ci/test/runtime-conf
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+set -ex
+# shellcheck source=./ci/test/common
+source ./ci/test/common
+init_test_repo
+cd "$test_repo_path"
+
+cat <<EOF > settings.py
+runtime_configuration = {
```
ivotron: we can name this `engine_configuration` instead. Also, we can remove the `docker` entry as that allows us to make the mix of kwargs (in the `mix_with_runtime_config` function) agnostic to each engine

cli/popper/gha.py
```diff
@@ -565,7 +572,19 @@ def docker_rm(self):
...
+                             *runtime_config.get('volumes', list())]
+        for k, v in runtime_config.get('environment', dict()).items():
+            config["environment"].update({k: v})
+
+        for k, v in runtime_config.items():
+            if k not in config.keys():
+                config[k] = runtime_config[k]
+        return config
+
+    def docker_create(self, img, runtime_config):
```
ivotron: we can have `runtime_config=None`

cli/test/test_gha.py
```diff
@@ -413,7 +413,7 @@ def test_docker_image_exists(self):
         'Skipping docker tests...')
     def test_docker_rm(self):
         self.docker_client.images.pull('debian:buster-slim')
-        self.runner.docker_create('debian:buster-slim')
+        self.runner.docker_create('debian:buster-slim', None)
```
ivotron: if we make `docker_create(..., runtime_config=None)`  as mentioned above, we don't need to provide value for it

cli/popper/gha.py
```diff
@@ -254,11 +254,17 @@ def run(self, action, skip_clone, skip_pull, skip, workspace,
         WorkflowRunner.instantiate_runners(
             runtime, new_wf, workspace, dry_run, skip_pull, self.wid)
 
+        # If a runtime configuration file is passed through the
+        # --engine-conf option, parse it and pass it to ActionRunner(s).
+        engine_config = pu.parse_engine_configuration(engine_conf)
```
ivotron: can this be passed to the `instantiate_runners` function instead? then the config dictionary can be given to the ActionRunner constructor and reference to it as a class property. This helps to avoid having to pass the configuration around all the internal methods of runners (e.g. `run`, `docker_run`, etc.)

JayjeetAtGithub: Yeah, that sounds much better. Thanks.

JayjeetAtGithub:
@ivotron Please have a look. Thanks.

cli/popper/commands/cmd_run.py
```diff
@@ -111,6 +111,11 @@
     hidden=True,
     default=popper.scm.get_git_root_folder()
 )
+@click.option(
+    '--engine-conf',
+    help='Specify the engine configuration file for a container.',
```
ivotron: maybe `File containing configuration options for the underlying container engine.`

docs/sections/cli_features.md
```diff
@@ -54,6 +54,36 @@ or to execute all the workflows in a project:
 popper run --recursive
 ```
 
+## Specifying extra engine configuration
```
ivotron: "Custom container engine configuration" or "Customizing container engine behavior"

docs/sections/cli_features.md
```diff
@@ -54,6 +54,36 @@ or to execute all the workflows in a project:
...
+
+Many times, workflows might require the containers with some special configuration
+like extra privileges, extra volume bindings, extra capabilites, etc i-e users
+might want to use the different runtime features that container technologies provide.
+Since, Popper by default spawns containers in their most basic and default configuration,
+a `settings.py` file can be passed to `--engine-conf` option of `popper run`
+to specify the extra configuration that is required for the particular use case.
```
ivotron: some tweaks:

By default, Popper instantiates containers in the underlying engine by using basic configuration options (see [here](https://popper.readthedocs.io/en/latest/sections/gha_workflows.html?highlight=mount#execution-runtime)). When these options are not suitable to your needs, you can modify or extend them by providing engine-specific options. These options allow you to specify fine-grained capabilities, bind-mounting additional folders, etc. In order to do this, you can provide a configuration file to modify the underlying container engine configuration used to spawn containers. This file is a python script that defines an `engine_configuration` dictionary with custom options and is passed to the `popper run` command via the `--engine-conf` flag.

docs/sections/cli_features.md
```diff
@@ -54,6 +54,36 @@ or to execute all the workflows in a project:
...
+```python
+engine_configuration = {
+  'privileged': True
+}
+```
+
+After creating the file, the workflow can be run using the above configuration by doing
+```
+popper run --engine-conf settings.py
+```
```
ivotron: the file can be called anything, right? So we can write instead:

For example, to make Popper spawn Docker containers in [privileged mode](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities), we can write the following options:

```python
engine_configuration = {
  'privileged': True
}
```

Assuming the above is stored in a file called `settings.py`, we pass it to Popper by running:

```bash
popper run --engine-conf settings.py
```

JayjeetAtGithub: Sorry, actually i somehow missed it before. I hardcoded the filename. Now its fine. We can keep any file at any path.

ivotron:
awesome stuff, thanks @JayjeetAtGithub !

PR #772, opened 2020-02-05
Unit tests for click based CLI interface 

### LINKED ISSUE #761: cli: unit test the click based CLI interface
Interested? Have a look at this. http://click.palletsprojects.com/en/7.x/testing/

vipulchhabra99:
Fixes #761 
I have added the Unit test for command ci currently.
And is the issue have to be solved in a similar way for all other commands?


ivotron:
hi @vipulchhabra99 thanks a lot. Yes, that's the idea: port all the commands from the shell scripts to python and delete all of them once all of them are ported.

vipulchhabra99:
hi @ivotron, I thought of creating a single file containing all the test cases is this fine?

ivotron:
Yeah, that'd be ideal

vipulchhabra99:
@ivotron, I think some of the has clashed with other tests and due to which the tests are failing. Am I guessing correct?

ivotron:
@vipulchhabra99 sorry for the confusion. The goal is to port the tests from shell scripts to python, that is, we won't be running `popper run --wfile ci/cli.workflow` from the test, but instead implement those tests in Python.

vipulchhabra99:
Hi @ivotron , I have one more question in porting them to python do I need to run the commands through os module or I have to pass the arguments of commands written in shell scripts to run function?

ivotron:
> Hi @ivotron , I have one more question in porting them to python do I need to run the commands through os module or I have to pass the arguments of commands written in shell scripts to run function(Through Click testing)?

the latter (through click testing)

vipulchhabra99:
ok, thanks @ivotron then I believe I would have to do in a separate file for each command.

vipulchhabra99:
@ivotron , I have updated the file by partially completing skip tests. Could you please review it once so that If I am progressing in right direction, I can continue for other parts

vipulchhabra99:
Yeah great @ivotron , I have understood the points just one last question. Can I do in parts and get it reviewed so that It becomes easy for me as well as you to review it completely?

ivotron:
> Yeah great @ivotron , I have understood the points just one last question. Can I do in parts and get it reviewed so that It becomes easy for me as well as you to review it completely?

whatever works best for you. Either is fine on my end

> Also, can I create the common file using the os module to execute the commands?

this is similar to the test scripts. That is, the functionality contained in the `ci/test/common` file needs to be ported to Python.

vipulchhabra99:
HI @ivotron, Sorry for the delays I have written the common file and the siginit function is still pending, I would add it soon and for some of the part, I have used the os module for them which I am planning to replace with the appropriate module. Please review the common file once.

cli/test/test_common.py
```diff
@@ -0,0 +1,54 @@
+import os
```
ivotron: Instead of defining functions, we can have define an abstract `PopperTest` class that has these as members. All the tests that we write as part of this PR can derive from this class, so that common functionality gets consolidated

cli/test/test_common.py
```diff
@@ -0,0 +1,54 @@
...
+import sys
+import shutil
+import contextlib
+import subprocess
+from git import Repo
+
+os.environ['test_repo_path'] = '/tmp/mypaper'
```
ivotron: this can be a member of the `PopperTest` class

cli/test/test_common.py
```diff
@@ -0,0 +1,54 @@
...
+os.environ['CI'] = 'false'
+
+pushstack = list()
+
+
+def pushdir(dirname):
```
ivotron: there's no need to have `pushdir` and `popdir` as the GitPython library can handle that. We do that on the shell scripts because we assume we run on the shell, but now that we are running these tests in Python, we can discard some shell-specific functionality

cli/test/test_common.py
```diff
@@ -0,0 +1,54 @@
...
+        os.system("docker run --rm -v /tmp:/tmp alpine:3.8 rm -rf "+path)
+
+
+def init_test_repo():
+
+    if(os.environ.get('POPPER_TEST_MODE') == 'with-git'):
+        path = os.environ.get('test_repo_path')
+        delete_dir(path)
+        os.mkdir(path)
+        pushdir(path)
```
ivotron: we can remove this `pushdir`, as described above

cli/test/test_common.py
```diff
@@ -0,0 +1,54 @@
...
+        repo = Repo.init(path)
+        file = open(path+'/README.md', 'w+')
+        file.write("foo")
+        repo.git.add(A=True)
+        repo.git.commit('-m', 'first')
+        popdir()
```
ivotron: same here

cli/test/test_cmd_commands.py
```diff
@@ -0,0 +1,113 @@
...
+    os.chdir(os.environ.get('test_repo_path'))
+
+    def create_workflow_file(self, content):
+        f = open(os.environ.get('test_repo_path')+'/a.workflow', 'w+')
+        f.write(content)
+        f.close()
+
+    def test_skip(self):
+
+        self.create_workflow_file("""
```
ivotron: let's create file-based fixtures so that we can test both yaml and hcl with the same test. Please see [this comment](https://github.com/systemslab/popper/pull/772#pullrequestreview-355561864)

ivotron:
@vipulchhabra99, now that #779 and #781 have been merged, we'll need to rebase master for this PR?

vipulchhabra99:
Ok, sure @ivotron I will rebase the branch, I will get back to you soon with new updates.

cli/test/test_common.py
vipulchhabra99: Yes @ivotron,  This would be a much better approach to create this file.

vipulchhabra99:
> thanks @vipulchhabra99! couple of observations:
> 
> * we need to add the functionality of `ci/test/common` file and the `init_test_repo` function to the tests. See for example the `ci/test/skip` script to see how it's making use of them.
> * as part of the above, we need to support the `POPPER_TEST_MODE` environment variable.
> * you might want to create a super class that does the above so there's no need to be repeating this code.
> * use file-based fixtures (inside a `cli/test/fixtures` folder) instead of inlining the content of workflows.
> 
> the last step above is motivated by the following. We will soon support a YAML syntax that's similar to HCL (#773), so it would be ideal if we can abstract the syntax from the tests. For example, we can have `cli/test/fixtures/skip.workflow` and `cli/test/fixtures/skip.yml` files, and let the test look for the appropriate file based on an environment variable (e.g. `POPPER_TEST_FORMAT`, selecting `.workflow` by default).
> 
> let me know if anything from above is not clear.
> 
> thanks!

In this like you mentioned for POPPER_TEST_FORMAT do it needs to accept it as an argument during the running of the tests?

ivotron:
hi @vipulchhabra99, we merged a major refactoring of the code base that also affected unit tests. Can you please rebase master? Or it might be easier to start all over with a new PR.

In short, it is not necessary to have `POPPER_TEST_MODE` or `POPPER_TEST_FORMAT` environment variables anymore. Take a look at the [`travis.yml`](https://github.com/systemslab/popper/blob/master/.travis.yml) file to see how tests are getting invoked. Also, you can take a look at the [parser tests](https://github.com/systemslab/popper/blob/master/cli/test/test_parser.py) to get an idea of how to make use of file-based fixtures. Also, it is OK to only use YAML for your tests, no need to write them both in HCL and YAML, as the differences in the syntax are already tested by the parser tests. In other words, for the click unit tests, we only care that the CLI can be invoked as it is described in the docstring help.

please let me know if you have any question

vipulchhabra99:
> hi @vipulchhabra99, we merged a major refactoring of the code base that also affected unit tests. Can you please rebase master? Or it might be easier to start all over with a new PR.
> 
> In short, it is not necessary to have `POPPER_TEST_MODE` or `POPPER_TEST_FORMAT` environment variables anymore. Take a look at the [`travis.yml`](https://github.com/systemslab/popper/blob/master/.travis.yml) file to see how tests are getting invoked. Also, you can take a look at the [parser tests](https://github.com/systemslab/popper/blob/master/cli/test/test_parser.py) to get an idea of how to make use of file-based fixtures. Also, it is OK to only use YAML for your tests, no need to write them both in HCL and YAML, as the differences in the syntax are already tested by the parser tests. In other words, for the click unit tests, we only care that the CLI can be invoked as it is described in the docstring help.
> 
> Also, for the click tests, it is not necessary to create a git repository, we can only create workflow files and scripts associated to the workflows. The tests that check for support for .git folders has been consolidated into the `test_scm.py` file.
> 
> please let me know if you have any question

Thanks @ivotron for clearing the doubts, Yes I have rebased the branch after which I start getting the error that I mentioned in the communication channel. I even tried today on one VM but I got a similar error. But I believe that It would be great to give this issue a fresh start since a lot of the features have changed. So I can close this pr and start from scratch.

Issue #773, opened 2020-02-05
cli: add YAML syntax similar to HCL


ivotron:
add a YAML syntax that is similar to HCL. For example, a workflow similar to the [minimal-python](https://github.com/popperized/popper-examples/blob/master/workflows/minimal-python/main.workflow) one when written in YAML:

```yaml
steps:

- uses: "popperized/bin/curl@master"
  args: [
    "--create-dirs",
    "-Lo workflows/minimal-python/data/global.csv",
    "https://github.com/datasets/co2-fossil-global/raw/master/global.csv"
  ]

- id: 'analysis 1'
  uses: "jefftriplett/python-actions@master"
  args: "workflows/minimal-python/scripts/get_mean_by_group.py"
    "workflows/minimal-python/data/global.csv",
    "5"
  ]

- id: 'analysis 2'
  uses: "jefftriplett/python-actions@master"
  args: "workflows/minimal-python/scripts/analysis2.py"

- uses: "jefftriplett/python-actions@master"
  needs: ['analysis 1', 'analysis 2']
  args: [
    "workflows/minimal-python/scripts/validate_output.py",
    "workflows/minimal-python/data/global_per_capita_mean.csv"
  ]
```

some comments:

  * if no `id` for a step is given, an internal one is generated (e.g. `1` for the first step above).
  * if no `needs` attribute is given for a step, the `steps` list is used to determine the order in which the step is executed. In the example above, `analysis 1` executes after the first step.
  * if a `needs` attribute is given, then that is used to determine the order in which to execute steps. For example, in the above, the last step references `analysis 1` and `analysis 2` as the steps needed to execute before it.
  * all the other attributes of HCL syntax action (`env`, `secrets`, etc..) are supported.

JayjeetAtGithub:
> steps list is used to determine the order in which the step is executed.

@ivotron I am confused about what will be the execution order for the above workflow.  Can you please tell me like how should steps without `needs` attribute be connected to the graph generated by the steps having `needs` attribute?

ivotron:
hey @JayjeetAtGithub, sorry for the confusion.

> Is it like we generate a DAG as far as possible connecting as many steps as possible to and then connect the remaining steps(actions) with the root of the forward edged graph in the order given in which they are given in the `steps` list?

yes. Another way to think about is like this: if a step doesn't have a `needs` entry, it has to wait for all previous steps to finish before it executes. So a more illustrative example (I ignore args for simplicity) would be:

```yaml
steps:

[...]

PR #779, opened 2020-02-13
Add YAML Syntax

JayjeetAtGithub:
Addresses #773 
### TODO:
* Integration tests need to be ported to `yml` syntax.

### DIFF
--- a/cli/popper/commands/cmd_dot.py
+++ b/cli/popper/commands/cmd_dot.py
@@ -77,7 +77,7 @@ def cli(ctx, wfile, skip, recursive, colors):
 
             stage_edges.add(edge)
 
-            for M in wf.get_action(n).get('next', []):
+            for M in wf.action[n].get('next', []):
                 dot_str = add_to_graph(dot_str, wf, n, [M],
                                        node_attrs, stage_edges)
         return dot_str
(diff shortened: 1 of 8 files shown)


cli/popper/commands/cmd_dot.py
```diff
@@ -90,6 +90,11 @@ def add_to_graph(dot_str, wf, parent, children, node_attrs, stage_edges):
         wfile_list.append(pu.find_default_wfile(wfile))
 
     for wfile in wfile_list:
+        if wfile.endswith('.workflow'):
```
ivotron: we can create a static factory method of `Workflow` that has this logic so that we don't need to repeat it every time we want to instantiate a `Workflow` object? Something like `Workflow.new_workflow(filename)`?

cli/popper/commands/cmd_run.py
```diff
@@ -7,7 +7,7 @@
 import popper.cli
 from popper.cli import pass_context, log
 from popper.gha import WorkflowRunner
-from popper.parser import Workflow
+from popper.parser import HCLWorkflow, YMLWorkflow
```
ivotron: as a follow up to the other comment: if we have a factory method, we allow callers not to worry about the format, so we can do:

```python
from popper.parser import Workflow

# ...

Workflow.new(kwargs['wfile'])
```

cli/popper/commands/cmd_dot.py
JayjeetAtGithub: Ah. Yes. This looks much cleaner.

ivotron:

PR #786, opened 2020-02-20
removed the `on` attribute from workflow

### LINKED ISSUE #763: cli: remove 'on' attribute from workflow
now that github has deprecated the HCL syntax, this attribute is not relevant anymore

WJayesh:
Please review the changes and suggest any corrections. Thanks
I have also slightly modified the description of `instantiate_runners` method. Shouldn't the `wid` parameter be of `bool` type? 
Closes #763

### DIFF
--- a/cli/popper/parser.py
+++ b/cli/popper/parser.py
@@ -11,13 +11,8 @@ import re
 import os
 
 
-<<<<<<< HEAD
-VALID_STEP_ATTRS = ["uses", "args", "needs", "runs", "secrets", "env"]
-VALID_WORKFLOW_ATTRS = ["resolves", "on"]
-=======
 VALID_ACTION_ATTRS = ["uses", "args", "needs", "runs", "secrets", "env"]
 VALID_WORKFLOW_ATTRS = ["resolves"]
->>>>>>> removed the `on` attribute from workflow
 
 
 class Workflow(object):
--- a/cli/popper/parser.py
+++ b/cli/popper/parser.py
@@ -11,8 +11,13 @@ import re
 import os
 
 
+<<<<<<< HEAD
 VALID_STEP_ATTRS = ["uses", "args", "needs", "runs", "secrets", "env"]
 VALID_WORKFLOW_ATTRS = ["resolves", "on"]
+=======
+VALID_ACTION_ATTRS = ["uses", "args", "needs", "runs", "secrets", "env"]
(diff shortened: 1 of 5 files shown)


ivotron:
thanks @WJayesh , can you please rebase master?

WJayesh:
Can you please review the changes again? I'm not sure what's causing the builds to fail; I've looked at the logs but i don't think they pertain to the changes I've made. thanks!

cli/popper/parser.py
```diff
@@ -12,7 +12,7 @@
 
 
 VALID_STEP_ATTRS = ["uses", "args", "needs", "runs", "secrets", "env"]
-VALID_WORKFLOW_ATTRS = ["resolves", "on"]
+VALID_WORKFLOW_ATTRS = ["resolves"]
```
ivotron: doesn't need to be removed from here.

cli/popper/parser.py
```diff
@@ -168,7 +168,7 @@ def validate_workflow_block(self):
 
         workflow_block = list(self.wf_dict['workflow'].values())[0]
         for key in workflow_block.keys():
-            if key not in VALID_WORKFLOW_ATTRS:
+            if key not in VALID_WORKFLOW_ATTRS and key != "on":
```
ivotron: we can also leave this as is

cli/popper/parser.py
```diff
@@ -180,9 +180,6 @@ def validate_workflow_block(self):
             log.fail('[resolves] attribute must be a string or a list '
                      'of strings.')
 
-        if workflow_block.get('on', None):
```
ivotron: these are the actual changes that are needed in order to ignore the `on` attribute. So in other words, a workflow can still have them, but we just ignore them.

cli/test/test_parser.py
```diff
@@ -1074,7 +1074,6 @@ def test_normalize(self):
         hcl_workflow.normalize()
         self.assertEqual(hcl_workflow.resolves, ['a'])
         self.assertEqual(hcl_workflow.name, 'sample workflow')
-        self.assertEqual(hcl_workflow.on, 'push')
```
ivotron: all tests in this file can also be left as is

WJayesh: in that case, won't these tests show errors if run?

ivotron:
hcl is not the main syntax now. This can be reopened if needed

Issue #764, opened 2020-01-17
cli: use f-strings


ivotron:
now that support for python2.7 has been deprecated, we can make use of [f-strings](https://realpython.com/python-f-strings/)

edeediong:
@ivotron, I'd like to pick this issue up.

PR #793, opened 2020-02-28
fix f string in codebase #764

edeediong:
The refactoring took a lot of the format method away from the repo. I fixed the remaining format usage. We're now covered with Python >3.6x syntax.

### DIFF
--- a/cli/popper/log.py
+++ b/cli/popper/log.py
@@ -34,12 +34,12 @@ class PopperFormatter(logging.Formatter):
     BOLD_RED = '[01;31m'
 
     log_format = {
-        'DEBUG':       '{}%(levelname)s: %(msg)s {}'.format(BOLD_CYAN, RESET),
+        'DEBUG':       f'{BOLD_CYAN}%(levelname)s: %(msg)s {RESET}',
         'STEP_INFO': '%(msg)s',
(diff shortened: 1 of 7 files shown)


cli/popper/log.py
```diff
@@ -34,12 +34,12 @@ class PopperFormatter(logging.Formatter):
     BOLD_RED = '[01;31m'
 
     log_format = {
-        'DEBUG':       '{}%(levelname)s: %(msg)s {}'.format(BOLD_CYAN, RESET),
+        'DEBUG':       f'{BOLD_CYAN}%(levelname)s: %(msg)s {RESET}',
```
ivotron: thanks @edeediong! Minor comment: the `%` can also be changed for f-strings

edeediong: Oh, I didn't know that. I'm gonna do that.

edeediong: @ivotron I saw https://docs.python.org/3/library/logging.html#logging.Formatter and https://stackoverflow.com/questions/56269302/how-to-setup-python-logging-format-using-f-string-style. But I'm still yet to understand how to format the loggers to f string

ivotron: that's ok. I'll merge as is. thanks!

PR #794, opened 2020-02-29
Unit Test for click Based Interface

vipulchhabra99:
fixes for #761.
Currently written for skip part other parts will be added soon.


vipulchhabra99:
Hi, @ivotron I was going through the integration test files. There on executing the shell command, the actions that are being run successfully in dry-run are checked via grep so should I have to also convert them in a similar way like calling them through the subprocess module and piping the output back to test file and checking its presence?.

cli/test/popper_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def init_test_repo(self):
+
+        path = self.test_repo_path
+        delete_dir(path)
+        os.mkdir(path)
+        file = open(path+'/README.md', 'w+')
+        file.write("foo")
+
+    def check_output(self, command, actions):
```
ivotron: the `click.testing.Result` class provides an `output` property containing the output of the command. So no need to use `subprocess` for this

cli/test/test_cmdskip.py
```diff
@@ -0,0 +1,103 @@
...
+            run.cli, ['--dry-run', '--wfile', workflow_file_loc,
+                      '--skip', 'e1'])
+        assert result.exit_code == 1
+
+        result = runner.invoke(
+            run.cli, ['--wfile', workflow_file_loc, '--dry-run', '--skip',
+                      'a1', '--skip', 'a2'])
+        assert result.exit_code == 1
+
+        result = self.check_output(['popper', 'run', '--dry-run', '--wfile',
```
ivotron: you can inspect the contents of `result.output`

cli/test/test_cmdversion.py
```diff
@@ -0,0 +1,12 @@
...
+import unittest
+from popper_test import PopperTest
+
+
+class TestCommandVersion(unittest.TestCase, PopperTest):
+
+    def test_version(self):
+        runner = CliRunner()
+        result = runner.invoke(version.cli)
+        assert result.exit_code == 0
```
ivotron: we can check that the version string returned is the same as [`popper.__init__.py`](https://github.com/systemslab/popper/blob/master/cli/popper/__init__.py)

cli/test/test_cmdskip.py
```diff
@@ -0,0 +1,103 @@
...
+
+        result = self.check_output(
+            ['popper', 'run', '--wfile', workflow_file_loc, '--dry-run',
+             '--skip', 'a2'], ['a1', 'b', 'c', 'd'])
+        assert result == 0
+
+        result = runner.invoke(
+            run.cli, ['--wfile', workflow_file_loc, '--dry-run',
+                      '--skip', 'b', '--skip', 'c'])
+        assert result.exit_code == 1
```
ivotron: we might want to also check that the error being reported is the one we expect.

cli/test/test_cmdskip.py
```diff
@@ -0,0 +1,103 @@
...
+
+def _wfile(name, format):
+    return f'{FIXDIR}/{name}.{format}'
+
+
+class TestSkip(unittest.TestCase, PopperTest):
+
+    def test_skip(self):
+
+        workflow_file_loc = _wfile("skip", "workflow")
```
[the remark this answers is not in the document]
vipulchhabra99: Thanks for Reviewing the pr @ivotron, I will make it to YAML, I made it workflow for the time being because of the issues that some features are currently not working in YAML files so to prevent failure of the tests I have currently decided to keep it workflow.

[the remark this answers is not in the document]
ivotron: OK. We can leave `.workflow` for skip then.

cli/test/test_cmdskip.py
vipulchhabra99: Yes, It would be nice to have it since we can debug the errors in a better way, Thanks for pointing it out.

cli/test/test_cmdversion.py
vipulchhabra99: Ok, I can add that here.

cli/test/test_cmdskip.py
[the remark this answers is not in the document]
vipulchhabra99: Yes, Sure.

cli/test/popper_test.py
vipulchhabra99: Actually, I forgot to mention that the output coming is only containing newlines. I think this is because the print is being done through internal functions and output is not being returned hence to pipe out the output I used subprocess.

cli/test/popper_test.py
```diff
@@ -0,0 +1,57 @@
...
+        self.ci = False
+
+    def delete_dir(self, path):
+
+        try:
+            os.rmdir(path)
+        except OSError as e:
+            os.system("docker run --rm -v /tmp:/tmp alpine:3.8 rm -rf "+path)
+
+    def init_test_repo(self):
```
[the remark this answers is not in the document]
vipulchhabra99: Sure, I will keep update with the new changes.

cli/test/test_cmdskip.py
vipulchhabra99: Since the output is blank so should I match using subprocess only?

cli/test/popper_test.py
ivotron: oh, you're right, everything that gets printed to stdout is via the logging framework. So we need to use the unit testing facilities for loggers as explained [here](https://docs.python.org/3/library/unittest.html?highlight=unittest#unittest.TestCase.assertLogs)

cli/test/test_cmdskip.py
ivotron: we can check the contents of the loggers, so no need to use subprocess

vipulchhabra99:
Hi @ivotron, I have made the required changes can you please review it once so that I can start with the conversion of other tests.

ivotron:
hi @vipulchhabra99 , they're looking great 👍

vipulchhabra99:
Ok cool @ivotron , I can convert other files in similar way and resolve these conflicts.

vipulchhabra99:
Hi @ivotron, I was writing a test for scaffold since It generates a file should I use a similar folder being used in CI tests?

vipulchhabra99:
@ivotron, I have removed one test in Sh test since both used external script with similar content just the diff was one uses the argument from the user while other doesn't so I kept only one which takes arguments from the user

ivotron:
> I was writing a test for scaffold since It generates a file should I use a similar folder being used in CI tests?

the `mk_repo()` function has code exemplifying the use of the [`tempfile`](https://docs.python.org/3/library/tempfile.html) module. You can use that to get to a temporary folder, then invoke scaffold, and then check the contents of the generated file

vipulchhabra99:
Hi @ivotron, I saw that test_runner_docker is removed from the current test files, Should I also delete that file?

ivotron:
> Hi @ivotron, I saw that test_runner_docker is removed from the current test files, Should I also delete that file?

please rebase master, as that file was removed in one of the commits that we merged recently

vipulchhabra99:
Ya, I rebased but I was still showing me.

vipulchhabra99:
There are some validation tests that are working fine locally but not using the validation python file so probably I would try them once everything else is completed. For time being I have commented them out.

ivotron:
breaking this down into multiple PRs, starting with #803 and following up with others

PR #799, opened 2020-03-20
Add slurm runner

JayjeetAtGithub:
Adds basic Slurm resource manager support


cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,155 @@
...
+
+        return ecode
+
+    def stop_srun_cmd(self):
+        for p in SlurmRunner.spawned_processes:
+            log.info(f'Stopping proces {p.pid}')
+            p.kill()
+
+
+class HostRunner(SlurmRunner):
```
ivotron: we can remove this and instead add at the top:

```
from popper.runner_host import HostRunner
```

since the `HostRunner` would be the same as the one  in `runner_host.py`

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,155 @@
...
+        log.debug(f'Environment:\n{pu.prettystr(os.environ)}')
+
+        ecode = self.exec_srun_cmd(cmd)
+        return ecode
+
+    def stop_running_tasks(self):
+        self.stop_srun_cmd()
+
+
+class DockerRunner(SlurmRunner, HostDockerRunner):
```
ivotron: very nice! I think this approach is going to help with other container engines. Really cool

cli/popper/runner_slurm.py
JayjeetAtGithub: Actually, the `runner_slurm.HostRunner` is quite different than in`runner_host.HostRunner`. It just prepares the command and uses the methods defined in `runner_slurm.SlurmRunner` to execute the command.

JayjeetAtGithub:
> thanks a lot @JayjeetAtGithub, this is looking great! I think the only thing that would be missing would be to add tests.

Yeah. Would be pushing tests soon.

cli/popper/runner_slurm.py
ivotron: I see. And how do you envision this being used from the point of view a workflow? Were you thinking the `runner_slurm.HostRunner` would be used for `uses: "sh"` steps? For example, when this runs:

```yaml
- uses: sh
  runs: ls
```

the semantics is that the command will run directly on the machine where `popper run` is being executed, not via `srun`.

JayjeetAtGithub: Done

ci/test/slurm
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
ivotron: removing this would help speedup the test in travis. Mocking the subprocess calls is enough to test that we're doing things right

cli/popper/runner.py
```diff
@@ -36,24 +36,29 @@ def __init__(self, config_file=None, workspace_dir=os.getcwd(),
...
-
-        if not self.config.resman:
-            self.config.resman = 'host'
+            # lets consider for now, name property will be present.
+            if config_from_file.get('engine', None):
+                self.config.engine_name = config_from_file['engine']['name']
+                self.config.engine_options = config_from_file['engine'].get(
+                    'options', None)
+            if config_from_file.get('resource_manager', None):
+                self.config.resman_name = config_from_file['resource_manager']['name']
```
ivotron: same here

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,127 @@
...
+
+    spawned_processes = set()
+
+    def __init__(self, config):
+        super(SlurmRunner, self).__init__(config)
+
+    def __exit__(self, exc_type, exc, traceback):
+        SlurmRunner.spawned_processes = set()
+
+    def exec_srun_cmd(self, cmd):
```
ivotron: given that this is very similar to the code in `runner_host.HostRunner`, maybe we can refactor the code in `runner_host.HostRunner` so that there's a member function that can be invoked here, that has all this logic and the only thing we pass that is different is the leading `srun` to the command.

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,127 @@
...
+            p.kill()
+
+
+class DockerRunner(SlurmRunner, HostDockerRunner):
+    """Runs steps in docker."""
+
+    def __init__(self, config):
+        super(DockerRunner, self).__init__(config)
+
+    def run(self, step):
```
ivotron: we can refactor this function so that we avoid repeating code from `runner_host.DockerRunner`. Maybe create utility function in `runner_host.DockerRunner` that can be reused here?

cli/popper/runner.py
```diff
@@ -36,24 +36,29 @@ def __init__(self, config_file=None, workspace_dir=os.getcwd(),
...
-            if hasattr(config_from_file, 'ENGINE'):
-                self.config.engine_options = config_from_file.ENGINE
-            if hasattr(config_from_file, 'RESOURCE_MANAGER'):
-                self.config.resman_options = config_from_file.RESOURCE_MANAGER
-
-        if not self.config.resman:
-            self.config.resman = 'host'
+            # lets consider for now, name property will be present.
+            if config_from_file.get('engine', None):
+                self.config.engine_name = config_from_file['engine']['name']
```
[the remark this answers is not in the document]
JayjeetAtGithub: I think, now is time we should implement the `PopperConfig` class, which would contain the logic of parsing and validating the config file. What do you say ?

cli/popper/runner_slurm.py
JayjeetAtGithub: Can we have the function in `utils.py` like it was before ?

cli/popper/runner.py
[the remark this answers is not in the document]
ivotron: yeah, great idea!

cli/popper/runner_slurm.py
ivotron: sure, that'd work. I thought of having it in the parent class because currently there's no other code that makes us of it. But having in utils.py is fine

ci/test/engine-conf
```diff
@@ -17,58 +17,58 @@ action "run" {
...
-ENGINE = {
-    "name": "dont use this name",
-    "image": "abc/xyz",
-    "hostname": "xYz.local"
-}
+engine:
+  name: docker
+  options:
+    image: abc/xyz
+    hostname: xYz.local
```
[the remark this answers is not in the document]
JayjeetAtGithub: Done

cli/popper/runner_slurm.py
JayjeetAtGithub: I did some refactoring. Please review.

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,70 @@
...
+            container.remove(force=True)
+
+        container = HostDockerRunner.create_container(step, self.config)
+        log.info(f'[{step["name"]}] srun docker start')
+
+        if self.config.dry_run:
+            return 0
+
+        HostDockerRunner.spawned_containers.append(container)
+        ecode = self.start_container(cid)
```
ivotron: in a slurm setup, the above runs on the machine where `popper run` is being executed (i.e. the frontend node). But the goal is to run this on multiple nodes, so all this needs to be executed by slurm. One thing we can do is create a script and instruct slurm to execute it, but it looks like `srun` doesn't support this, so we'll need to use `sbatch`.

ivotron: another thing that this brings up is how to expose options for slurm jobs. This can be done at the level of the step:

```yaml
- id: 'mpi-step'
  uses: 'docker://an-mpi-image'
  options:
    slurm:
      cpus-per-task: 4
      nodes: 64
      ...
```

and the above would be communicated to `sbatch` in the script as explained [here](https://help.rc.ufl.edu/doc/Sample_SLURM_Scripts). The other alternative is to have it in the config file like:

```yaml
resource_manager:
  name: slurm
  options:
    mpi-step:
      cpus-per-task: 4
      nodes: 64
      ...
```

cli/popper/runner.py
```diff
@@ -33,27 +34,29 @@ def __init__(self, config_file=None, workspace_dir=os.getcwd(),
...
         if config_file:
-            # read options from config file
-            config_from_file = pu.load_config_file(config_file)
+            popper_cfg = PopperConfigParser(config_file)
 
-            if hasattr(config_from_file, 'ENGINE'):
-                self.config.engine_options = config_from_file.ENGINE
-            if hasattr(config_from_file, 'RESOURCE_MANAGER'):
-                self.config.resman_options = config_from_file.RESOURCE_MANAGER
+            if popper_cfg.engine:
```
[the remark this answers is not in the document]
JayjeetAtGithub: @ivotron Even if we put this in the class, We need to again have some kind of `if else` statement in this fn. I think if we put this logic in the `PopperConfigParser` class, the complexity will increase here.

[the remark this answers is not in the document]
ivotron: > can we have a resource manager option in `popper run` also, to keep things uniform?

yeah, that's a good idea. Maybe `--resource-manager` and short form `-r`? And we could also take this opportunity to introduce a short version of `--engine` (`-e`)?

> also, if a user passes both `--engine` option value and also defines an engine value in `settings.yml` which one to prioritize if they have different value? Based upon these i would think of the design.

common practice in CLI applications is to have flags win (w.r.t. precedence) over those defined in config files. So we can do the same here: flag overrides the value in the conf file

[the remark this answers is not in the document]
JayjeetAtGithub: Ok thanks.

cli/popper/runner_slurm.py
JayjeetAtGithub: I think having the options in the `settings.yml` is a good idea.

cli/popper/commands/cmd_run.py
```diff
@@ -50,11 +50,18 @@
...
 )
 @click.option(
+    '-e',
     '--engine',
-    help='Specify runtime for executing the workflow [default: docker].',
+    help='Specify runtime for executing the workflow.',
     type=click.Choice(['docker', 'singularity', 'vagrant']),
-    required=False,
-    default='docker'
+    required=False
```
ivotron: this can default to docker, and remove the `required=False`

cli/popper/commands/cmd_run.py
```diff
@@ -50,11 +50,18 @@
...
+)
+@click.option(
+    '-r',
+    '--resource-manager',
+    help='Specify resource manager for executing the workflow.',
+    type=click.Choice(['slurm']),
+    required=False
```
ivotron: this is missing the `host` option. And `host` can be the default and we can remove the `required=False`

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,87 @@
...
+            container.remove(force=True)
+
+        container = HostDockerRunner.create_container(cid, step, self.config)
+        log.info(f'[{step["name"]}] srun docker start')
+
+        if self.config.dry_run:
+            return 0
+
+        HostDockerRunner.spawned_containers.append(container)
+        ecode = self.start_container(cid)
```
ivotron: same comment as before for lines 60-75. This has to run in an `sbatch` script

cli/popper/runner.py
```diff
@@ -32,28 +33,33 @@ def __init__(self, config_file=None, workspace_dir=os.getcwd(),
...
+            [self.config.engine_name, self.popper_cfg.engine.name, 'docker'])
+        self.config.resman_name = pu.select_not_none(
+            [self.config.resman_name,
+             self.popper_cfg.resource_manager.name,
+             'host'])
 
-        if not self.config.resman:
-            self.config.resman = 'host'
+        self.config.resman_options = self.popper_cfg.resource_manager.options
+        self.config.engine_options = self.popper_cfg.engine.options
```
ivotron: Lines 27-51 can be handled by the `PopperConfig` class. The idea is to hide all config-related functionality in that object, not only reading/parsing the configuration file. We can think of PopperConfig as an encapsulation of anything that has to do with configuration. The resulting object is passed around to access that information

cli/popper/commands/cmd_run.py
JayjeetAtGithub: @ivotron if we default this to `docker` or the `--resource-manager` flag to `host`, then we would have no way to check if the flags were given or not, and hence we cant write the logic of prioritizing flags over config files.

cli/popper/commands/cmd_run.py
JayjeetAtGithub: Same as above

cli/popper/commands/cmd_run.py
ivotron: Oh, nevermind, based on what we discussed on slack over option preferences, this is how we should expose this option

cli/popper/commands/cmd_run.py
ivotron: same here, it is fine. Nevermind

ivotron: but we do have to add `host` as a choice

cli/popper/runner.py
ivotron: In other words. objects making use of the config object don't need to be aware that it's a `DotMap`. They only use the `config.foo` in order to access the value of `foo`

cli/popper/commands/cmd_run.py
JayjeetAtGithub: Yeah i will be removing the `required=False` and add the host option.

cli/popper/commands/cmd_run.py
JayjeetAtGithub: np, will be removing `required=False`

ivotron: > if we default this to `docker` or the `--resource-manager` flag to `host`, then we would have no way to check if the flags were given or not, and hence we cant write the logic of prioritizing flags over config files.

sorry, we were writing at the same time. You're right, I forgot what we had discussed on slack. Leaving `required=False` is what we should do in order to let the engine info be obtained from the config file

cli/popper/commands/cmd_run.py
ivotron: we can leave `required=False`, my first comment didn't take into account what we discussed on slack regarding config precedence (CLI flag winning over config file)

cli/popper/runner_slurm.py
JayjeetAtGithub: `srun` works fine. it has all the options as for `sbatch`. Do you mean we need to run all docker related functionality like creating container, finding container, pulling images etc through as slurm script ?

cli/popper/commands/cmd_run.py
JayjeetAtGithub: sorry, no problem.

cli/popper/runner_slurm.py
ivotron: yes, these tasks need to run on the remote machines (via slurm), not on the local host. So we need to have a bash script that stops/removes the container if it's running, and then launch the container. If all this can be done with a single `srun` then it's OK, but if we have to invoke more than one `srun`, then it'll be better to use `sbatch` because there is an overhead associated to `srun` commands (there's some waiting involved when getting a node allocated).

JayjeetAtGithub: I see. I would check and change it.

cli/popper/runner.py
JayjeetAtGithub: I see. Ok.

ivotron: another way of looking at this. The idea would be to do:

```python
kwargs = locals()
kwargs.pop('self')
self.config = PopperConfig(**kwargs)
```

following code uses `self.config` directly without knowing it's backed by a `DotMap`. That means we can remove `dotmap` from being imported here.

Another thing that this might mean is that we can mover the `PopperConfig` from parser and put it in `popper.config`. This object represents the configuration, not only parses it.

JayjeetAtGithub: > As a folllowup of the above, we can move the PopperConfig from parser and put it in popper.config. This object represents the configuration, not only parses it.

You mean having a new `config.py` module right ? Yeah, thats makes sense.

cli/popper/runner.py
```diff
@@ -18,48 +18,31 @@ class WorkflowRunner(object):
...
-            if hasattr(config_from_file, 'ENGINE'):
-                self.config.engine_options = config_from_file.ENGINE
-            if hasattr(config_from_file, 'RESOURCE_MANAGER'):
-                self.config.resman_options = config_from_file.RESOURCE_MANAGER
-
-        if not self.config.resman:
-            self.config.resman = 'host'
+        kwargs = locals()
+        kwargs.pop('self')
+        self.config = PopperConfig(**kwargs)
```
ivotron: really cool 😺

ivotron: we can remove the comment on line 25, or change it to something like "create a config object from the given arguments"

cli/popper/runner_host.py
```diff
@@ -157,14 +103,62 @@ def run(self, step):
...
-
         return e
 
     def stop_running_tasks(self):
         for c in DockerRunner.spawned_containers:
             log.info(f'Stopping container {c.name}')
             c.stop()
 
+    @staticmethod
+    def prepare_image(step, config):
```
ivotron: excellent. This refactoring is going to be very useful to reuse on runners for other platforms

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,206 @@
...
+        job_script = os.path.join(temp_dir, f"{job_id}.sh")
+        out_file = os.path.join(temp_dir, f"{job_id}.out")
+        err_file = os.path.join(temp_dir, f"{job_id}.err")
+
+        self.generate_script(cmd, job_id, job_script)
+
+        sbatch_cmd = "sbatch --wait "
+        sbatch_cmd += f"--job-name {job_id} "
+        sbatch_cmd += f"--output {out_file} "
+        sbatch_cmd += f"--error {err_file} "
```
ivotron: since we're going to attach to the job, we can skip configuring output files and let that to be given by the user if they want

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,206 @@
...
+                    if isinstance(config_val, bool):
+                        if len(config_key) == 1:
+                            sbatch_cmd += f"-{config_key} "
+                        else:
+                            sbatch_cmd += f"--{config_key} "
+                    else:
+                        if len(config_key) == 1:
+                            sbatch_cmd += f"-{config_key} {config_val} "
+                        else:
+                            sbatch_cmd += f"--{config_key} {config_val} "
```
ivotron: since boolean values can be assigned true/false, another way to write this that saves a couple of lines and indentation levels:

```python
resman_options = self.config.resman_options
step_options = resman_options.get(step['name'], [])

for config_key, config_val in step_options:
    flags += "-" if len(config_key) == 1 else "--"
    flags += f"{config_key}='{config_val}'"
```

the above also takes care of having values for options that have `=` in them

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,206 @@
...
+        SlurmRunner.spawned_jobs.remove(job_id)
+        return ecode
+
+    def cancel_job(self):
+        for job_id in SlurmRunner.spawned_jobs:
+            log.info(f'Cancelling job {job_id}')
+            pu.exec_cmd(["scancel", "--name", job_id])
+
+
+class DockerRunner(SlurmRunner):
```
ivotron: excellent 😎

cli/popper/runner_host.py
JayjeetAtGithub: Yeah

cli/popper/runner.py
JayjeetAtGithub: Yeah, sorry.

cli/popper/runner_slurm.py
ivotron: I just realized that the above assumes that if `resman_options` is always assigned a value, which we can fix if we make the `PopperConfig` object always assign a dict, regardless of whether a config file is passed or not. In the latter case, the dict would be empty

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def __init__(self, config):
+        super(SlurmRunner, self).__init__(config)
+
+    def __exit__(self, exc_type, exc, traceback):
+        SlurmRunner.spawned_jobs = set()
+
+    def stream_output(self, job_id):
+        log.debug(f"Attaching to Job {job_id}")
+        while True:
```
ivotron: I cannot quite grasp the need for this loop and the subsequent `sattach` invocation. If this is due to `sattach` failing because the job doesn’t exist yet, then this could potentially loop infinitely when `docker run` returns a non-exit code (when the job was created and immediately exited between loop iterations).

Also, this loop can potentially bombard the slurm server, so we can add use the `SLURM_EXIT_ERROR` environment variable as described [here](https://slurm.schedmd.com/sattach.html) to obtain an arbitrary exit code for when slurm returns a failure (as when the job doesn't exist). In this way, we loop as long as we find this exit code, and we don't need to make a subsequent call to `sattach`

JayjeetAtGithub: @ivotron Yeah, the loop checks whether for the job is whether the job is in pending state or not. As soon as the job changes state from `P` to `R`, `sattach` receives a exit code of 0. This the point we break from the loop. Note, the commands in the script are still not executed.  Now the subsequent exit code tells us whether the commands pass/fail and the action fails accordingly.

ivotron: I see, so then we need to use another mechanism to check the status of the job. The sattach docs imply that the exit code returned is coming from the underlying `srun`, which in this case would correspond to the exit code from `docker run`. Two alternatives:
 1. do what I mentioned above (use `SLURM_EXIT_ERROR` to disambiguate between `docker run` and `sattach` errors; and remove the 2nd invocation to `sattach`).
 2. use `squeue` to check the status of the job, and then run `sattach` once we know is running.

JayjeetAtGithub: Sorry, but what I meant was there's a stage which returns `0` exit code between `sattach failing for the last time with 1 exit code` and `docker run starting may be 1 or 0`. So, theres no chance of overlapping of `sattach` and `docker run` error. That's the reason i am looping and checking for the first `0` and breaking out there. Does that make sense?

JayjeetAtGithub: Let me try it out with `squeue`. looks like it will be easier.

JayjeetAtGithub: @ivotron looks like the complexity is the same or increases to some extent on using `squeue` to check job status. Even if we use `SLURM_EXIT_ERROR` environment variable, we still need to distinguish between the `0` exit code from `sattach` and the `0` exit code from `docker run`, thus we break and call `sattach` again. The `sattach` command isn't much user friendly, part of which is leading to this increased complexity. I really can't think of any other way to optimize this.

cli/popper/utils.py
```diff
@@ -191,9 +182,49 @@ def assert_executable_exists(command):
...
+                   universal_newlines=True, preexec_fn=os.setsid,
+                   env=env, cwd=cwd) as p:
+
+            spawned_processes.add(p)
+            log.debug('Reading process output')
+
+            output = ""
+            for line in iter(p.stdout.readline, ''):
+                line_decoded = decode(line)
+                output += line_decoded
```
ivotron: this can potentially end up consuming a lot of memory. If `stream=True` we can ignore and return an empty output

cli/popper/utils.py
```diff
@@ -191,9 +182,49 @@ def assert_executable_exists(command):
...
         a = a.toDict()
-    if type(a) == os._Environ:
+    if isinstance(a, os._Environ):
         a = dict(a)
-    if type(a) == dict:
+    if isinstance(a, dict):
         return f'{yaml.dump(a, default_flow_style=False)}'
+
+
+def exec_cmd(cmd, env=None, cwd=os.getcwd(), spawned_processes=set(), stream=True):
```
ivotron: stream usually means that this is a generator function, so maybe we can rename that arg to `log` which indicates that output is to be logged rather than returned

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def start_output_stream(self, out_file):
+        self.stream_thread = threading.Thread(
+            target=self._stream_output, args=(out_file,))
+        self.stream_thread.start()
+
+    def stop_output_stream(self):
+        proc = list(self.tail_proc_pid)[0]
+        proc.kill()
+        self.stream_thread.join()
```
ivotron: nice!

cli/popper/runner_slurm.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # submit the job and wait, then parse the job_id
+        ecode, output = pu.exec_cmd(sbatch_cmd.split(" "), stream=False)
+        job_id = int(output.split(" ")[-1].strip("\n"))
+        
+        # kill the tail process
+        self.stop_output_stream()
+
+        SlurmRunner.spawned_jobs.remove(job_name)
+        return ecode
```
ivotron: pleasantly readable :-)

cli/popper/runner.py
```diff
@@ -4,11 +4,11 @@
 import sys
 
 from dotmap import DotMap
-from hashlib import shake_256
 
 import popper.scm as scm
 import popper.utils as pu
 
+from popper.config import PopperConfig
```
ivotron: we can also remove the `DotMap` import

cli/test/test_parser.py
```diff
@@ -633,3 +633,37 @@ def test_complete_graph(self):
                 'name': 'end'}
         }
         self.assertDictEqual(wf.steps, steps_dict)
+
+
+class TestPopperConfigParser(unittest.TestCase):
```
ivotron: maybe move this to a new test_config.py?

JayjeetAtGithub: Done

PR #802, opened 2020-04-08
slurm tweaks

ivotron:
several tweaks to slurm runner, along with refactoring of parent classes (runner.py and runner_host.py)


JayjeetAtGithub:
@ivotron LGTM.

cli/popper/runner_slurm.py
```diff
@@ -1,249 +1,144 @@
...
-        temp_dir = "/tmp/popper/slurm/"
-        os.makedirs(temp_dir, exist_ok=True)
+        with open(job_script, 'w') as f:
+            f.write('#!/bin/bash\n')
+            f.write('\n'.join(cmd))
 
-        job_script = os.path.join(temp_dir, f"{job_name}.sh")
-        out_file = os.path.join(temp_dir, f"{job_name}.out")
-        err_file = os.path.join(temp_dir, f"{job_name}.err")
+        sbatch_cmd = f'sbatch --wait --job-name {job_name} --output {out_file}'
```
JayjeetAtGithub, line 49: @ivotron Shouldn't there be a `--error {err_file}` also. So, if there is an error in the command running in the job script, that could be streamed to the user. Maybe we can have 2 threads, one looking at the out file and other looking at the err file. What do you say ?

cli/popper/runner_slurm.py
```diff
@@ -1,249 +1,144 @@
...
-        elif not self.config.skip_pull and not step.get('skip_pull', False):
-            DockerRunner.docker_pull(step, img, self.config.dry_run)
+        cmd.append(f'docker rm -f {cid} || true')
 
-        # remove container if it exists
-        DockerRunner.docker_rm(step, cid, self.config.dry_run)
+        if build:
+            cmd.append(f'docker build {img}:{tag} {dockerfile}')
+        elif not self._config.skip_pull and not step.get('skip_pull', False):
+            cmd.append(f'docker pull {img}:{tag}')
```
JayjeetAtGithub, line 104: Can we redirect the output of these build/pull commands to `/dev/null`, otherwise they would show up in the output/error stream ?

cli/popper/runner_slurm.py
```diff
@@ -1,249 +1,144 @@
...
-        build, img, dockerfile = HostDockerRunner.get_build_info(
-            step, self.config.workspace_dir, self.config.workspace_sha)
+        build, img, tag, dockerfile = self._get_build_info(step)
 
-        if build:
-            DockerRunner.docker_build(
-                step, img, dockerfile, self.config.dry_run)
-        elif not self.config.skip_pull and not step.get('skip_pull', False):
-            DockerRunner.docker_pull(step, img, self.config.dry_run)
+        cmd.append(f'docker rm -f {cid} || true')
```
JayjeetAtGithub, line 99: Same here. `/dev/null` maybe

cli/popper/runner_host.py
```diff
@@ -36,225 +38,215 @@ def run(self, step):
...
+        log.debug(f'Docker info: {pu.prettystr(self._d.info())}')
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
-        if DockerRunner.d:
-            DockerRunner.d.close()
-        DockerRunner.spawned_containers = []
+        if self._d:
+            self._d.close()
+        self._spawned_containers = []
         return True
```
JayjeetAtGithub, line 118: We don't need this `return=True` statement right ?

cli/popper/runner_slurm.py
ivotron: when no `--error` is given, sbatch combines the output of stderr and stdout in the same file.

cli/popper/runner_host.py
ivotron: true, I'll remove

cli/popper/runner_slurm.py
ivotron: that's the goal, to show the output of this operations, unless `--quiet` is given. But now that you mention it, we should be using `log.step_info` instead of `info()` from the `HostRunner._exec_cmd()` function. I'll change it

cli/popper/runner_slurm.py
JayjeetAtGithub: Oh, sorry, i didn't knew it.

cli/popper/runner_slurm.py
JayjeetAtGithub: Yeah. Thanks

cli/popper/runner_host.py
```diff
@@ -36,225 +38,215 @@ def run(self, step):
...
+            log.info(f'Stopping proces {pid}')
+            os.kill(pid, signal.SIGKILL)
+
+    @staticmethod
+    def _exec_cmd(cmd, env=None, cwd=os.getcwd(), pids=set(), logging=True):
+        pid = 0
+        try:
+            with Popen(cmd, stdout=PIPE, stderr=STDOUT,
+                       universal_newlines=True, preexec_fn=os.setsid,
+                       env=env, cwd=cwd, text=True) as p:
```
JayjeetAtGithub, line 65: @ivotron I checked the docs, and it tells that `text=True` option is not supported before Python 3.7. So, we can either keep the previous way of doing `decode('utf-8')` on the output logs or drop py3.6 support. What do you say? I vote for the first one.

ivotron: oh, good catch! We should use `universal_newlines=True` then

cli/popper/runner_host.py
ivotron: can you add this on your end? in that way I avoid disrupting the state of this branch on your end

cli/popper/runner_host.py
JayjeetAtGithub: Yeah. I will change to it. Thanks.

cli/popper/runner_host.py
JayjeetAtGithub: yes. sure.

cli/popper/runner_host.py
```diff
@@ -36,225 +38,214 @@ def run(self, step):
...
+        if self._config.dry_run:
             return 0
 
-        DockerRunner.spawned_containers.append(container)
+        self._spawned_containers.add(container)
 
         container.start()
         cout = container.logs(stream=True)
         for line in cout:
-            log.step_info(pu.decode(line).strip('\n'))
```
JayjeetAtGithub, line 103: @ivotron  Without decoding and stripping the logs, byte strings are getting printed like this, 

```
b'hello\n'
```
Can we bring the decoding back ?

ivotron: I think we need to investigate why this is not decoded already. Based on the documentation, when `stream=True`, this is a generator of lines, so it should be returning text without new line characters

JayjeetAtGithub: @ivotron I just checked this by writing an independent script with docker py to spawn a container and stream logs. I found it prints out byte string by default.
```python3
import docker

client = docker.from_env()
c = client.containers.create('alpine:3.9', 'ls', detach=True)
c.start()
cout = c.logs(stream=True)
for line in cout:
	print(line)
```

gives output, 

```sh
b'bin\n'
b'dev\n'
b'etc\n'
b'home\n'
b'lib\n'
b'media\n'
b'mnt\n'
b'opt\n'
b'proc\n'
b'root\n'
b'run\n'
b'sbin\n'
b'srv\n'
b'sys\n'
b'tmp\n'
b'usr\n'
b'var\n'
```
So, looks like this is the default behaviour.

ivotron: thanks Jayjeet. The above reproduces the problem but doesn't explain why this is the default behavior. One option is to look at the docker-py code and see why when `stream=True` the library returns raw lines instead of decoded ones.

cli/test/test_runner_host.py
```diff
@@ -1,13 +1,19 @@
 import os
+import time
 import unittest
 
+from subprocess import Popen
+
+import psutil
```
ivotron, line 7: `psutil` is not a pure-python library (it gets compiled when installed), so this makes popper less portable and harder to install. Since this is only a test dependency, we could add it as a test dependency like we had before (where we would install using `pip install cli/[dev]`)

cli/test/test_runner_host.py
```diff
@@ -74,19 +80,123 @@ def test_exec_cmd(self):
...
         self.assertTrue('TESTACION' in output)
 
-    # TODO: add test_stop_running_tasks
+    def test_stop_running_tasks(self):
+        pid = Popen(["sleep", "2000"]).pid
+        host_runner = HostRunner()
+        host_runner._spawned_pids.add(pid)
+        host_runner.stop_running_tasks()
+        time.sleep(2)
+        proc = psutil.Process(pid)
```
ivotron, line 89: this can be checked with `kill -0 <pid>`

cli/popper/runner_host.py
JayjeetAtGithub: Ok will do that

JayjeetAtGithub: Looking at the code of `docker-py`, I found this 
[this](https://github.com/docker/docker-py/blob/030af62dca982e62ed418314c857da1f90ce4282/docker/api/client.py#L393) function where if stream and tty is enabled, it looks like binary logs are streamed, but I am still not sure.

ivotron: aha! that's the answer. We need to do `tty=True` when we create the container

JayjeetAtGithub: I did `tty=True` but it then streams a sequence of bytes. I debugged the docker-py library all the combinations, looks like there's no way to get text output from `container.logs()`.

ivotron: I see. I misread the code, you're right. Since we don't support python2, we could do this directly:

```python
cout = container.logs(stream=True)
for raw_line in cout:
    log.step_info(raw_line.rstrip().decode('utf-8'))
```

we might need to the deal with this differently if `tty: True` is defined in the options. Something like docker-compose's [log printer](https://github.com/docker/compose/blob/abef11b2a6fa5bed4a86c3ebc022fb5e8898217a/compose/cli/log_printer.py), cause we also need to deal with the output of build and pull

cli/popper/runner_host.py
```diff
@@ -136,7 +136,7 @@ def run(self, step):
...
+            log.step_info(pu.decode(line).rstrip())
```
ivotron, line 139: the utility function was introduced to deal with compatibilities between python 2 and 3. We don't support python 2 now, so we can remove that and invoke `decode` directly on the `bytes` variable

cli/test/test_runner_slurm.py
```diff
@@ -24,14 +28,16 @@ def tearDown(self):
...
-        # TODO: assert that tail was invoked
+        self.assertEqual(
+            self.slurm_runner._tail_output('slurm-x.out'), 0)
+        self.assertEqual(len(self.slurm_runner._out_stream_pid), 1)
 
     def test_stop_running_tasks(self):
         self.Popen.set_command('scancel --name job_a', returncode=0)
         self.slurm_runner._spawned_jobs.add('job_a')
+        # If no exception is raised, implies scancel was called and
+        # got caught in mock Popen.
```
ivotron: this is not necessarily true. The ideal is to check by invoking the `self.Popen.all_calls()` to ensure `scancel` got called.

cli/test/test_runner_slurm.py
```diff
@@ -1,133 +1,241 @@
...
+    def test_submit_job_failure(self, mock_kill):
+        self.Popen.set_command(
+            'sbatch --wait --job-name popper_1_123abc '
+            '--output /tmp/popper/slurm/popper_1_123abc.out '
+            '/tmp/popper/slurm/popper_1_123abc.sh', returncode=12)
+
+        self.Popen.set_command(
+            'tail -f /tmp/popper/slurm/popper_1_123abc.out',
+            returncode=0)
+        repo = testutils.mk_repo().working_dir
```
ivotron, line 83: same comment here regarding the creation of a git repo (we can use `tempfile` instead)

cli/test/test_runner_slurm.py
```diff
@@ -1,133 +1,241 @@
...
+        with WorkflowRunner(config) as r:
+            wf = YMLWorkflow("""
+            version: '1'
+            steps:
+            - uses: 'popperized/bin/sh@master'
+              runs: [cat]
+              args: README.md
+            """)
+            wf.parse()
+            self.assertRaises(SystemExit, r.run, wf)
```
ivotron, line 107: assert the value of `all_cals()`

cli/test/test_runner_slurm.py
```diff
@@ -1,133 +1,241 @@
...
+            version: '1'
+            steps:
+            - uses: 'popperized/bin/sh@master'
+              runs: [cat]
+              args: README.md
+            """)
+            wf.parse()
+            r.run(wf)
+
+        self.assertEqual(self.Popen.all_calls, [])
```
ivotron, line 128: cool!

cli/setup.py
```diff
@@ -25,6 +22,9 @@
         'pyyaml',
         'docker'
     ],
+    extras_require={
+        'dev': ['psutil', 'testfixtures']
```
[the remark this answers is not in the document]
ivotron: btw, this is what I do on my machine:

```
cd src/popper
source ~/virtualenvs/popper
ENGINE=docker python -X tracemalloc -m unittest discover -f cli/test/
```

the above assumes that there's a virtualenv for popper on `~/virtualenvs/`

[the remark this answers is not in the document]
JayjeetAtGithub: Yeah, that is pretty much how I run tests on my machine also. Thanks a lot. I would add it to the guide.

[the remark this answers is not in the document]
JayjeetAtGithub: @ivotron Should we also mention about how to run the integration tests locally?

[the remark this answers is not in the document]
ivotron: we'll soon move away from running integration tests with bash scripts, and this will be the only way in which we run unit/integration tests

cli/popper/runner_host.py
```diff
@@ -36,225 +38,215 @@ def run(self, step):
...
+                self._d.images.build(path=dockerfile, tag=f'{img}:{tag}',
+                                     rm=True, pull=True)
+        elif not self._config.skip_pull and not step.get('skip_pull', False):
+            log.info(f'[{step["name"]}] docker pull {img}:{tag}')
+            if not self._config.dry_run:
+                self._d.images.pull(repository=f'{img}:{tag}')
+
+        log.info(f'[{step["name"]}] docker create {img}:{tag}')
+        if self._config.dry_run:
+            return
```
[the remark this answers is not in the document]
JayjeetAtGithub: @ivotron If we want this info statement to contain all the arguments, then the arguments would need a lot of preprocessing like converting list , dict arguments to options properly and hence would make the logic complex. I think the current one is simple and we might change it to like 
```
        log.info(f'[{step["name"]}] docker create {img}:{tag} {cmd}')
```

What do you say ?

cli/setup.py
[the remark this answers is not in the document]
JayjeetAtGithub: Okay.

cli/popper/runner_host.py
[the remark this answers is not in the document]
ivotron: `container_args` is a dict, so I think it's fine to print it using an f-string like:

```
log.info(f'[{step["name"]}] docker create {container_args} {img}:{tag}
```

the main goal is to provide the information that someone would need in order to type the command by themselves. Ideally we'd want the equivalent `docker create` command, but as you mentioned it'd be more involved in order to produce proper flags, so printing the dictionary inline is OK, as this shows the info needed. Users need to still work on creating a command out of it but at least they have what is needed (which btw also contains engine config options)

cli/popper/commands/cmd_run.py
```diff
@@ -120,7 +120,21 @@
 def cli(ctx, step, wfile, debug, dry_run, log_file, quiet, reuse,
         engine, resource_manager, skip, skip_pull, skip_clone,
         substitution, allow_loose, with_dependencies, workspace, conf):
-    """Runs a Popper workflow. Only execute STEP if given."""
+    """Runs a Popper workflow. Only execute STEP if given.
+    
+    `popper run` allows specifying the engine and the resource manager
```
ivotron, line 125: couple of comments:

  * we can use "This command" instead of `popper run`, as this is the help message for `popper run`.
  * the list mentions "two different ways" but the list has four bullet points. So we can either turn everything into a paragraph, or only show the two points and turn the last two points into a paragraph.
  * For the second bullet point, we can say through the "through a configuration file specified with the `--conf` flag"
  * For the third point, instead of "Otherwise", we can say "If none of the above are given, Popper uses..."

CONTRIBUTING.md
```diff
@@ -78,6 +78,21 @@ the above approach you have both (1) popper installed in your machine
 and (2) an environment where you can modify popper and test the 
 results of such modifications.
 
+## Running the unittests locally
```
ivotron, line 81: can we take this opportunity to move the `Keywords in issue titles`, `Branches` and `Install from source` sections to be level `##`?

JayjeetAtGithub: Yeah, that looks better.

cli/popper/runner_host.py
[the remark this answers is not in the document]
JayjeetAtGithub: Yeah, fine

PR #803, opened 2020-04-14
Common file and Version Test Created

vipulchhabra99:


cli/test/popper_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+        try:
+            os.rmdir(path)
+        except OSError as e:
+            os.system("docker run --rm -v /tmp:/tmp alpine:3.8 rm -rf "+path)
+
+    def _wfile(self, name, format):
+        return f'{FIXDIR}/{name}.{format}'
+
+    def mk_repo(self):
```
ivotron, line 29: now that this was moved here, we can remove the `cli/popper/test/utils.py` module

cli/test/popper_test.py
```diff
@@ -0,0 +1,45 @@
...
+        self.ci = False
+
+    def delete_dir(self, path):
+
+        try:
+            os.rmdir(path)
+        except OSError as e:
+            os.system("docker run --rm -v /tmp:/tmp alpine:3.8 rm -rf "+path)
+
+    def _wfile(self, name, format):
```
ivotron, line 26: this function is not needed anymore, so we can safely remove it

cli/test/popper_test.py
```diff
@@ -0,0 +1,45 @@
...
+import shutil
+import contextlib
+import subprocess
+import tempfile
+import git
+
+FIXDIR = f'{os.path.dirname(os.path.realpath(__file__))}/fixtures'
+
+
+class PopperTest:
```
ivotron, line 12: can we add a test for the `mk_repo` function in this same class?

cli/test/test_cmdversion.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class TestCommandVersion(unittest.TestCase, PopperTest):
+
+    def test_version(self):
+
+        with self.assertLogs('popper') as test:
+            runner = CliRunner()
+            result = runner.invoke(version.cli)
+            assert result.exit_code == 0
```
ivotron, line 15: this and the following two assertions can use `unittest` assertions instead, like we do in other tests

cli/test/popper_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __init__(self):
+
+        self.test_repo_path = '/tmp/mypaper'
+        self.ci = False
+
+    def delete_dir(self, path):
```
ivotron, line 19: this can also be removed

vipulchhabra99:
@ivotron Updated the pr with required changes.

cli/test/test_common.py
```diff
@@ -0,0 +1,41 @@
...
+import contextlib
+import subprocess
+import tempfile
+import git
+import unittest
+
+FIXDIR = f'{os.path.dirname(os.path.realpath(__file__))}/fixtures'
+
+
+class PopperTest:
```
ivotron, line 13: one last change it just ocurred to me. We can derive from `unittest.TestCase` here so that children classes only derive from `PopperTest`

vipulchhabra99: Updated for all those classes that were inheriting the PopperTest class.

PR #807, opened 2020-04-19
Adds 'popper scaffold' test 

vipulchhabra99:
The popper scaffold test has been updated using click unit testing.

### DIFF
--- a/cli/test/test_scaffold.py
+++ b/cli/test/test_scaffold.py
@@ -25,21 +25,25 @@ class TestScaffold(PopperTest):
 
         wf = Workflow.new(file_loc)
         self.assertDictEqual(wf.steps, {
-        	'1': {
-        		'uses': 'popperized/bin/sh@master',
-        		'args': ['ls'],
-        		'name': '1',
-        		'next': {'2'}},
-        	'2': {
-        		'uses': 'docker://alpine:3.11',
-        		'args': ['ls'],
-        		'name': '2',
-        		'needs': ['1']}})
+            '1': {
+                'uses': 'popperized/bin/sh@master',
+                'args': ['ls'],
+                        'name': '1',
+                        'next': {'2'}},
+            '2': {
+                'uses': 'docker://alpine:3.11',
+                'args': ['ls'],
+                'name': '2',
+                        'needs': ['1']}})
 
         with self.assertLogs('popper') as test_logger:
 
             result = runner.invoke(run.cli, ['-f', file_loc])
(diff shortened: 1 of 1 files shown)


cli/test/test_scaffold.py
```diff
@@ -0,0 +1,29 @@
...
+import os
+
+
+class TestScaffold(PopperTest):
+
+    def test_scaffold(self):
+
+        repo = self.mk_repo()
+        runner = CliRunner()
+        file_loc = repo.working_dir+'/wf.yml'
```
ivotron, line 17: `mk_repo` creates a git repo. In this case, we don't really need one, so we can instead use `tempfile`:

```python
wf_dir = tempfile.mkdtemp()
file_loc = f'{wf_dir}/wf.yml'
```

BTW, we can use f-strings here as well as shown above, in order to keep it consistent with the rest of the codebase

cli/test/test_scaffold.py
```diff
@@ -0,0 +1,29 @@
...
+
+        result = runner.invoke(scaffold.cli, ['-f', file_loc])
+
+        self.assertEqual(result.exit_code, 0)
+        self.assertTrue(os.path.isfile(file_loc))
```
ivotron, line 22: we can also check the content of the file to ensure that scaffolding did what it should. To do this, we can create a `Workflow` object from the file, and then compare it as shown [here](https://github.com/systemslab/popper/blob/master/cli/test/test_parser.py#L246).

cli/test/test_scaffold.py
```diff
@@ -0,0 +1,29 @@
...
+
+        with self.assertLogs('popper') as test_logger:
+
+            result = runner.invoke(run.cli, ['-f', file_loc])
+            self.assertEqual(result.exit_code, 0)
+
+        self.assertTrue(len(test_logger.output))
```
ivotron, line 29: this should be indented within the `with` statement, as the `test_logger` variable is only valid in this scope

ivotron: we can also inspect the content of the output to ensure that it ran as it should

vipulchhabra99: Can I inspect the number of steps it ran because the other output is very lengthy?

ivotron: yeah, that works fine. we only need to solve the pep8 issues that might arise

vipulchhabra99: Updated.

ivotron:

PR #808, opened 2020-04-19
Kubernetes Runner

edeediong:


cli/test/test_runner_kubernetes.py
```diff
@@ -0,0 +1,58 @@
...
+class TestKubernetesDockerRunner(unittest.TestCase):
+    def setUp(self):
+        log.setLevel('CRITICAL')
+
+    @unittest.skipIf(
+        os.environ['ENGINE'] != 'docker' or os.environ['RESMAN'] != 'kubernetes',
+        'ENGINE != docker or RESMAN != kubernetes')
+    def test_docker_basic_run(self):
+        repo = testutils.mk_repo()
+        conf = PopperConfig(workspace_dir=repo.working_dir)
```
ivotron, line 22: thanks a lot again @edeediong for working on this, really appreciate it! 

In order to instruct popper to use the kubernetes runner, we can need to add `resman_name='kubernetes'` to these arguments.

edeediong: Alright sure!

cli/test/test_runner_kubernetes.py
```diff
@@ -0,0 +1,58 @@
+import os
+import unittest
+
+import utils as testutils
```
ivotron, line 4: also, I forgot to mention, this module disappeared in one of our recent changes to `master`

cli/test/test_runner_kubernetes.py
```diff
@@ -0,0 +1,58 @@
...
+
+from popper.config import PopperConfig
+from popper.parser import YMLWorkflow
+from popper.runner import WorkflowRunner
+
+from popper.cli import log as log
+
+
+class TestKubernetesDockerRunner(unittest.TestCase):
```
ivotron, line 13: there is a `PopperTest` that we can derive from that has the `mk_repo` function as a member. Take a look at how this is done [here](https://github.com/systemslab/popper/blob/master/cli/test/test_runner_host.py#L19).

cli/test/test_runner_kubernetes.py
```diff
@@ -0,0 +1,58 @@
...
+    def setUp(self):
+        log.setLevel('CRITICAL')
+
+    @unittest.skipIf(
+        os.environ['ENGINE'] != 'docker' or os.environ['RESMAN'] != 'kubernetes',
+        'ENGINE != docker or RESMAN != kubernetes')
+    def test_docker_basic_run(self):
+        repo = testutils.mk_repo()
```
ivotron, line 21: once we derive from `PopperTest`, then this would become `self.mk_repo()`

edeediong: Done

cli/test/test_runner_kubernetes.py
edeediong: Done

cli/test/test_runner_kubernetes.py
edeediong: Noted

cli/test/test_runner_kubernetes.py
edeediong: done

edeediong: If I add `resman_name='kubernetes'` to `PopperConfig` as a parameter, it sorta breaks the codebase.

PR #810, opened 2020-04-22
Adds 'dot' test

vipulchhabra99:


cli/test/test_cmddot.py
```diff
@@ -0,0 +1,43 @@
...
+class TestDot(PopperTest):
+
+    def test_dot(self):
+
+        wf_dir = tempfile.mkdtemp()
+        runner = CliRunner()
+        git.Git(wf_dir).clone(
+            "https://github.com/popperized/github-actions-demo.git")
+
+        workflow_loc = f'{wf_dir}/github-actions-demo/.github/main.workflow'
```
JayjeetAtGithub, line 23: May we can change the name to `workflow_path`.

vipulchhabra99: Yes sure

JayjeetAtGithub:
Since the whole point of adding python tests is removing the bash script tests, we can take this opportunity to keep removing the files for which we are adding unit tests. Like this one and scaffold may be.
What do you think?

vipulchhabra99:
> Since the whole point of adding python tests is removing the bash script tests, we can take this opportunity to keep removing the files for which we are adding unit tests. Like this one and scaffold may be.
> What do you think?

Yes sure, it would be nice to remove them and have better clean codebase.

JayjeetAtGithub:
> > Since the whole point of adding python tests is removing the bash script tests, we can take this opportunity to keep removing the files for which we are adding unit tests. Like this one and scaffold may be.
> > What do you think?
> 
> Yes sure, it would be nice to remove them and have better clean codebase.

Perfect

JayjeetAtGithub:
@ivotron  @vipulchhabra99 LGTM

PR #813, opened 2020-04-25
Tests for 'Substitution' flag added

vipulchhabra99:


ivotron:
thanks @vipulchhabra99 ! And apologies for the confusion. This is already tested in https://github.com/systemslab/popper/blob/master/cli/test/test_parser.py#L387

Issue #811, opened 2020-04-22
cli: add short versions of quiet and debug flag


JayjeetAtGithub:
Since `--quiet` and `--debug` are one of the frequently used flags of the [`popper run`](https://github.com/systemslab/popper/blob/master/cli/popper/commands/cmd_run.py) command, we can add support for short forms of them also like `-q` and `-d` respectively just like in the `--wfile` or `--engine` flags.

ivotron:
solved in 414f76ea728fa9f2ab688752e2a517536846d407

PR #814, opened 2020-04-25
add shorter flag for quiet and debug

michiboo:
cli: add short versions of quiet and debug flag #811


JayjeetAtGithub:
@michiboo Nice work ! Thanks a lot.

PR #826, opened 2020-05-06
Kubernetes Documentation

edeediong:


docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,65 @@ popper run -f sample.yml -c config.yml
 
 This runs the step `one` locally in the host and step `two` through slurm on 2 nodes.
 
+### Kubernetes
+
+Popper can be run on a cluster of multiple containers using Kubernetes as the fundamental resource manager to orchestrate the execution of a step to several nodes.
```
ivotron, line 364: - "cluster of multiple containers" is a bit confusing. maybe change to "Kubernetes cluster" and then add a hyperlink to kubernetes.

- we can remove "fundamental"

- a step runs on a single node, so "execution of a step to several nodes" is incorrect

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,65 @@ popper run -f sample.yml -c config.yml
...
+You can get started with running Popper workflows through K8s(another way to represent Kubernetes) by following the example below.
```
ivotron, line 365: we can remove the k8s reference and leave only "Kubernetes" to avoid confusion

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,65 @@ popper run -f sample.yml -c config.yml
...
+```
+
+To run all the steps of the workflow through kubernetes,
+use the `--resource-manager` or `-r` option of the `popper run` subcommand to specify the resource manager.
+
+```bash
+popper run -f sample.yml -r kubernetes
+```
+
+When this command is run, what happens underneath is the following: `stepname kubernetes run namespace podname`
```
ivotron, line 387: that's really not a command, it's only a message shown to the user. So maybe we can remove this sentence.

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,65 @@ popper run -f sample.yml -c config.yml
...
+
+Things to note:
+
+* stepname : is the step id of the workflow
+* kubernetes run: command to run kubernetes
+* namespace: Namespaces are a way to divide cluster resources between multiple users (via resource quota).
+* podname: Group of containers
```
ivotron, line 394: We can also remove this list. What we can do is give a high-level description of what happens when a workflow runs on kubernetes. Something like:

> For a workflow, we first create a volume in the namespace associated to the kubctl config. Then, for each step in the workflow, we create a pod in the namespace, run it, and destroy it. If the step is referencing a dockerfile (instead of an image), we build it locally, push it to a registry so that kubernetes can pull it from the registry when it executes the corresponding

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,65 @@ popper run -f sample.yml -c config.yml
...
+
+You can also have more control over the steps in the kubernetes resource manager.
```
ivotron, line 396: it's not clear what it's meant with "have more control over the steps".

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,65 @@ popper run -f sample.yml -c config.yml
...
+  name: docker
+  options:
+    privileged: True
+    hostname: example.local
+
+resource_manager:
+  name: kubernetes
+  options:
+    two:
+      nodes: 2
```
ivotron, line 410: these options are not supported by kubernetes. We currently don't support passing any options to kubernetes

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,65 @@ popper run -f sample.yml -c config.yml
...
+* namespace: Namespaces are a way to divide cluster resources between multiple users (via resource quota).
+* podname: Group of containers
+
+You can also have more control over the steps in the kubernetes resource manager.
+To do that we create a config file `config.yml` with the specifications as shown below.
+
+```yaml
+engine:
+  name: docker
+  options:
```
ivotron, line 402: these options for docker are not really needed, since the workflow doesn't run in docker . So we can remove the `options` part.

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,65 @@ popper run -f sample.yml -c config.yml
...
+      nodes: 2
+```
+
+Now, we execute `popper run` with this config file as follows:
+
+```bash
+popper run -f sample.yml -c config.yml
+```
+
+This runs the step `one` locally in the host and step `two` through kubernetes on 2 nodes.
```
ivotron, line 419: - "locally in the host" is a bit misleading.

- "through kubernetes on 2 nodes" that is not correct

edeediong:
@ivotron I'd make the necessary changes. Also I do apologize for the pep8 messages. I have deleted the commit from history though.

edeediong:
> thanks @edeediong, can you please configure your IDE so that it doesn't remove spaces at the end of a line? And once that's done, then could you please revert that change (so that those are not shown as changes)?
> 
> Other than that, this is looking good, left a couple of comments.

I have configured my IDE to not removing trailing spaces. Sorry for that really. Let me try and revert the changes

docs/sections/cn_workflows.md
```diff
@@ -385,7 +385,22 @@ use the `--resource-manager` or `-r` option of the `popper run` subcommand to sp
 popper run -f sample.yml -r kubernetes
 ```
 
-For a workflow we build an image and create a volume in the namespace defined in the kubefig. This volume is the with a ReadWrite access.
+For a workflow we build an image and create a volume in the namespace defined in the [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/).
```
ivotron, line 388: the narrative style of the documentation is in third person. So this would be "Given a workflow such as the one shown above, Popper builds an image and creates a volume..."

docs/sections/cn_workflows.md
```diff
@@ -385,7 +385,22 @@ use the `--resource-manager` or `-r` option of the `popper run` subcommand to sp
 popper run -f sample.yml -r kubernetes
 ```
 
-For a workflow we build an image and create a volume in the namespace defined in the kubefig. This volume is the with a ReadWrite access.
+For a workflow we build an image and create a volume in the namespace defined in the [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/). 
+A sample kubeconfig looks like this:
```
ivotron, line 389: we can skip explaining what a `kubconfig` file is and instead only mention that the kubernetes runner makes use of the same authentication mechanism of `kubectl`, which is via the `kubeconfig` file (and then we add a link to the kubernetes documentation where this is explained)

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,55 @@ popper run -f sample.yml -c config.yml
 
 This runs the step `one` locally in the host and step `two` through slurm on 2 nodes.
 
+## Kubernetes
+
+Popper can be run [Kubernetes cluster](https://kubernetes.io/docs/setup/) as the resource manager to orchestrate the execution of a step to one a single node.
```
ivotron, line 364: This sentence needs a bit of work. Maybe something like "Popper can execute a workflow using a Kubernetes cluster as the underlying resource manager".

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,55 @@ popper run -f sample.yml -c config.yml
...
+You can get started with running Popper workflows through Kubernetes by following the example below.
```
ivotron, line 365: "running Popper workflows **on** Kubernetes"

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,55 @@ popper run -f sample.yml -c config.yml
...
+preferences: {}
+contexts:
+- context:
+    cluster: development
+    namespace: sample
+    user: developer
+  name: dev
+```
+
+The volume created is configured with a ReadWrite access.
```
ivotron, line 403: "with `ReadWrite` access"

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,55 @@ popper run -f sample.yml -c config.yml
...
+Then for each step in the workflow, we create a pod in the namespace using information from the step and pass commands args defined in the step.
+The pod is mounted to `/workspace` and it's logs are read.
```
ivotron, line 405: We can say that the pod mounts the volume for the workflow in the `/workspace` folder and uses that folder as the working directory, similarly to how it is done when the workflow runs locally.

Not sure what "it's logs are read", but we can say something like: "The output of the step being executed is streamed to the the machine where `popper run` is executed".

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,55 @@ popper run -f sample.yml -c config.yml
...
+Finally, the pods are exited and deleted.
```
ivotron, line 406: - maybe "After the step finishes executing, the pod [singular] is destroyed"

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,55 @@ popper run -f sample.yml -c config.yml
...
+```
+
+To run all the steps of the workflow through kubernetes,
+use the `--resource-manager` or `-r` option of the `popper run` subcommand to specify the resource manager.
+
+```bash
+popper run -f sample.yml -r kubernetes
+```
+
+For a workflow we build an image and create a volume in the namespace defined in the [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/).
```
ivotron, line 388: another important point is to mention that the volume that gets created is persisted throughout the execution of the workflow.

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,55 @@ popper run -f sample.yml -c config.yml
...
+  name: dev
+```
+
+The volume created is configured with a ReadWrite access.
+Then for each step in the workflow, we create a pod in the namespace using information from the step and pass commands args defined in the step.
+The pod is mounted to `/workspace` and it's logs are read.
+Finally, the pods are exited and deleted.
+The next step if available follows a similar pattern: create pod, run and destroy.
+
+If the step is referencing a Dockerfile (instead of an image), we build it locally, push it to a registry so that kubernetes can pull it from the registry and then.
```
ivotron, line 409: "if the step is referencing a Dockerfile instead of an image (as shown in step `two` of the example), Popper builds...."

docs/sections/cn_workflows.md
```diff
@@ -359,6 +359,55 @@ popper run -f sample.yml -c config.yml
...
+
```
ivotron, line 410: one last thing is important to mention is that the volume that was created does NOT get destroyed. It's up to users to decide what to do with it. We can add a link to the kubernetes documentation that shows how to deal with volumes using `kubctl`

docs/sections/cn_workflows.md
edeediong: Awesome! I'd take note of this

docs/sections/cn_workflows.md
edeediong: Noted, I'd make adequate changes

docs/sections/cn_workflows.md
edeediong: Oh I didn't know this. Thanks for adding this

docs/sections/cn_workflows.md
edeediong: Third-person explanation here too. I'd take note of that

docs/sections/cn_workflows.md
edeediong: Does the pod mount the volume *for* the workflow or *of* the workflow?

docs/sections/cn_workflows.md
edeediong: Nice! This way I don't think I will need to include _orchestrate the execution of a step to run a single node_. Or do you think it is important to include it?

docs/sections/cn_workflows.md
```diff
@@ -0,0 +1,415 @@
...
+Now, we execute `popper run` with this config file as follows:
+
+```bash
+popper run -f sample.yml -c config.yml
+```
+
+This runs the step `one` locally in the host and step `two` through slurm on 2 nodes.
+
+## Kubernetes
```
edeediong, line 362: @ivotron I'm trying to ensure this is stored somewhere safe. So when @JayjeetAtGithub finishes with the Kubernetes runner we can be sure there's a documentation for it. Also, what's the hold up with the Kubernetes resource manager?

edeediong:
Closing this now and it will be resurrected when the Kubernetes runner is ready

Issue #809, opened 2020-04-22
cli: fail with proper error message workflow file if workflow file is empty


JayjeetAtGithub:
Let's take an empty file, `wf.yml`. On doing,

```bash
touch wf.yml
popper run -f wf.yml
```

we get a long and verbose stacktrace like this,

```bash
Traceback (most recent call last):
  File "/Users/jayjeetchakraborty/mac-venv/bin/popper", line 8, in <module>
    sys.exit(cli())
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/popper/commands/cmd_run.py", line 164, in cli
    include_step_dependencies=with_dependencies)
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/popper/parser.py", line 112, in new
    wf = YMLWorkflow(file)
  File "/Users/jayjeetchakraborty/mac-venv/lib/python3.7/site-packages/popper/parser.py", line 555, in __init__
    self.wf_list = yaml.safe_load(fp)['steps']
TypeError: 'NoneType' object is not subscriptable
```

It would be great if we would have a `log.fail()` with a message like,

```
ERROR: Workflow file {wfile} is empty.
```

[...]

PR #830, opened 2020-05-08
changes for #809 (error message for empty workflow)

nicolecastillo:
Handling error message for empty workflow and adds the test cases for it.

### DIFF
--- a/cli/popper/parser.py
+++ b/cli/popper/parser.py
@@ -556,7 +556,7 @@ class YMLWorkflow(Workflow):
 
             if not dict_from_yaml:
                 log.fail(f'Workflow string is empty.')
-            
+
             # try to load string
             self.wf_list = yaml.safe_load(wfile)['steps']
         else:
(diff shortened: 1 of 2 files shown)


edeediong:
@nicolecastillo Thanks for working on this. Before your PR is reviewed you'd need to fix the failing builds. I checked the logs and it seems you haven't formatted the codes to conform with black. [This](https://github.com/systemslab/popper/blob/master/CONTRIBUTING.md#codestyle) explains what you need to do to fix that. Cheers!

nicolecastillo:
@edeediong Thank you, that was the problem!

JayjeetAtGithub:
@nicolecastillo Thanks a lot. Maybe we can take this opportunity to add a file reader abstraction, maybe something like a `read_file(path, type)` function in `utils.py`, which would be used to read both the workflow file, i.e. `wf.yml` and config file, i.e. `config.yml`. The abstraction would: 

1. Check if the path is `not None`.
2. Check if the extension of the file requested is `== type`.
3. Check if the file exists.
4. Check if the file is empty.

If all the checks pass, read the contents into a `dict` and return.

nicolecastillo:
@JayjeetAtGithub Sure, I'll look into it and work on that.

ivotron:
thanks a lot @nicolecastillo , let us know if you have any question!

ivotron:
the recent changes on `master` are causing this to be conflicted. So maybe it would be better to close it and port the changes to the tip of `master` in order to address #809. @nicolecastillo if it works better for you, maybe we can also break this in two, and work on what Jayjeet and Eddie suggested separately as another PR.

PR #831, opened 2020-05-13
Changes to dev version output

FernandaDguez:
- The version command output now includes the commit in development mode
- Added `--version` option


cli/popper/__init__.py
```diff
@@ -1 +1,21 @@
...
+# check if dunamai is available
+dunamai_spec = importlib.util.find_spec("dunamai")
+dunamai_found = dunamai_spec is not None
+if dunamai_found:
+    # if dunamai is found, then we use it to display the version
+    import dunamai
+
+    __version__ = dunamai.Version.from_any_vcs().serialize()
+    ver = f'__popper_version__ = "{__version__}"'
+    path = os.path.split(os.getcwd())
```
ivotron, line 13: since `popper version` might not be executed from the popper repository folder always, the value returned by `os.getcwd()` doesn't necessarily correspond to the project root directory. Instead, what we can do is obtain the path to the current script, which in python 3 can be done by doing:

```python
import pathlib
_init_script_dir = pathlib.Path(__file__).parent.absolute()
```

So the `_init_script_dir` will always point to the folder containing the `__init__.py` script.

cli/popper/_version.py
```diff
@@ -0,0 +1 @@
+__popper_version__ = "2.5.1"
```
ivotron, line 1: this could be assigned to `0.0.0`, and then the tests we execute could check that the value is not `0.0.0` but instead is the value that `dunamai` returns.

cli/test/test_cmd_version.py
```diff
@@ -13,5 +13,5 @@ def test_version(self):
             runner = CliRunner()
             result = runner.invoke(cmd_version.cli)
             self.assertEqual(result.exit_code, 0)
-
+            self.assertEqual(__version__, __popper_version__)
```
ivotron, line 16: if we make the default be `0.0.0`, we can add one more assertion that tests that the value of the version is not `0.0.0`.

ivotron:
thanks @FernandaDguez, and congrats on your first merged PR! 🎉

PR #833, opened 2020-05-15
Error message for empty workflow file (#809)

nicolecastillo:
- Adds the error message when the workflow file is empty
- Adds the test


ivotron:
thanks a lot 😄

PR #838, opened 2020-05-22
cli: conditional container creation fix

vipulchhabra99:


cli/popper/runner_host.py
```diff
@@ -131,10 +131,12 @@ def run(self, step):
...
         container = self._find_container(cid)
-        if container and not self._config.reuse and not self._config.dry_run:
-            container.remove(force=True)
 
-        container = self._create_container(cid, step)
+        if not self._config.reuse and not self._config.dry_run:
+            if container:
+                container.remove(force=True)
+
+            container = self._create_container(cid, step)
```
ivotron, line 139: the `_create_container` function becomes a noop (only prints to the log) when `self._config.reuse` is `True`. So maybe we can avoid nested conditionals by doing:

```python
if not container and self._config.reuse:
    log.fail(f"Cannot find an existing container for step '{step.id}' to be reused")

if container and not self._config.reuse and not self._config.dry_run:
    container.remove(force=True)

if not container and not self._config.reuse:
    # create_container does not run if dry-run is true
    container = self._create_container(cid, step)
```

ivotron: oh, and we need to deal with conflicting request of `--reuse` but no container exists

vipulchhabra99: I think if the container already exists and the reuse flag is not used then we need to remove the container and create a new one.

cli/popper/runner_host.py
```diff
@@ -131,10 +131,18 @@ def run(self, step):
...
         container = self._find_container(cid)
+
+        if not container and self._config.reuse:
+            log.fail(
+                f"Cannot find an existing container for step '{step.id}' to be reused"
+            )
+
         if container and not self._config.reuse and not self._config.dry_run:
             container.remove(force=True)
+            container = self._create_container(cid, step)
```
ivotron, line 142: we can make `container = None`

vipulchhabra99: Yes works.

ivotron:
thanks @vipulchhabra99 !

Issue #847, opened 2020-06-04
cli: add 'dir' attribute for steps


ivotron:
allow a step to specify the working dir where it should run. For example:

```yaml
steps:

- uses: docker://alpine:3.11
  dir: /etc/
  args: ['pwd']
```

and the above should show `/etc` instead of the default which is `/workspace`.

Three things that need to be done:
  * [ ] the [YAML spec](https://github.com/getpopper/popper/blob/master/cli/popper/parser.py#L24) needs to be updated so a new `dir` attribute is valid
  * [ ] a new test for this can be added in the [`cli/test/test_parser.py`](https://github.com/getpopper/popper/blob/master/cli/test/test_parser.py).
  * [ ] document this feature in the [Workflow Syntax](https://github.com/getpopper/popper/blob/master/docs/sections/cn_workflows.md#syntax) and the [CLI Features](https://github.com/getpopper/popper/blob/master/docs/sections/cli_features.md) sections.

RicardoHE97:
Can I do this issue?

ivotron:
Please go ahead! Let us know if you have any questions

RicardoHE97:
Hi @ivotron, I already submitted a PR #853 but I should point out that I found an problem in my implementation to solve this issue. 

If I define `dir` as `/etc/` and if the container uses a curl, this will throw the following error

```
curl: (77) error setting certificate verify locations:
  CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: none
```
**How to recreate this error**
Using the following workflow will trigger this error
```yaml
steps:
- id: download CSV file with data on global CO2 emissions
  uses: docker://byrnedo/alpine-curl:0.1.8
  args: [-LO, https://github.com/datasets/co2-fossil-global/raw/master/global.csv]
  dir: /tmp/
```

**Expected result**
A clean download of the csv file that has the url that was passed in the arguments for the step.


Where I could trace this error comes from is from the file `runner_host.py` in the function `run` from the DockerRunner class. 

I have some idea why this is happening but I don't know for sure and I couldn't be able to fix it.

PR #853, opened 2020-06-10
Added 'dir' attribute for steps on workflow #847

RicardoHE97:


docs/sections/cn_workflows.md
```diff
@@ -188,6 +190,21 @@ cd /path/to
 popper run -f wf.yml
 ```
 
+To specify a workspace directory for a step you can adding `dir` in the
```
ivotron, line 193: since this is distinct to the workspace, maybe we can have this as a subsection titled `Changing the working directory`, and explain that this is only changing the working directory where the specified command is being executed from.

And we can also add a note that reminds the reader that if the directory is outside the workspace, then anything that gets written to it won't be persisted.

ivotron:
oh, also, can you please rebase master? there were some changes that were added recently that are conflicting with the changes in this PR

docs/sections/cn_workflows.md
RicardoHE97: Got it! I'll change it.

docs/sections/cn_workflows.md
```diff
@@ -211,7 +211,26 @@ popper run -f /home/user/proj/wf.yml -w /home/user/proj/
 
 The above writes the `/home/user/proj/myfile` even though Popper is 
 being invoked from `/tmp`, since the `-w` flag is being passed to 
-`ppopper run`.
+`popper run`.
+
+### Changing the working directory
+
+To specify a working directory for a step you can use the `dir` attribute
+in the workflow. This will only going to change where the specified
```
ivotron, line 219: "this changes where..." or "this is going to change where..."

docs/sections/cn_workflows.md
```diff
@@ -211,7 +211,26 @@ popper run -f /home/user/proj/wf.yml -w /home/user/proj/
...
+command is executed.
+
+For example, adding `dir` to a workflow would looks like the following:
```
ivotron, line 222: this can be ".. to a workflow would look like the following:" or "to a workflow results in the following:"

docs/sections/cn_workflows.md
```diff
@@ -211,7 +211,26 @@ popper run -f /home/user/proj/wf.yml -w /home/user/proj/
...
+
+```yaml
+version: '1'
+steps:
+- uses: docker://alpine:3.9
+  args: [touch, ./myfile]
+  dir: /path/to/dir/
+```
+
+It is worth mentioning that if the directory specified is outside the workspace,
```
ivotron, line 232: we can make it more explicit: "is specified outside the `/workspace` folder, then..."

docs/sections/cn_workflows.md
```diff
@@ -211,7 +211,26 @@ popper run -f /home/user/proj/wf.yml -w /home/user/proj/
...
+then anything that gets written to it won't persist.
```
ivotron, line 233: "won't persist (see below for more)."

RicardoHE97: This would have a link to section "Filesystem namespaces and persistence" or is this not necessary?

ivotron: since this is referring to the next section (so it's literally below), it's OK to not put a link.

RicardoHE97: Got it! Thanks for the clarification.

ivotron:
thanks a lot @RicardoHE97 , really appreciated 🙏

Issue #851, opened 2020-06-09
cli: add --allow-undefined-secrets flag to 'popper run'


ivotron:
Add a new `--allow-undefined-secrets` flag allows the execution of a workflow when environment variables used in `secrets` are not defined. This is useful in scenarios such as CI workflows in travis forks, where variables from the upstream repository are [not available in their forks](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml) and thus are undefined in the  (see https://github.com/uccross/skyhookdm-ceph-cls/pull/41 for an example).

For this we need to:
  * [ ] add a new flag to the `popper run` command by adding it to [`src/popper/commands/cmd_run.py`](https://github.com/getpopper/popper/blob/master/cli/popper/commands/cmd_run.py)
  * [ ] add a new `allow_undefined_secrets` option to be read by the [configuration loader](https://github.com/systemslab/popper/blob/master/cli/popper/config.py).
  * [ ] add a conditional to [the method that obtains the environment variables](https://github.com/getpopper/popper/blob/master/cli/popper/runner.py#L60) so that it checks that this configuration is enabled, in which case, it can ignore it and not throw an error.
  * [ ] add a [new test](https://github.com/getpopper/popper/blob/master/cli/test/test_runner.py#L21).

PR #854, opened 2020-06-12
add "help" command

### LINKED ISSUE #787: cli: add 'help' command
add a `help` command with the following spec:

```bash
popper help [subcommand]
```

When `[subcommand]` is not given, it is equivalent to the current `popper --help`. When the `[subcommand]` argument is given, it is equivalent to `popper subcommand --help`.

For this, the help message doesn't have to be recreated, but instead we can invoke Click's [`get_help()`](http://click.palletsprojects.com/en/7.x/api/#click.Command.get_help) function of each subcommand.

fsr313:
fixes #787

### DIFF
--- a/src/popper/commands/cmd_help.py
+++ b/src/popper/commands/cmd_help.py
@@ -1,0 +1,23 @@
+import click
+import os
+
+from popper.cli import pass_context, log, PopperCLI, cmd_folder
+
+popper_cli = PopperCLI()
+
+
+@click.command("help", short_help="shows help for a given command")
+@click.argument(
+    "subcommand",
+    type=click.Choice(popper_cli.list_commands(click.Context(popper_cli))),
+    required=False,
+)
+@pass_context
+def cli(ctx, subcommand):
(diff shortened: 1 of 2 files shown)


JayjeetAtGithub:
Hi @fsr313. Thanks a lot for working on this. Can you please add unit tests for this feature? You can look at the [`popper version`](https://github.com/getpopper/popper/blob/master/src/test/test_cmd_version.py) tests as a reference.

fsr313:
yes can do

fsr313:
@JayjeetAtGithub may I get some help understanding why Travis failing the checks?
This is the log where the check fails
```
$ popper run
ERROR: Secret DOCKER_USERNAME not defined
The command "popper run" exited with 1.
```

ivotron:
thanks @fsr313, and apologies for that terrible contributor experience. This is due to the way Travis deals with secrets (see #851). I see that tests are passing fine, so I'll merge as that error doesn't have to do with your changes.

PR #857, opened 2020-06-15
map env vars into the popper container from current shell

### LINKED ISSUE #856: cli: secrets not available in containers while running popper installed by the installation script
## How to reproduce?

1. Install `popper` using the installation script
2. Take a workflow (`wf.yml`) like the one shown below,
```
steps:
- uses: "popperized/bin/sh@master"
  args: ["echo $NAME"]
  secrets: ["NAME"]

- uses: "docker://alpine:3.11"
  args: ["ls"]
```
As the workflow requires a secret `NAME` to be defined in the current shell, define it by doing,
```
$ export NAME=xxx
```
3. Now, try running the workflow
```
$ popper run -f wf.yml
```
It would be found that although the NAME environment variable is defined in the current shell, popper does not detect it and prompts the user to enter the value of the secret which should not happen. Maybe this is because currently, the `docker run` command in the executable script does not bind the environment variables from the current shell.

JayjeetAtGithub:
Fix #856 

Tested in Centos, Ubuntu and Mac

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -12,10 +12,14 @@ command -v docker >/dev/null 2>&1 || { echo >&2 "docker command not found. Abort
 
 cat > ./popper << "EOF"
 #!/usr/bin/env sh
+
+printenv > /tmp/.envfile
+
 docker run --rm -ti \
   --volume /var/run/docker.sock:/var/run/docker.sock \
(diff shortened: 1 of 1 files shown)


ivotron:
thanks a lot!

PR #860, opened 2020-06-17
add `GIT_REMOTE_ORIGIN_URL` to runner.py and update test case

### LINKED ISSUE #852: cli: add GIT_REMOTE_ORIGIN_URL variable
Currently, Popper makes the `GIT_COMMIT`, `GIT_BRANCH` and `GIT_SHA_SHORT` variables available for steps whenever `popper run` is executed inside a git repository. For this issue, we will also add support for defining a `GIT_REMOTE_ORIGIN_URL` variable that contains the URL of the remote from which the repository was cloned from.

For this, we can make use of the `git config --get remote.origin.url` functionality via GitPython (`repo.remotes.origin.url` property of a `git.Repo` object), so that there is a new `get_remote_origin_url()` function in the [`src/popper/scm.py`](https://github.com/getpopper/popper/blob/master/src/popper/scm.py)

We can also add tests for this on the [`src/test/test_scm.py`](https://github.com/getpopper/popper/blob/master/src/test/test_scm.py) file.

DanielOsunaV:
Fixes #852

### DIFF
--- a/src/popper/runner.py
+++ b/src/popper/runner.py
@@ -225,6 +225,7 @@ class StepRunner(object):
                     "GIT_COMMIT": self._config.git_commit,
                     "GIT_BRANCH": self._config.git_branch,
                     "GIT_SHA_SHORT": self._config.git_sha_short,
+                    "GIT_REMOTE_ORIGIN_URL": scm.get_remote_url(self._config.repo),
                 }
             )
         return step_env
--- a/src/popper/runner.py
+++ b/src/popper/runner.py
@@ -225,7 +225,7 @@ class StepRunner(object):
                     "GIT_COMMIT": self._config.git_commit,
                     "GIT_BRANCH": self._config.git_branch,
                     "GIT_SHA_SHORT": self._config.git_sha_short,
-                    "GIT_REMOTE_ORIGIN_URL": scm.get_remote_url(self._config.repo),
+                    "GIT_REMOTE_ORIGIN_URL": self._config.git_remote_origin_url,
                 }
             )
         return step_env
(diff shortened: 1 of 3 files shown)


luislizarraga:
Looks good! 👌

src/popper/runner.py
```diff
@@ -225,6 +225,7 @@ def _prepare_environment(self, step, env={}):
                     "GIT_COMMIT": self._config.git_commit,
                     "GIT_BRANCH": self._config.git_branch,
                     "GIT_SHA_SHORT": self._config.git_sha_short,
+                    "GIT_REMOTE_ORIGIN_URL": scm.get_remote_url(self._config.repo),
```
ivotron, line 228: in order to be consistent with the way in which the other variables (commit, branch, etc) are defined/used, we need to obtain this information as part of the `ConfigLoader.load(https://github.com/getpopper/popper/blob/master/src/popper/config.py)` function, in particular [here](https://github.com/getpopper/popper/blob/master/src/popper/config.py#L56). In this way, we only read the value instead of obtaining the URL, as it is done for the other variables.

src/test/test_runner.py
```diff
@@ -125,6 +126,7 @@ def test_prepare_environment_with_git(self):
                 "GIT_COMMIT": conf.git_commit,
                 "GIT_BRANCH": conf.git_branch,
                 "GIT_SHA_SHORT": conf.git_sha_short,
+                "GIT_REMOTE_ORIGIN_URL": scm.get_remote_url(repo),
```
ivotron, line 129: the same comment applies here, that is, we would like to not invoke `popper.scm` directly but instead read the value from the config object.

ivotron:
thanks a lot 🙏

PR #861, opened 2020-06-18
Update docs for custom container configuration

FernandaDguez:
Update "Customizing container engine behavior" section in the `docs/sections/cli_features.md` file with instructions to use a YAML file instead of a python dictionary


ivotron:
thank you!

PR #867, opened 2020-06-19
add id sanitization

### LINKED ISSUE #849: cli: Custom Dockerfile fails with upper-case step ID
# Steps to reproduce
```bash
git init myrepo
cd myrepo
# Create "wf.yml" with the content below
mkdir popper
echo "FROM alpine:latest" > mydocker/Dockerfile
popper run -f wf.yml
```
`wf.yml` looks like this:
```yaml
steps:
- id: Do Something # At least one capital letter.
  uses: ./mydocker
  args: [echo, "hello world"]
```
## Error message
```
[Do Something] docker build popper_Do_Something_step:0e17abb /home/user/myrepo/./mydocker
ERROR: 500 Server Error: Internal Server Error ("invalid reference format: repository name must be lowercase")
```
If the step ID is all lower-case (`id: do something`), there is no error.
## Versions
- Popper version: Current `master`, in virtual environment with Python 3.8.3
- Operating system: Arch Linux
- Docker version 19.03.11-ce, build 42e35e61f3

# Expected behavior
- Popper warns the user if an ID in the workflow file is not all lowercase, or
- Popper converts the ID to lowercase when calling Docker.

fsr313:
fixes #849 

add casting to lowercase for img and modify test to include upper-case case

### DIFF
--- a/src/popper/parser.py
+++ b/src/popper/parser.py
@@ -24,7 +24,7 @@ class WorkflowParser(object):
                         "type": "map",
                         "mapping": {
                             "uses": {"type": "str", "required": True},
-                            "id": {"type": "str", "pattern": "^[a-z0-9]+"},
+                            "id": {"type": "str"},
                             "args": {"type": "seq", "sequence": [{"type": "str"}]},
                             "runs": {"type": "seq", "sequence": [{"type": "str"}]},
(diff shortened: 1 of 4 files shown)


ivotron:
sorry for the delay, can you please rebase master? thanks a lot for this PR 🙏

PR #870, opened 2020-06-24
created 'GIT_TAG' variable

DanielOsunaV:
this is the first approach to the git_tag variable, I might being dismissing some scenarios in the [get_tag](https://github.com/DanielOsunaV/popper/blob/0275cc4ad603f6f8149d92fa9896a7d20e7bac2b/src/popper/scm.py#L203) function, If you find some dismissing scenarios let me know and I'll be happy to make the changes onto the function

### DIFF
--- a/src/popper/scm.py
+++ b/src/popper/scm.py
@@ -198,3 +198,11 @@ def get_branch(repo):
         return branch
 
     return get_sha(repo)
+
+
+def get_tag(repo=None):
+    """Obtains the git tag fot the currently checked our version on the given repository object
+        if possible, otherwise returns ""
+    """
+    if repo:
+        return repo.describe('--tags', '--dirty','--broken', '--long')
--- a/src/test/test_runner.py
+++ b/src/test/test_runner.py
@@ -126,6 +126,7 @@ class TestStepRunner(PopperTest):
                 "GIT_BRANCH": conf.git_branch,
                 "GIT_SHA_SHORT": conf.git_sha_short,
                 "GIT_REMOTE_ORIGIN_URL": conf.git_remote_origin_url,
+                "GIT_TAG": conf.git_tag,
             }
             self.assertDictEqual(expected, env)
             os.environ.pop("A")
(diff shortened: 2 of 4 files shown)


src/popper/scm.py
```diff
@@ -198,3 +198,11 @@ def get_branch(repo):
         return branch
 
     return get_sha(repo)
+
+
+def get_tag(repo=None):
+    """Obtains the git tag fot the currently checked our version on the given repository object
+        if possible, otherwise returns ""
```
ivotron, line 205: when a tag is not found, this is returning None rather than an empty string, which is fine, but we can update the comment to reflect that

src/popper/scm.py
```diff
@@ -198,3 +198,11 @@ def get_branch(repo):
...
+    """
+    if repo:
+        return repo.describe('--tags', '--dirty','--broken', '--long')
```
ivotron, line 208: sorry for not explaining this better in the issue description. The idea for having a `GIT_TAG` variable is that it would be defined only if the current commit is tagged, so this could work:

```
git tag --points-at $(git rev-parse HEAD)
```

(we'd need to find the way of doing the above with GitPython).

The other issue to keep in mind is that in some scenarios (e.g. in Travis), tags are not fetched when doing a checkout, and instead they are obtained via github and defined like `TRAVIS_TAG`. So if `git tag --points-at` doesn't point return anything, we can check for `CIRCLE_TAG`, `TRAVIS_TAG`, etc. as it is done currently for `git_branch()` so that we can obtain the tag from the CI env variables.

src/test/test_runner.py
```diff
@@ -126,6 +126,7 @@ def test_prepare_environment_with_git(self):
                 "GIT_BRANCH": conf.git_branch,
                 "GIT_SHA_SHORT": conf.git_sha_short,
                 "GIT_REMOTE_ORIGIN_URL": conf.git_remote_origin_url,
+                "GIT_TAG": conf.git_tag,
```
ivotron, line 129: we also need to tag the HEAD of the branch we created, and then check that that's what we're getting back

DanielOsunaV:
I'll make the needed research :)

edeediong:
> I'll make the needed research :)

If you run into any issues or you don't understand anything don't fail to reach out @DanielOsunaV

PR #874, opened 2020-06-30
Podman runner

edeediong:
This contains the necessary code changes to enable popper run workflows on Podman


edeediong:
> this is looking really great, really cool!!
> 
> one other thing it is required is to add new items in the Travis matrix, where we add `ENGINE=podman`, in the same way, we do it for singularity.

Thanks a lot! Also, could you help me check what is failing my tests? Seems to be an interactive issue

src/test/test_runner_host.py
```diff
@@ -304,6 +304,216 @@ def test_docker_basic_run(self):
...
+            )
+            _, _, c2 = HostRunner._exec_cmd(
+                cmd + ["alpine:3.9", "sleep", "10000", "-q"], logging=False
+            )
+            c1 = c1.rstrip()
+            c2 = c2.rstrip()
+            pr._spawned_containers.add(c1)
+            pr._spawned_containers.add(c2)
+            pr.stop_running_tasks()
+            c1_status_cmd = [
```
ivotron, line 326: you could instead single an issue command that asks for the containers that are running and then assert that that is empty

src/test/test_runner_host.py
```diff
@@ -304,6 +304,216 @@ def test_docker_basic_run(self):
...
+            _, _, c1 = HostRunner._exec_cmd(
+                cmd + ["debian:buster-slim", "sleep", "10000", "-q"], logging=False,
+            )
+            _, _, c2 = HostRunner._exec_cmd(
+                cmd + ["alpine:3.9", "sleep", "10000", "-q"], logging=False
+            )
+            c1 = c1.rstrip()
+            c2 = c2.rstrip()
+            pr._spawned_containers.add(c1)
+            pr._spawned_containers.add(c2)
```
ivotron, line 324: and also, before you proceed with stopping the containers, you could ensure that they're running

edeediong: I'd check if their running before stopping them. Thanks for that

src/popper/runner_host.py
```diff
@@ -293,6 +293,212 @@ def _update_with_engine_config(self, container_args):
...
+        if self._config.dry_run:
+            return 0
+
+        self._spawned_containers.add(container)
+
+        try:
+            cmd = ["podman", "start", container]
+            HostRunner._exec_cmd(cmd, logging=False)
+
+            cmd = ["podman", "wait", container]
```
ivotron, line 345: this shouldn't be needed. The previous command should block and not continue until the container exits

src/popper/runner_host.py
```diff
@@ -293,6 +293,212 @@ def _update_with_engine_config(self, container_args):
...
+        if "hostname" in container_args and container_args["hostname"]:
+            cmd.extend(["-h", container_args["hostname"]])
+        if "tty" in container_args and container_args["tty"]:
+            cmd.extend(["-t", container_args["tty"]])
+        if "domainname" in container_args and container_args["domainname"]:
+            cmd.extend(["--domainname", container_args["domainname"]])
+        if "image" in container_args and container_args["image"]:
+            cmd.append(container_args["image"])
+
+        _, _, container = HostRunner._exec_cmd(cmd, logging=False)
```
ivotron, line 496: we need to check for failure, as `HostRunner._exec_cmd` doesn't throw an exception if the command failed

src/popper/runner_host.py
```diff
@@ -293,6 +293,212 @@ def _update_with_engine_config(self, container_args):
...
+        msg += f' image={container_args["image"]}'
+        if container_args["entrypoint"]:
+            msg += f' entrypoint={container_args["entrypoint"]}'
+        if container_args["command"]:
+            msg += f' command={container_args["command"]}'
+        log.info(msg)
+
+        cmd = ["podman", "create"]
+        if "command" in container_args and container_args["command"]:
+            cmd.append(container_args["command"])
```
ivotron, line 473: we can avoid using if statements here by doing:

```python
cmd.append(container_args.get("command", ""))
```

src/popper/runner_host.py
```diff
@@ -293,6 +293,212 @@ def _update_with_engine_config(self, container_args):
...
+
+        if self._config.dry_run:
+            return
+
+        container_args = self._get_container_kwargs(step, f"{img}:{tag}", cid)
+
+        log.debug(f"Container args: {container_args}")
+
+        msg = f"[{step.id}] podman create name={cid}"
+        msg += f' image={container_args["image"]}'
```
ivotron, line 464: to be consistent with the code bellow, this can do `msg.append()` instead of `msg +=...`

src/popper/runner_host.py
```diff
@@ -293,6 +293,212 @@ def _update_with_engine_config(self, container_args):
...
+        if container_args["entrypoint"]:
+            msg += f' entrypoint={container_args["entrypoint"]}'
```
ivotron, line 466: we can avoid doing an if statement here if we do:

```python
msg.append(container_args.get("entrypoint", ""))
```

ivotron:
also, you can remove the `Draft` status of this PR, as this is not a draft anymore 😄

src/test/test_runner_host.py
```diff
@@ -304,6 +304,205 @@ def test_docker_basic_run(self):
...
+                    "environment": {"FOO": "bar"},
+                    "entrypoint": None,
+                    "detach": False,
+                    "stdin_open": True,
+                    "tty": True,
+                    "privileged": True,
+                    "hostname": "popper.local",
+                    "domainname": "www.example.org",
+                },
+            )
```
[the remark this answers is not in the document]
ivotron: also, in order to get the above working, I tweaked the `cmd_run.py` file so that I could pass `--engine` to `popper run`

ivotron:
can you please rebase `master`, I pushed a fix to bump up to use the bionic image, as well as to deal with the skip of pushing images to dockerhub from forks

src/scripts/install_podman.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -ex
+
+if [ "$ENGINE" == "podman" ]; then
+    . /etc/os-release
+    echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/testing/xUbuntu_${VERSION_ID}/ /" | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:testing.list
+    curl -L https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/testing/xUbuntu_${VERSION_ID}/Release.key | sudo apt-key add -
+    sudo apt-get update -qq
+    sudo apt-get -qq -y install slirp4netns podman
```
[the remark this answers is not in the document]
edeediong: Oh, it's needed when installing podman in the travis environment. Both debian and alpine try to find it but don't see it hence trigger an error. Installing that into the environment fixes that.

edeediong:
Hi Ivo! I've taken a look at some of the comments and I'll try to respond to each of the changes requested. Also, I've changed the status of the PR, lol. It's not a Draft again afterall

src/popper/runner_host.py
edeediong: Oh, SImilar to how it was done in the previous Docker implementation. I didn't really know this could work. Thanks I'd try this out.

src/popper/runner_host.py
edeediong: Oh, including a try and except in here would fix that. I felt there was a way to handle the failure of a workflow from previous functions in the codebase. I will try to make changes however.

src/popper/runner_host.py
```diff
@@ -293,6 +293,212 @@ def _update_with_engine_config(self, container_args):
...
+
+    def stop_running_tasks(self):
+        """Stop containers started by Popper."""
+        for c in self._spawned_containers:
+            log.info(f"Stopping container {c}")
+            _, ecode, _ = HostRunner._exec_cmd(["podman", "stop", c], logging=False)
+            if ecode != 0:
+                log.warning(f"Failed to stop the {c} container")
+
+    def _find_container(self, cid):
```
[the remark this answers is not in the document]
edeediong: Oh there was no test for the mirror implementation of this in Docker. However, I will ensure tests are written for this should I also write tests for the docker implementation of this? Also, it seems I don't get code coverage reports anymore when I make PRs. Is this right?

src/popper/runner_host.py
edeediong: Yeah I found [this](https://github.com/getpopper/popper/blob/250d69e7627808842843cb2316c33874e078b7eb/src/popper/runner_host.py#L93), I thought it would cover exceptions.

src/popper/runner_host.py
edeediong: Actually I implemented this because of https://github.com/getpopper/popper/blob/250d69e7627808842843cb2316c33874e078b7eb/src/popper/runner_host.py#L165. We will need to return the status code considering that's what was implemented in Docker. I do not know if the command blocks though, I felt that's the need of `wait`

src/popper/runner_host.py
```diff
@@ -293,6 +293,212 @@ def _update_with_engine_config(self, container_args):
...
+        cmd = ["podman", "container", "ls", "--filter", f"id={cid}", "-q"]
+        _, _, containers = HostRunner._exec_cmd(cmd, logging=False)
+        filtered_containers = [c for c in containers if c == cid]
+
+        if len(filtered_containers):
+            return filtered_containers[0]
+
+        return None
+
+    def _get_build_info(self, step):
```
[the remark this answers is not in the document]
edeediong: Oh alright... I'd make changes to this

src/test/test_runner_host.py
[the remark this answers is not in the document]
edeediong: I didn't know how to implement this,I think I posted this on Slack. This particular end-to-end workflow execution doesn't reference docker in anyway so I really did not know how to implement this for Podman. I'd push some changes regarding this for you to see if I have perfectly understood this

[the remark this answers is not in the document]
edeediong: Okay I'd remove vagrant and put podman instead.

src/popper/runner_host.py
```diff
@@ -293,6 +293,212 @@ def _update_with_engine_config(self, container_args):
...
+        log.info(f"[{step.id}] podman start")
+
+        if self._config.dry_run:
+            return 0
+
+        self._spawned_containers.add(container)
+
+        try:
+            cmd = ["podman", "start", container]
+            HostRunner._exec_cmd(cmd, logging=False)
```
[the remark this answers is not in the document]
edeediong: Would we need to store the output in a variable?

src/scripts/install_podman.sh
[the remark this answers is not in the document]
ivotron: if this package is needed by the `podman` package, then that will be done automatically by `apt`. For example:

```
root@70c7503fe555:/# apt install podman
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  catatonit conmon containernetworking-plugins containers-common containers-golang containers-image
  crun dmsetup iptables libdevmapper1.02.1 libglib2.0-0 libglib2.0-data libgpgme11 libicu60 libip4tc0
  libip6tc0 libiptc0 libmnl0 libnetfilter-conntrack3 libnfnetlink0 libxml2 libxtables12 libyajl2
  multiarch-support podman-plugins runc shared-mime-info slirp4netns uidmap varlink xdg-user-dirs
Suggested packages:
  kmod
The following NEW packages will be installed:
  catatonit conmon containernetworking-plugins containers-common containers-golang containers-image
  crun dmsetup iptables libdevmapper1.02.1 libglib2.0-0 libglib2.0-data libgpgme11 libicu60 libip4tc0
  libip6tc0 libiptc0 libmnl0 libnetfilter-conntrack3 libnfnetlink0 libxml2 libxtables12 libyajl2
  multiarch-support podman podman-plugins runc shared-mime-info slirp4netns uidmap varlink
  xdg-user-dirs
0 upgraded, 32 newly installed, 0 to remove and 6 not upgraded.
Need to get 37.3 MB of archives.
After this operation, 181 MB of additional disk space will be used.
```

see `slirp4netns` above

src/popper/runner_host.py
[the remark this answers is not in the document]
ivotron: yes, that'd be great if we can have tests for docker as well!

src/popper/runner_host.py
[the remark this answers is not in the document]
ivotron: when `logging=True`, the output is not returned. But we do want to catch the exit code of the command, so something like:

```python
_, ecode, _ = HostRunner._exec_cmd(cmd)
```

so that we can then return the exit code of the container

src/popper/runner_host.py
ivotron: see my other comment

src/popper/runner_host.py
ivotron: Failure here can be checked by looking at the value of the exit code returned by the `_exec_cmd` function. If it's non-zero, then something failed

src/test/test_runner_host.py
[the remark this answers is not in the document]
ivotron: this can be the same test as for the docker, but modifying [this line](https://github.com/getpopper/popper/blob/1d7daf2/src/test/test_runner_host.py#L274) so that the `engine_name="podman"` is passed. So copy-pasting that test function, and modifying this argument

src/scripts/install_podman.sh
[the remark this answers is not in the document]
edeediong: Is this the same as the Travis environment? Let me retry that.

src/popper/runner_host.py
edeediong: I did something like this when refactoring `find_container` method. I should've updated my comments here. Thanks

src/popper/runner_host.py
edeediong: Oh I totally understand now

src/test/test_runner_host.py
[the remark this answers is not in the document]
edeediong: Thanks for this. I've made some changes which have helped me replicate the error you had when running a workflow. Let me get on fixing that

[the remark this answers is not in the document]
edeediong: I've been able to replicate the errors you have here, let me get on to fixing it

src/scripts/install_podman.sh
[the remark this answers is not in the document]
edeediong: I tried removing `slirp4netns` in my [local](https://travis-ci.com/github/edeediong/popper/jobs/358297860) Travis and it failed. It actually requires `slirp4netns`. The above snippet is it from a Travis environment, a bare metal machine or a virtual machine?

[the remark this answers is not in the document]
edeediong: Same with [popper](https://travis-ci.com/github/getpopper/popper/jobs/358297891) travis environment. It seems to me it doesn't get installed into the VM travis provides.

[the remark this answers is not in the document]
ivotron: I tested in an ubuntu:18.04 docker image, and copy-pasted the commands, but excluded `slirp4netns`. Weird... we can leave it as you have it then, it might be something related to the travis images. Sorry for the noise

src/popper/runner_host.py
edeediong: Hi Ivo! This refactoring can't be done because `msg` is a string and a string has no append attribute. We use msg for logging purposes.

src/scripts/install_podman.sh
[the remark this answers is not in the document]
edeediong: No it's fine. That's what I was trying to say earlier, all environments didn't throw an error except Travis. So hopefully no one will encounter this issue when using on their local.

src/popper/runner_host.py
edeediong: Hi Ivo! I attempted this and I ran into an issue. `None` gets appended to the list if the argument isn't found. How do I handle such?

src/popper/runner_host.py
ivotron: yes, sorry, I meant that `msg` can be made a list and then append items to it

src/popper/runner_host.py
ivotron: oh, I see. To consider the case where the key exists but is `None` we can do:

```python
cmd.append(container_args.get("command", "") or "")
```

ivotron:
@edeediong regarding the `runs` attribute of a step, and how that’s passed to the --entrypoint flag of podman. For example, if we have this workflow:

```yaml
steps:
- uses: docker://alpine:3.12
  runs: [ls, -l]
  args: [-a]
```

then we can have the podman runner take the first element of the `runs` list (if given), and make that the value of `--entrypoint`, then we take the rest of the list and pass it as arguments to the command. So in this example, this would be something like:

```
podman run --entrypoint=ls alpine:3.12 -l -a
```

src/popper/runner_host.py
edeediong: Hmmmmm... I will try this

edeediong:
@ivotron can you help me check what's wrong. When logging I get a `FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpgfolyfi9/hello.txt'`... But I believe the container args are passed in correctly

src/popper/runner_host.py
```diff
@@ -293,6 +293,198 @@ def _update_with_engine_config(self, container_args):
...
 
 
+class PodmanRunner(StepRunner):
+    """Runs steps in podman on the local machine."""
+
+    def __init__(self, init_podman_client=True, **kw):
+        super(PodmanRunner, self).__init__(**kw)
+
+        self._spawned_containers = set()
+        _, _, self._p_info = HostRunner._exec_cmd(["podman", "info"], logging=False)
```
ivotron, line 303: this has to be moved, otherwise the `init_podman_client` would not be having an effect.

src/popper/runner_host.py
```diff
@@ -293,6 +293,198 @@ def _update_with_engine_config(self, container_args):
...
+            _, ecode, _ = HostRunner._exec_cmd(["podman", "stop", c], logging=False)
+            if ecode != 0:
+                log.warning(f"Failed to stop the {c} container")
+
+    def _find_container(self, cid):
+        """Checks whether the container exists."""
+        cmd = ["podman", "inspect", "-f", str("{{.Id}}"), cid]
+        _, ecode, containers = HostRunner._exec_cmd(cmd, logging=False)
+
+        if ecode != 0:
```
ivotron, line 363: this needs to check whether the error is the one being expected, otherwise an actual error would be swallowed

src/popper/runner_host.py
```diff
@@ -293,6 +293,198 @@ def _update_with_engine_config(self, container_args):
...
+        log.info(msg)
+
+        cmd = ["podman", "create"]
+        cmd.extend(["--name", container_args.get("name") or ""])
+        cmd.extend(["-v", container_args.get("volumes")[0] or ""])
+        if container_args.get("environment"):
+            for i, j in container_args["environment"].items():
+                cmd.extend(["-e", f"{i}={j}"])
+        cmd.append("-d" if container_args.get("detach") else "")
+        cmd.extend(["-w", container_args.get("working_dir") or ""])
```
ivotron, line 473: `working_dir` will always have a value, so this can use it's value directy without checking for existence

src/popper/runner_host.py
```diff
@@ -293,6 +293,198 @@ def _update_with_engine_config(self, container_args):
...
+        cmd.extend(["-h", container_args.get("hostname") or ""])
```
ivotron, line 474: when `hostname` is not defined, this will only pass `-h`, which will cause the command to fail

src/popper/runner_host.py
```diff
@@ -293,6 +293,198 @@ def _update_with_engine_config(self, container_args):
...
+        # cmd.extend(["-t", container_args.get("tty") or ""])
+        # cmd.extend(["--domainname", container_args.get("domainname") or ""]) #This is commented out because tty and domain name cannot be left blank
+        cmd.append(container_args.get("image") or "")
```
ivotron, line 477: `image` is always defined

src/popper/runner_host.py
```diff
@@ -293,6 +293,198 @@ def _update_with_engine_config(self, container_args):
...
+        for i in container_args["command"]:
+            cmd.extend([i or ""])
+
+        _, ecode, container = HostRunner._exec_cmd(cmd, logging=False)
+        if ecode != 0:
+            return
```
ivotron, line 483: this can be `return None` to make it explicit

edeediong: Alright thanks for this

src/popper/runner_host.py
edeediong: I corrected this! Argh! I didn't make the PR. it failed and I noticed this

edeediong: So how do you advise I pass this command using container.args?

src/popper/runner_host.py
edeediong: I've done this

src/popper/runner_host.py
edeediong: When you say error being expected what do you mean **Error: error inspecting object: no such object:** right? So how do you advise I represent this? I implemented that based on what I saw [here](https://github.com/getpopper/popper/blob/6831e0131eb48ce93461457a5b309d388595596a/src/popper/runner_host.py#L273). It's more of an implementation of DockerRunenr and there is nothing to do a check in DockerRunner

edeediong: Maybe I should log it? This is exactly what I did though

src/popper/runner_host.py
edeediong: I've done this

src/popper/runner_host.py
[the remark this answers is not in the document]
edeediong: I just saw this. Thanks

src/popper/runner_host.py
ivotron: yeah, checking that the returned output contains the `no such object` substring would be fine. This could then error out for other errors

ivotron: for the docker case, this is not a problem because we're making use of the python API, so any error there would result in an exception being thrown

src/popper/runner_host.py
ivotron: so this would be either empty or `-h <hostname>`, any way you find to express that is fine, for example:

```python
h = container_args.get("hostname", None)
if h:
  cmd.extend(f"-h {h}")
```

src/popper/runner_host.py
edeediong: Alright got it I totally understand this.

src/popper/runner_host.py
edeediong: Oh excellent, I was thinking of one-liners that's why I couldn't come up with this

src/popper/runner_host.py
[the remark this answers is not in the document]
edeediong: I've refactored this and all the tests are passing.

src/popper/runner_host.py
edeediong: I think this is the only thing remaining because everything works except the result

edeediong: I figured out a better way to handle this. The error code when the object isn't available is `125`. I'll test that and if it isn't 125, raise an exception

src/popper/runner.py
```diff
@@ -230,6 +231,107 @@ def _prepare_environment(self, step, env={}):
             )
         return step_env
 
+    def _get_build_info(self, step, class_name="DockerRunner"):
```
ivotron, line 234: this is kind of defeating the purpose of having a class hierarchy. Instead, this method should do those things that are generic to all children, and then, each child can reuse "as is", override partially (e.g. declare the same function, call the parent method and then do something else, or completely override the method).

src/popper/runner_host.py
```diff
@@ -281,24 +229,160 @@ def _find_container(self, cid):
...
+
+        if self._config.dry_run:
+            return 0
+
+        self._spawned_containers.add(container)
+
+        cmd = ["podman", "start", "-a", container]
+        _, e, output, = HostRunner._exec_cmd(cmd)
+
+        cmd1 = ["podman", "inspect", "-f", "{{.State.ExitCode}}", container]
```
ivotron, line 282: this shouldn't be needed. Just doing `podman start` then checking for exit code should be enough

edeediong: Oh yes so when I do a `podman start`, the status code returned by HostRunner is always 0. It never throws an error in this case, which is why I have to inspect the container for the exit code.

edeediong:
> thanks @edeediong! In order to simplify this PR, it'd be best to separate the refactoring from the PodmanRunner logic. This would simplify also the implementation. So we can have first only the refactoring changes, then the podmanrunner changes, then the slurm runner changes for podman.
> 
> or if you prefer, you could first only include the podmanrunner changes, and then the refactoring ones

@ivotron Can I open multiple PRs to this effect? I wouldn't want to tamper with the already working code on this branch.

src/popper/runner_host.py
```diff
@@ -232,6 +199,10 @@ def _create_container(self, cid, step):
             return
 
         container_args = self._get_container_kwargs(step, f"{img}:{tag}", cid)
+        if container_args["volumes"]:
```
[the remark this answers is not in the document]
edeediong: Alright I understand this

src/popper/runner.py
edeediong: @ivotron I know I'm going to split the PR but can you please take a look at the new changes I made as regards this? I think I have handled the hierarchical issue

src/popper/runner_host.py
```diff
@@ -438,7 +479,12 @@ def _get_container_options(self):
         return options
 
     def _create_container(self, step, cid):
-        build, image, build_ctx_path = self._get_build_info(step)
+        build, image, _, _, build_ctx_path = self._get_build_info(step)
+
+        if "shub://" in step.uses or "library://" in step.uses:
+            build = False
+            image = step.uses
+            build_ctx_path = None
```
ivotron, line 487: nice! that looks great, really clean

src/popper/runner.py
```diff
@@ -230,6 +231,96 @@ def _prepare_environment(self, step, env={}):
...
+            "detach": not self._config.pty,
+            "tty": self._config.pty,
+            "stdin_open": self._config.pty,
+        }
+
+        self._update_with_engine_config(args)
+        args.update(step.options)
+        log.debug(f"container args: {pu.prettystr(args)}\n")
+
+        return args
```
ivotron, line 322: looks good!

edeediong: Thank you :+1: :)

ivotron:
can we close this Eddie?

edeediong:
Oh sure, I sorta used it as a testing ground. Since everything seems almost ready this can be closed

PR #878, opened 2020-07-05
show proper error message when head commit not found

JayjeetAtGithub:


src/popper/scm.py
```diff
@@ -167,10 +167,16 @@ def get_sha(repo, short=None):
...
 
-    if short:
-        return repo.git.rev_parse(repo.head.object.hexsha, short=short)
+    try:
+        if short:
+            sha = repo.git.rev_parse(repo.head.object.hexsha, short=short)
+        else:
+            sha = repo.git.rev_parse(repo.head.object.hexsha)
+
+    except ValueError as e:
```
ivotron, line 176: I was thinking about this and was wondering whether it'd be more friendly to show a warning instead of failing. Letting the user know that git information is ignored because the repository doesn't have a commit. We show the message and then return `None`, what do you think?

JayjeetAtGithub: Yeah, I think that should be fine. Thanks

src/popper/scm.py
```diff
@@ -167,10 +167,17 @@ def get_sha(repo, short=None):
...
+        sha = None
+        log.warning(str(e))
```
ivotron, line 178: let's make this more friendly by showing a less git-cryptic message, something like `Could not obtain commit ID (SHA1) due to the Git repository at /path/to/repo being empty.`

ivotron:

PR #881, opened 2020-07-13
Implementation of Docker container options for steps

Jswig:
This contains changes allowing to pass step-specific container options to the Docker runner.
Edited some tests to reflect new behavior.

### DIFF
--- a/src/popper/parser.py
+++ b/src/popper/parser.py
@@ -35,7 +35,7 @@ class WorkflowParser(object):
                                 "matching-rule": "any",
                                 "mapping": {"regex;(.+)": {"type": "str"}},
                             },
-                            "container_config": {"type": "map", "allowempty": True},
+                            "options": {"type": "map", "allowempty": True},
                         },
                     }
                 ],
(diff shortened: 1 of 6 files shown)


src/popper/parser.py
```diff
@@ -35,6 +35,7 @@ class WorkflowParser(object):
                                 "matching-rule": "any",
                                 "mapping": {"regex;(.+)": {"type": "str"}},
                             },
+                            "options": {"type": "map", "allowempty": True},
```
Jswig, line 38: Preferred to keep the yml validation simple as it would be unwieldy to specify all allowable options for all container runtimes

src/popper/runner_host.py
```diff
@@ -257,7 +257,7 @@ def _get_container_kwargs(self, step, img, name):
         }
 
         self._update_with_engine_config(args)
-
+        args.update(step.options)
```
Jswig, line 260: after `update_with_engine_config` to ensure overall configuration is overriden

ivotron:
thanks @Jswig !!

Issue #785, opened 2020-02-20
cli: show docker build output


ivotron:
currently, the output of `docker build` is hidden. We can show it by default and hide it when the `--quiet` flag to `popper run` is given

WJayesh:
I would like to work on this!

ivotron:
sure, please go ahead!

nicolecastillo:
I would like to work on this one.

PR #882, opened 2020-07-13
show docker build output (#785)

nicolecastillo:
Shows docker build output by default and when the `--quiet flag` is given to `popper run` it hides it.


src/popper/runner_host.py
```diff
@@ -212,9 +212,17 @@ def _create_container(self, cid, step):
         if build:
             log.info(f"[{step.id}] docker build {img}:{tag} {build_ctx_path}")
             if not self._config.dry_run:
-                self._d.images.build(
+
+                _, build_logs = self._d.images.build(
```
ivotron, line 216: this blocks, so the output only shows until `images.build()` returns. If the output of the build is too big, this might cause memory issues. In order to avoid this, this code can use the low-level API as exemplified [here](https://stackoverflow.com/a/54715475). Very similar to what what's already here, but with a couple of tweaks.

ivotron:
thanks! I made some tweaks based on what you have in this PR and pushed them to https://github.com/getpopper/popper/tree/pr/882 . Does all that make sense to you? If it does, can you please rebase that branch from this PR?

nicolecastillo:
nice! I did the rebase and it works smoothly.

ivotron:
great, thanks! can you please push the changes to this PR, so I can merge from here? otherwise your authorship is going to get lost.

nicolecastillo:
sure :grin:

ivotron:
thanks!

Issue #863, opened 2020-06-19
example: add a CI pipeline for bazel


ivotron:
Add an example workflow in the `examples/ci/` folder that exemplifies how to make use of Popper to create a CI pipeline for projects that use [`bazel`](https://bazel.build/) as their build system. The image used by the workflow can be [this one](https://hub.docker.com/r/insready/bazel).

PR #887, opened 2020-07-17
Add Bazel example workflow

FernandaDguez:
Adds an example workflow in `examples/ci/` of a ci pipeline for bazel projects #863

### DIFF
--- a/examples/ci/bazel.yml
+++ b/examples/ci/bazel.yml
@@ -2,24 +2,24 @@ options:
   env:
     GOPATH: /workspace
 steps:
-- id: get-ws-gen
+- id: get-workspace-gen
   uses: docker://tikhoplav/go
   args: [go, get, github.com/jin/create-bazel-workspace]
-  
-- id: install
+
+- id: install-gen
   uses: docker://tikhoplav/go 
   args: [go, install, github.com/jin/create-bazel-workspace]
-  
-- id: gen-ws
+
+- id: gen-workspace
   uses: docker://tikhoplav/go
   args: [./bin/create-bazel-workspace]
-  
+
 - id: build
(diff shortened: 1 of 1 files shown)


examples/ci/bazel.yml
```diff
@@ -0,0 +1,25 @@
...
+  uses: docker://tikhoplav/go
+  args: [go, get, github.com/jin/create-bazel-workspace]
+
+- id: install-gen
+  uses: docker://tikhoplav/go 
+  args: [go, install, github.com/jin/create-bazel-workspace]
+
+- id: gen-workspace
+  uses: docker://tikhoplav/go
+  args: [./bin/create-bazel-workspace]
```
ivotron, line 15: these steps could be consolidated in a single one:

```yaml
- id: setup-workspace
  uses: docker://tikhoplav/go
  runs: [bash, -uxec]
  env:
    GOPATH: /workspace
  args:
  - |
    go get ...
    go install ..
    bin/create-bazel-workspace
```

or alternatively, a docker image for `create-bazel-workspace` could be created and uploaded to https://github.com/getpopper/library, and that image can be invoked instead of the base go image.

One question: is `create-bazel-workspace` idempotent? what happens if it is executed multiple times?

ivotron: another alternative would be to assume that the workspace already exists and only provide as example the bazel commands. This is what the other CI examples do (for example the NodeJS one assumes a packages.json file already exists).

FernandaDguez: yes, I would better assume that the BUILD and WORKSPACE files exist to preserve step idempotency

ivotron:
thanks!

PR #889, opened 2020-07-21
Kubernetes

JayjeetAtGithub:


JayjeetAtGithub:
@ivotron You can follow [these](https://github.com/getpopper/popper/blob/kubernetes/docs/sections/cn_workflows.md#running-on-a-cluster-without-shared-storage) instructions to test out the runner.

src/scripts/install_k8s.sh
```diff
@@ -0,0 +1,38 @@
...
+  storageClassName: manual
+  capacity:
+    storage: 1Gi
+  accessModes:
+    - ReadWriteMany
+  hostPath:
+    path: "/tmp"
+EOF
+
+kubectl apply -f pv.yaml
```
ivotron, line 36: this could be done by popper, or do you see anything that wouldn't work if it did?

docs/sections/cn_workflows.md
```diff
@@ -414,6 +414,68 @@ either through the `--resource-manager/-r` option or through the config file.
...
+    path: "/tmp"
+EOF
+```
+
+2. Then create the persistent volume.
+```bash
+$ kubectl apply -f pv.yaml
+```
+
+3. Write a configuration file similar to this.
```
ivotron, line 458: ideally we wouldn't need to add anything, and everything would be available in the kubctl config file.  So running in k8s would only require getting a k8s kubctl config file from an admin, and just run

```
popper run -r kubernetes
```

I think we have all we need, assuming a workflow is not building images, right?

src/scripts/install_k8s.sh
JayjeetAtGithub: Users can have a PV created from before that they would like to use which might contain some data that the workflow might be using.
This is particularly important in a shared cluster where each user might have been allocated some PV by the admin. 
Also, since everything created by popper is ephemeral but PVs are not, i felt lets keep PV and popper separate.

What we can do is we can always ask a PV from user (considering it will be a complex PV like aws ebs, etc.) through config and if not provided popper will create a local PV  and select a node randomly (if some node is not schedulable , popper can try out other nodes) and schedule all pods on that node. What do you think?

docs/sections/cn_workflows.md
JayjeetAtGithub: Yes, registry user need not be passed if our workflow doesnt build images.

JayjeetAtGithub: Also, as mentioned in the previous comment, if we can make local pv a default case, rather than an exception (which it is currently), i believe we will not require a config file.

src/scripts/install_k8s.sh
ivotron: > What we can do is we can always ask a PV from user (considering it will be a complex PV like aws ebs, etc.) through config and if not provided popper will create a local PV and select a node randomly (if some node is not schedulable , popper can try out other nodes) and schedule all pods on that node. What do you think? Basically I am thinking if we can make local PV the default case instead of an exception.

this sounds good!

docs/sections/cn_workflows.md
ivotron: sounds good! Once #907 is merged, that would also help to not have to specify the registry here as well. The k8s workflows I am thinking are:
 1. no images built. All is needed is the kubectl config file (that includes the pv config).
 2. images built. Same as above, but before running a workflow users do `docker login` first, then specify images being built via the `build` attribute (as proposed in #907), and then do `docker run`. This also keeps the config out of the picture

src/scripts/install_k8s.sh
JayjeetAtGithub: Ok. I would update the code accordingly

docs/sections/cn_workflows.md
JayjeetAtGithub: > images built. Same as above, but before running a workflow users do docker login first, then specify images being built via the build attribute (as proposed in #907), and then do docker run. This also keeps the config out of the picture

Yeah, correct. this should not require any config. thanks !

JayjeetAtGithub:
@ivotron I have made the changes. Now if you have a workflow that doesn't build an image , you can simply do 
```
popper run -f wf.yml -r kubernetes
```

Issue #890, opened 2020-07-22
cli: Running some example workflows throws an error


edeediong:
When I try to run the [go.yml](https://github.com/getpopper/popper/blob/master/examples/ci/go.yml) workflow on both Docker and Podman. I get the following error

```ERROR: Schema validation failed:
 - Key 'name' was not defined. Path: '/steps/1'.
 - Key 'name' was not defined. Path: '/steps/2'.
```

Also when running [ansible.yml](https://github.com/getpopper/popper/blob/master/examples/cd/ansible.yml) workflow, I get 
```ERROR: Schema validation failed:
 - Value 'False' is not of type 'str'. Path: '/steps/0/env/ANSIBLE_HOST_KEY_CHECKING'.
```

PR #891, opened 2020-07-22
fix: corrected syntax of example workflows

edeediong:
This fix is as regards the issue opened at #890


edeediong:
@ivotron

ivotron:
thanks @edeediong !

PR #894, opened 2020-07-24
Cosmetic fixes to install.sh

wtraylor:
- Add -e flag to echo so \n is interpreted
- Quote $PWD so that directories with spaces work

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -18,15 +18,15 @@ printenv > /tmp/.envfile
 docker run --rm -ti \
   --volume /tmp:/tmp \
   --volume /var/run/docker.sock:/var/run/docker.sock \
-  --volume $PWD:$PWD \
-  --workdir $PWD \
+  --volume "$PWD":"$PWD" \
+  --workdir "$PWD" \
   --env-file /tmp/.envfile \
   getpopper/popper:v2.7.0 $@
 EOF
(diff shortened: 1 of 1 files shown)


ivotron:
thanks for working on this @wtraylor! I tested on a couple of machines and this looks like the following:

```
-e
Installed version v2.7.0 to executable file '/home/ivotron/tmp/popper'

Do you wish to move this binary to /usr/local/bin/? [Y/n] n
-e
To make the popper command globally available, add it
-e to a folder reachable by the PATH variable.
```

so it looks like `-e` is not part of the [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html). So one alternative would be to add empty `echo` statements instead of new lines. The other would be to use [`printf`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html) instead. Either of the two would work.

wtraylor:
Thank you @ivotron for checking POSIX compliance.
I have replaced `\n` with empty `echo`s.

During testing I found some edge cases that should be handled: If `sudo` is not installed on the system, and if there’s no write permission to create the `./popper` file.

ivotron:
thanks a lot @wtraylor! 🙏

PR #897, opened 2020-07-28
Allow spaces in step IDs

wtraylor:
The call to `docker` must receive the command-line arguments verbatim,
including quotes. Therefore `$@` needs to be quoted.

Without quotes (before):
`popper run -f wf.yml "first step"`
would execute:
`docker run <...> getpopper/popper:v2.7.0 run -f wf.yml first step`

With quotes (after fix):
`docker run <...>  getpopper/popper:v2.7.0 run -f wf.yml "first step"`


ivotron:
thanks a lot @wtraylor!

Issue #898, opened 2020-07-28
Singularity not working through dockerized `popper`


wtraylor:
# Steps to reproduce
```bash
# Create a dummy project
mkdir project
cd project
# Download the popper script
curl -sSfL https://raw.githubusercontent.com/getpopper/popper/master/install.sh | sh
chmod +x popper
./popper scaffold
# Try to run workflow with singularity
./popper run -f wf.yml --engine singularity
```
Error message:
```
[popper] Cloning step repositories
[popper] - https://github.com/popperized/bin@master
[1] singularity build popper_1_1970273b.sif /home/living/.cache/popper/1970273b/github.com/popperized/bin/sh
ERROR: [Errno 2] No such file or directory: 'sudo': 'sudo'
```
Version: current `master`, commit 227bfee

# The Problem
Singularity and `sudo` are not installed in the Docker container that is running `popper`.

# Tried Solutions
Install Singularity in Popper’s Alpine-based Docker image (`scr/Dockerfile`):
```dockerfile
FROM python:3.7.7-alpine3.12

COPY . /popper

RUN apk --no-cache add git && \
    pip install --no-cache-dir /popper && \
    rm -r /popper

# NEW: Install Singularity from the testing repository.
RUN echo http://dl-cdn.alpinelinux.org/alpine/edge/testing >> /etc/apk/repositories && \
    apk update && \
    apk add sudo singularity

ENTRYPOINT ["popper"]
```

However, with that installed I get an error about `mksquashfs`:
```
$ popper run -f wf.yml -e singularity
[popper] Cloning step repositories
[popper] - https://github.com/popperized/bin@master
[1] singularity build popper_1_1970273b.sif /home/living/.cache/popper/1970273b/github.com/popperized/bin/sh
WARNING: 'nodev' mount option set on /tmp, it could be a source of failure during build process
FATAL:   Unable to create build: while searching for mksquashfs: exec: "mksquashfs": executable file not found in $PATH
WARNING: 'nodev' mount option set on /tmp, it could be a source of failure during build process
FATAL:   Unable to create build: while searching for mksquashfs: exec: "mksquashfs": executable file not found in $PATH
[1] singularity run popper_1_1970273b.sif ['ls']
[...]

PR #899, opened 2020-07-28
Add error message that Singularity needs PIP

wtraylor:
It says it in the `README.md` that Singularity is not supported by
Popper installed through `install.sh`, but not every user has read it.

This fixes part of #898.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -37,6 +37,13 @@ command -v docker >/dev/null 2>&1 || { echo >&2 "docker command not found. Abort
 cat > ./popper << "EOF"
 #!/usr/bin/env sh
 
+if echo "$@" | grep '\e\s*singularity\|--engine\s*singularity'; then
+  echo >&2 'Singularity is not supported.',
+  echo >&2 'Please install Popper as a Python Package. Instructions can be found here:'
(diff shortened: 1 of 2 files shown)


ivotron:
thanks @wtraylor, that's a great catch! One thing we can do, in order to make this more generic, is to identify this within popper itself. We can check whether we're running inside docker when `--engine=singularity` is given (as shown here https://stackoverflow.com/a/20012536), for example as part of the constructor of the [Singularity runner](https://github.com/getpopper/popper/blob/master/src/popper/runner_host.py#L309). what do you think?

wtraylor:
That’s a good idea to move the message to `SingularityRunner`.
I implemented it in a new commit.

I tested it like this:
```sh
docker run -it --rm -v $PWD:$PWD -w $PWD --network host --entrypoint sh python:3.7.7-alpine3.12
# Within the container:
apk add --no-cache git
apk pip install --no-cache-dir popper/src
cd /tmp
popper scaffold
popper run -e singularity -f wf.yml
# Error message comes here as expected.
popper run -f wf.yml
# No error message.
```

ivotron:
thanks a lot @wtraylor ! This is awesome. And thanks a lot for taking the time to test it on your end!

PR #903, opened 2020-07-31
feat: Added PodmanRunner to the codebase

edeediong:


edeediong:
@ivotron can you please review this?

src/test/test_runner_host.py
```diff
@@ -197,6 +197,147 @@ def test_docker_basic_run(self):
...
+            cmd = ["podman", "container", "rm", c]
+            HostRunner._exec_cmd(cmd, logging=False)
+
+    @unittest.skipIf(os.environ.get("ENGINE", "docker") != "podman", "ENGINE != podman")
+    def test_podman_basic_run(self):
+        repo = self.mk_repo()
+        conf = ConfigLoader.load(engine_name="podman", workspace_dir=repo.working_dir)
+
+        with WorkflowRunner(conf) as r:
+            # wf_data = {"steps": [{"uses": "popperized/bin/sh@master", "args": ["ls"],}]}
```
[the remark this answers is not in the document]
edeediong: Yes it is.

src/popper/runner_host.py
```diff
@@ -302,13 +439,6 @@ def _get_recipe_file(build_ctx_path, cid):
         else:
             log.fail("No Dockerfile was found.")
 
-    @staticmethod
-    def _in_docker():
```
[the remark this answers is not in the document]
edeediong: Ouch, I'll correct it

src/test/test_runner_host.py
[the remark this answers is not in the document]
ivotron: hmm, it'd be great if we could to find the root cause, as this shouldn't fail.

Also, it'd be good to include a workflow that fails so that we ensure that the runners return non-zero exit code. This would catch the issue with 2.0.3 we faced.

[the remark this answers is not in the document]
edeediong: I think I've found the issue. Let me give a detailed explanation.

```['podman', 'create', '--name', 'popper_1_483f7d07', '-v', '/tmp/tmpn9poi6r_:/workspace:Z', '-e', 'GIT_COMMIT=f81810d4e55a4586d7143760722a522f51e08570', '-e', 'GIT_BRANCH=master', '-e', 'GIT_SHA_SHORT=f81810d', '-e', 'GIT_REMOTE_ORIGIN_URL=https://github.com/my/repo', '-d', '-w', '/workspace', 'popperized/bin:master', 'ls']
```
When the above command runs, Podman tries to pull the image `popperized/bin:master` from the docker repository as evidenced by the results shown below.

```', 'to', 'retrieve', 'auth', 'token:', 'invalid', 'username/password:', 'unauthorized:', 'Please', 'login', 'to', 'the', 'Red', 'Hat', 'Registry', 'using', 'your', 'Customer', 'Portal', 'credentials.', 'Further', 'instructions', 'can', 'be', 'found', 'here:', 'https://access.redhat.com/RegistryAuthentication', 'Trying', 'to', 'pull', 'docker.io/popperized/bin:master...', 'Getting', 'image', 'source', 'signatures', 'Copying', 'blob', 'sha256:ed42820a5a6a336e9ba51dd866d035f9500fdc4fbb07acd2bc40283820769778', 'Copying', 'blob', 'sha256:55e542d1686ed5e6a39f48ca2ee55c47d9cc1513a97f8783dad60e5c7bfbd8ae', 'Copying', 'blob', 'sha256:412bf51e209fa1182aa4d0b61fb144a9dc8ab885b614fd29327773c6baa4a39d', 'Copying', 'config', 'sha256:036ccbea75797737bd1824eeb8372509f75b2575a7b96245c1fea7503881a874', 'Writing', 'manifest', 'to', 'image', 'destination', 'Storing', 'signatures', 'e63945985d63231b26b307054c40cd9177275023226f89876718270b22e80ba0']
```

The [line](https://github.com/edeediong/popper/blob/ddc4e2e5538b49d8b6becca26d499c8874264c9c/src/popper/runner_host.py#L379) here tells the runner to pick the first value which in this case is `Trying/n` and not the container id. Seems when pulling from the `popperized/bin:master`, there's a delay in the VM. 

However, I found a workaround: `container[0]` will be changed to `container[-1]` to accommodate this issue.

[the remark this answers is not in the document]
edeediong: I've written a workflow which [fails](https://travis-ci.com/github/getpopper/popper/jobs/367912849). I just have to specify the version of Podman in the Travis script.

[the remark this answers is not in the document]
edeediong: Update: There seems to be no way to specify version of podman to install on travis because it doesn't exist in the openSuse [repository](https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/testing/xUbuntu_18.04/). The test presently failing @ivotron is the test that returns non-zero code. It works on v1.x but not on 2.x at least not until the code fixes are pushed to upstream

edeediong:
@ivotron I've commented out the test

src/popper/runner_host.py
```diff
@@ -229,6 +228,155 @@ def _find_container(self, cid):
...
+        for i in container_args["command"]:
+            cmd.extend([i or ""])
+
+        _, ecode, container = HostRunner._exec_cmd(cmd, logging=False)
+
+        if ecode != 0:
+            return None
+
+        container = container.rsplit()
+        return container[-1]
```
ivotron, line 377: this might fail if `rsplit()` returns an empty list. To avoid that scenario, the code can check the size of the list and add logic for handling the case of an empty list

src/popper/runner_host.py
```diff
@@ -229,6 +228,155 @@ def _find_container(self, cid):
...
+        if self._config.dry_run:
+            return 0
+
+        self._spawned_containers.add(container)
+
+        cmd = ["podman", "start", "-a", container]
+        _, e, output = HostRunner._exec_cmd(cmd, logging=False)
+
+        if e != 0:
+            log.fail(f"{output}")
```
ivotron, line 281: failure is handled by the `StepRunner`, so this can return `e` directly without checking its value

src/popper/runner_host.py
```diff
@@ -229,6 +228,155 @@ def _find_container(self, cid):
...
+
+        log.info(f"[{step.id}] podman start")
+
+        if self._config.dry_run:
+            return 0
+
+        self._spawned_containers.add(container)
+
+        cmd = ["podman", "start", "-a", container]
+        _, e, output = HostRunner._exec_cmd(cmd, logging=False)
```
ivotron, line 278: this is swallowing the step output. Instead, the code should allow logging.

Since this is not causing any failure on the unit tests, we can open an issue to add new tests for this and other runners, as a way of ensuring that when a step prints information to stdout, the loggers are indeed capturing this output. Would it be possible for you to open an issue for this?

edeediong: I added a print statement to show the step output. I did this because I needed a way to log the output when the start command gets executed and stream it to the terminal whether there is a failure or not. So it's sorta swallowing the output to release it later

I can open the issue, I'd need ideas on how to fix it though

src/popper/runner_host.py
```diff
@@ -374,6 +374,9 @@ def _create_container(self, cid, step):
             return None
 
         container = container.rsplit()
+        if len(container) < 1:
```
ivotron, line 377: should this be `len(container) == 0`? Doing `container[-1]` with a 1-size list works fine

src/popper/runner_host.py
```diff
@@ -275,10 +275,10 @@ def run(self, step):
         self._spawned_containers.add(container)
 
         cmd = ["podman", "start", "-a", container]
-        _, e, output = HostRunner._exec_cmd(cmd, logging=False)
+        _, e, _ = HostRunner._exec_cmd(cmd)
 
         if e != 0:
-            log.fail(f"{output}")
+            log.fail(f"{e}")
```
ivotron, line 281: So there's no need to fail here, this is already handled by the StepRunner ([this function](https://github.com/getpopper/popper/blob/master/src/popper/runner.py#L139-L164)). Instead, this can return `e` directly without checking its value, as there is nothing especial that needs to be done when a failure is encountered

src/popper/runner_host.py
ivotron: Sorry, I didn't understand what you mean. Can you please expand a bit on what you mean by "added print statement to show output"? Where do you have to explicitly print? That shouldn't be needed as output to stdout is generated by the `_exec_cmd` method via the logging framework. Removing `logging=False` or explicitly doing `logging=True` have that effect.

The new issue I mentioned (we'd open a new one) is related to adding tests that ensure that steps are printing to stdout when they are expected to (and test the opposite case as well). If we had that test already, it would have catched the above (where `logging=False` was given, and thus would have triggered a failure).

src/popper/runner_host.py
edeediong: Oh oh true! Thanks!!

src/popper/runner_host.py
```diff
@@ -280,8 +280,6 @@ def run(self, step):
         if e != 0:
             log.fail(f"{output}")
 
-        print(output)
```
edeediong: @ivotron This is where I added a print statement to print the results after swallowing the outputs. I now understand the importance of having a test to check stdout is actually being captured by the logger.

src/popper/runner_host.py
```diff
@@ -229,6 +228,155 @@ def _find_container(self, cid):
...
+
+    def _create_container(self, cid, step):
+        build, _, img, tag, build_ctx_path = self._get_build_info(step)
+
+        if build:
+            log.info(f"[{step.id}] podman build {img}:{tag} {build_ctx_path}")
+            if not self._config.dry_run:
+                pass
+        elif not self._config.skip_pull and not step.skip_pull:
+            log.info(f"[{step.id}] podman build {img}:{tag}")
```
ivotron, line 311: this should be `podman pull`

src/popper/runner_host.py
```diff
@@ -229,6 +228,155 @@ def _find_container(self, cid):
...
+
+        return containers.strip()
+
+    def _create_container(self, cid, step):
+        build, _, img, tag, build_ctx_path = self._get_build_info(step)
+
+        if build:
+            log.info(f"[{step.id}] podman build {img}:{tag} {build_ctx_path}")
+            if not self._config.dry_run:
+                pass
```
ivotron, line 309: a `podman build` should be executed here instead of `pass`.

ivotron: the above also means that we are missing a test that ensures that a Dockerfile from a public github repository is properly build as part of the execution of a workflow. Maybe we can add this to the issue we mentioned earlier, the one regarding missing tests

src/popper/runner_host.py
edeediong: Oh that's true since I'm doing a podman pull below, thanks

src/popper/runner_host.py
edeediong: I just looked at DockerRunner, I meant to put a `TODO:` instead of a pass. I've opened the issue and I'm gonna push changes to this effect.

PR #909, opened 2020-08-10
Podman slurm

edeediong:
Support for Slurm Runner with Podman


edeediong:
@ivotron Can you take a quick look at this?

ivotron:
thanks Eddie! 🙏

PR #912, opened 2020-08-14
slurm runner refactored

JayjeetAtGithub:
# Goals

* To simplify the SLURM runner, use `sbatch` only for MPI jobs and `srun` otherwise
* To minimize and simplify the configuration needed to pass
* To support MPI by default

# Thoughts
* Docker and Podman are not designed for HPC environments, So not supported.


JayjeetAtGithub:
@ivotron You can review this PR now. Updating the documentation is left.

PR #918, opened 2020-08-28
update k8s runner docs

JayjeetAtGithub:
Updated the documentation for the Kubernetes runner to make the runner ready for release.

### DIFF
--- a/docs/sections/cn_workflows.md
+++ b/docs/sections/cn_workflows.md
@@ -417,12 +417,12 @@ by default.
 ### Kubernetes
 
 Popper enables leveraging the compute and storage capabilities of the cloud by allowing running workflows on 
-Kubernetes clusters. One just need to have access to the cluster config file.
+Kubernetes clusters. One just need to have access to the cluster config file in order to run workflows on Kubernetes.
 Popper provisions all the required resources and takes care of the workflow execution from there.
 
 When a workflow is executed, Popper first creates a persistent volume claim and spawns an init pod and uses it to copy the workflow context (compressed in the form of a `.tar.gz` file) into the persistent volume. Then, popper teardowns this init pod and execute the steps of a workflow in separate pods of their own. After the execution of each step, the respective pods are deleted but the persistent volume claim is not deleted, so that it can reused by subsequent workflow executions.
 
-For running workflows on Kubernetes, some configuration options may need to be passed to the kubernetes resource manager through the popper configuration file. All the available configuration options is described below:
+For running workflows on Kubernetes, some configuration options can be passed to the kubernetes resource manager through the popper configuration file to customize the execution environment. All the available configuration options is described below:
 
 * `namespace`: The namespace to use to provision resources like PVCs and Pods for a workflow execution. If not provided the `default` namespace will be used.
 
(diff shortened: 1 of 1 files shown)


docs/sections/cn_workflows.md
```diff
@@ -417,64 +417,49 @@ by default.
...
 
-For running workflows on Kubernetes, some configuration options need to be passed to the kubernetes resource manager through the popper configuration file.
+For running workflows on kubernetes, several configuration options can be passed to the kubernetes resource manager through the popper configuration file to customize the execution environment. All the available configuration options has been described below.
 
-#### Running on a cluster without shared storage
+* `namespace`: The namespace within which to provision resources like PVCs and Pods for a workflow execution. If not provided the       `default` namespace will be used.
 
-**NOTE:** If your workflow needs to build an image from a `Dockerfile`, make sure you are logged in to dockerhub.
+* `persistent_volume_name`: Any pre-provisioned persistent volume like an NFS or EBS volume can be supplied through this option. Popper will then claim storage space from the supplied persistent volume. In the default case, a local persistent volume of 1Gi will be created by Popper automatically.
```
ivotron, line 430: 1GB is more common than 1Gi

ivotron: is there a default name ?

docs/sections/cn_workflows.md
```diff
@@ -417,64 +417,49 @@ by default.
...
 
-1. Write a persistent volume defination similar to the one shown below.
+* `volume_size`: The amount of storage space to claim from a persistent volume for use by a workflow. Default is 500Mi.
```
ivotron, line 432: 500MB

docs/sections/cn_workflows.md
```diff
@@ -417,64 +417,49 @@ by default.
...
-  storageClassName: manual
-  capacity:
-    storage: 1Gi
-  accessModes:
-    - ReadWriteMany
-  hostPath:
-    path: "/tmp"
-EOF
-```
+* `pod_host_node`: The node on which to restrict the deployment of all the pods. This options is important when you are using a local persistent volume. In this case, users need to restrict all the pods to a particular node. If this option is not provided, Popper will leave the task of scheduling the pods upon kubernetes. The exception to this is, when both the `pod_host_node` and `persistent_volume_name` options are not provided, Popper will try to find out a schedulable pod and schedule all the pods (init-pods + step-pods) on that node in order to use the local persistent volume of 1Gi which will be automatically created.
```
ivotron, line 434: this has a statement in active voice ("you are using"), whereas the rest of the text is using the passive voice ("users need to restrict").

docs/sections/cn_workflows.md
```diff
@@ -417,64 +417,49 @@ by default.
...
+
+* `registry`: The registry to which to push images after building them on the host machine. The default image registry is Docker hub.
```
ivotron, line 436: Docker Hub

docs/sections/cn_workflows.md
```diff
@@ -417,64 +417,49 @@ by default.
...
-$ cat<< EOF > config.yml
-resource_manager: 
+$ cat << EOF > config.yml
+resource_manager:
   name: kubernetes
   options:
-    node_selector_host_name: mynode
-    persistent_volume_name: myvol
-    registry_user: myuser
+    registry_user: myusername
```
ivotron, line 458: this can have a preceding line describing what the command does

docs/sections/cn_workflows.md
JayjeetAtGithub: So, when volume name is not provided a volume with a name similar to `pv-hostpath-popper-a3rf5fy` is created automatically.

JayjeetAtGithub: Should we mention the default name ?

Issue #871, opened 2020-06-27
cli: add support for running on remote docker hosts


ivotron:
Popper can run workflows on remote docker hosts by use of docker-py's support for the `DOCKER_HOST` variable. We added a commit for this but had to revert it in https://github.com/getpopper/popper/commit/07f911bb596e416229f7b9d2dadc96719fa77575 since the [`src/Dockerfile`](./src/Dockerfile) doesn't build correctly when `paramiko` is added as a dependency. This `Dockerfile` is executed in the [`.popper.yml`](./.popper.yml) pipeline that uploads images to dockerhub (which in turn gets invoked from [`.travis.yml`](./.travis.yml)).
 
For this issue we'll track the changes needed to be done to the `Dockerfile` so that paramiko installs properly.

CarterB-UT:
Can me and a group of other UT students tentatively claim this issue for now? We've also asked after some other tasks.

ivotron:
yes, please go ahead! Let us know if you have any question

rishiUT:
We have successfully made changes to the dockerfile such that the install runs successfully, even when paramiko is added as a dependency. Is there anything else we can do to add support for remote docker hosts, or is the paramiko install sufficient?

ivotron:
cool, that's great! That would be sufficient for this issue

PR #935, opened 2020-11-25
Edits to Dockerfile to enable paramiko to install

rishiUT:
The edit adds imports to the Dockerfile that allow paramiko to be installed as a dependency without causing build errors.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -6,5 +6,4 @@ RUN apk --no-cache add git py3-pip py3-paramiko && \
     pip install --no-cache-dir /popper && \
     rm -r /popper
 
-
 ENTRYPOINT ["popper"]
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -4,7 +4,7 @@ COPY . /popper
 
 RUN apk --no-cache add git && \
     apk --no-cache add gcc && \
-    apk --no-cache add py-configobj linux-headers musl-dev libffi-dev && \
+    apk --no-cache add make linux-headers musl-dev python3-dev libffi-dev openssl-dev && \
     pip install --no-cache-dir /popper && \
     rm -r /popper
 
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -1,11 +1,10 @@
-FROM python:3.7.7-alpine3.12
+FROM alpine:20200917
 
 COPY . /popper
 
-RUN apk --no-cache add git && \
-    apk --no-cache add gcc && \
-    apk --no-cache add make linux-headers musl-dev python3-dev libffi-dev openssl-dev && \
+RUN apk --no-cache add git py3-pip py3-paramiko && \
     pip install --no-cache-dir /popper && \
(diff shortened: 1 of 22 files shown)


rishiUT:
This should solve issue #871

src/Dockerfile
```diff
@@ -3,6 +3,13 @@ FROM python:3.7.7-alpine3.12
 COPY . /popper
 
 RUN apk --no-cache add git && \
+    apk --no-cache add gcc && \
+    apk --no-cache add make && \
+    apk --no-cache add linux-headers && \
+    apk --no-cache add musl-dev && \
+    apk --no-cache add python3-dev && \
+    apk --no-cache add libffi-dev && \
+    apk --no-cache add openssl-dev && \
```
ivotron, line 12: this will make the image size significantly bigger. So one alternative is to remove all the build-time dependencies, but that might complicate things too much. The other option is to install the pre-packaged version of paramiko available in the edge branch of alpine, so something like this should work:

```dockerfile
FROM alpine:20200917

COPY . /popper

RUN apk --no-cache add git py3-pip py3-paramiko
    pip install --no-cache-dir /popper && \
    rm -r /popper
```

ivotron (changes requested):
thanks a lot @rishiUT, looking great! please take a look at the comments. Also, seems like some docstring changes were inserted by an IDE (or something alike). Would it be possible to have those in a separate PR? thanks!

ivotron:
oh, also, would it be possible to add a section in the docs about how one would run a workflow on a remote host? As a subsection of the "Docker" section would be a good place to write this: https://github.com/getpopper/popper/blob/master/docs/sections/cn_workflows.md#docker

src/Dockerfile
rishiUT: Hi @ivotron! We will definitely look further into running with a pre-packaged version of paramiko. I'm not certain how to remove build-time dependencies from a pre-existing library, but I can research that as well if the pre-packaged version doesn't work out. 

Regarding the docstring changes, we will find and remove those as well!

rishiUT:
Hello @ivotron! We reverted the changes the IDE made to the docstrings, implemented your suggested fix to the Dockerfile, and re-added the section in the docs regarding running a workflow on a remote host from the commit reverted in issue #871. Is there anything else we can do to repair this issue?

src/setup.py
```diff
@@ -22,6 +22,7 @@
         "docker==4.3.1",
         "dockerpty==0.4.1",
         "GitPython==3.1.7",
+        "paramiko==2.7.1",
```
ivotron, line 25: could this be bumped up to 2.7.2?

ivotron (commented):
looking great! just left one quick question!

src/setup.py
rishiUT: Done!

ivotron:
sounds good, I've merged it! Thanks a lot for your contribution 🎉

PR #936, opened 2020-11-28
Fix #926: Create the workflow context tar in the popper cache dir

### LINKED ISSUE #926: cli: k8s: create the workflow context tar file in a temp dir
Currently, the [`ctx.tar.gz`](https://github.com/getpopper/popper/blob/01eadd5e54bd686291cef20bcb4e2fd3b493829f/src/popper/runner_kubernetes.py#L137), which is the tar file that contains the workflow context, is generated inside the workspace on the local machine. This leads to the same tarball to getting zipped recursively upon executing the workflow again. Instead of creating the tarball there in the workspace dir, create it in the [`POPPER_CACHE_DIR`](https://github.com/getpopper/popper/blob/master/src/popper/config.py#L33).   Also, we can append the [workflow id](https://github.com/getpopper/popper/blob/master/src/popper/config.py#L64) to the tarball name like `ctx_12345.tar.gz` so that we can identify and reuse it when the same workflow is run again.

vivekr123:
We added the wid to the generated tar file's name and used the cache_dir wherever possible.

### DIFF
--- a/src/popper/runner_kubernetes.py
+++ b/src/popper/runner_kubernetes.py
@@ -132,10 +132,17 @@ class KubernetesRunner(StepRunner):
         """Tar up the workspace context and copy the tar file into
         the PersistentVolume in the Pod.
         """
+        source_dir = os.path.join(self._config.cache_dir, "kubernetes")
+        if not os.path.exists(source_dir):
+            os.makedirs(source_dir)
+
+        target_file = "ctx_" + str(self._config.wid) + ".tar.gz"
+        source_file = os.path.join(source_dir, target_file)
+        destination_file = "/workspace/" + target_file
+
         files = os.listdir(self._config.workspace_dir)
-        with tarfile.open(
-            os.path.join(self._config.workspace_dir, "ctx.tar.gz"), mode="w:gz"
-        ) as archive:
+
+        with tarfile.open(source_file, mode="w:gz") as archive:
             for f in files:
                 archive.add(f)
 
@@ -151,9 +158,6 @@ class KubernetesRunner(StepRunner):
             tty=False,
             _preload_content=False,
         )
-
-        source_file = os.path.join(self._config.workspace_dir, "ctx.tar.gz")
-        destination_file = "/workspace/ctx.tar.gz"
 
         with open(source_file, "rb") as image_file:
             encoded_string = base64.b64encode(image_file.read())
@@ -177,7 +181,7 @@ class KubernetesRunner(StepRunner):
         response.close()
 
         # extract the archive inside the pod
-        exec_command = ["tar", "-zxvf", "/workspace/ctx.tar.gz"]
+        exec_command = ["tar", "-zxvf", destination_file]
         response = stream(
             self._kclient.connect_get_namespaced_pod_exec,
             self._init_pod_name,


src/popper/runner_kubernetes.py
```diff
@@ -152,8 +155,8 @@ def _copy_ctx(self):
             _preload_content=False,
         )
 
-        source_file = os.path.join(self._config.workspace_dir, "ctx.tar.gz")
-        destination_file = "/workspace/ctx.tar.gz"
+        source_file = os.path.join(self._config.cache_dir, target_file)
```
JayjeetAtGithub, line 158: Can we have the source file at `$POPPER_CACHE_DIR/kubernetes/ctx_{wid}.tar.gz` ? It would help keep the cache directory more organized.

src/popper/runner_kubernetes.py
```diff
@@ -133,8 +133,11 @@ def _copy_ctx(self):
         the PersistentVolume in the Pod.
         """
         files = os.listdir(self._config.workspace_dir)
+
+        target_file = "ctx_" + str(self._config.wid) + ".tar.gz"
```
JayjeetAtGithub, line 137: Nice

src/popper/runner_kubernetes.py
JayjeetAtGithub: Can we bring the `source_file` and `destination_file` variable declarations at the top to just after the `target_file` declaration? In that way, we can prevent joining multiple times.

JayjeetAtGithub (changes requested):
Thanks a lot for working on this @vivekr123 . The changes work great. Can you also please squash the commits into 1?

src/popper/runner_kubernetes.py
JayjeetAtGithub: We can also bring the `source_file` and `destination_file` declarations to the top just after the `target_file` variable is declared. In that way, we can prevent joining paths multiple times.

vivekr123: Got it! Will do that

src/popper/runner_kubernetes.py
```diff
@@ -133,9 +133,12 @@ def _copy_ctx(self):
         the PersistentVolume in the Pod.
         """
         files = os.listdir(self._config.workspace_dir)
-        with tarfile.open(
-            os.path.join(self._config.workspace_dir, "ctx.tar.gz"), mode="w:gz"
-        ) as archive:
+
+        target_file = "ctx_" + str(self._config.wid) + ".tar.gz"
+        source_file = os.path.join(self._config.cache_dir, "kubernetes", target_file)
```
JayjeetAtGithub, line 138: When we specify the source as `$POPPER_CACHE_DIR/kubernetes/ctx_abc.tar.gz`, the "kubernetes" directory is not present and the workflow errors out. What we can do is add 2 lines of code to check if the
`$POPPER_CACHE_DIR/kubernetes/` is present or not , and if not present create it.  This can be added at the very top of the `_copy_ctx` function.

JayjeetAtGithub (changes requested):
Thanks a lot @vivekr123 . Just a single fix and we will be good to merge this.

src/popper/runner_kubernetes.py
vivekr123: I just committed with those changes. Let me know what you think!

JayjeetAtGithub (approved):
Thanks a lot. It works great. Let's merge.

Issue #928, opened 2020-11-19
cli: k8s: allow users to set options for creating host path volumes


JayjeetAtGithub:
Currently, the [default hostPath volume is created](https://github.com/getpopper/popper/blob/01eadd5e54bd686291cef20bcb4e2fd3b493829f/src/popper/runner_kubernetes.py#L263) at `/tmp` and is of `1Gi` size. We can allow the user to set these parameters according to their needs through the config file with options like `hostpathvol-path` and `hostpathvol-size`.

rishiUT:
Hi! Is it alright if I work on this issue with a group of UT students?

JayjeetAtGithub:
Hi @rishiUT , Thanks a lot for looking into this issue. Please go ahead. Let us know if there are any issues setting up the development environment.  You can begin by taking a look at the  [installation guide](https://github.com/getpopper/popper/blob/master/docs/installation.md) and the [contribution guidelines](https://github.com/getpopper/popper/blob/master/CONTRIBUTING.md).

JayjeetAtGithub:
Fixed in #937

PR #937, opened 2020-11-28
Issue #928: Suggesting a fix for non-default hostPath volumes

vivekr123:
Allows users to use options to specify non-default path and volume size in the config file. Added these new options to the documentation (cn_workflows.md).

### DIFF
--- a/docs/sections/cn_workflows.md
+++ b/docs/sections/cn_workflows.md
@@ -440,6 +440,10 @@ All the available configuration options have been described below:
   If this option is not provided, Popper will leave the task of scheduling the pods upon Kubernetes. 
   The exception to this is, when both the `pod_host_node` and `persistent_volume_name` options are not provided, Popper will try to find out a pod and schedule all the pods (init-pods + step-pods) on that node to use the `HostPath` persistent volume of 1GB which will be automatically created.
 
+* `hostpathvol_path`: The path to use for creating a HostPath volume. If not provided, /tmp will be used.
+
+* `hostpathvol_size`: The size of the HostPath volume. If not provided, 1GB will be used.
+
 To run workflows on Kubernetes:
 
 ```bash
--- a/src/popper/runner_kubernetes.py
+++ b/src/popper/runner_kubernetes.py
@@ -263,16 +263,22 @@ class KubernetesRunner(StepRunner):
     def _vol_create(self, volume_name):
         """Create a default PersistentVolume of hostPath type.
         """
+        hostpathvol_path = "/tmp"
+        hostpathvol_size = "1Gi"
+        if self._config.resman_opts.get("hostpathvol_path", None):
+            hostpathvol_path = self._config.resman_opts.hostpathvol_path
+        if self._config.resman_opts.get("hostpathvol_size", None):
+            hostpathvol_size = self._config.resman_opts.hostpathvol_size
         vol_conf = {
             "kind": "PersistentVolume",
             "apiVersion": "v1",
-            "metadata": {"name": volume_name, "labels": {"type": "host"},},
+            "metadata": {"name": volume_name, "labels": {"type": "host"}},
             "spec": {
                 "persistentVolumeReclaimPolicy": "Recycle",
                 "storageClassName": "manual",
-                "capacity": {"storage": "1Gi",},
+                "capacity": {"storage": hostpathvol_size,},
                 "accessModes": ["ReadWriteMany"],
-                "hostPath": {"path": "/tmp"},
+                "hostPath": {"path": hostpathvol_path},
             },
         }
 


install.sh
```diff
@@ -21,6 +21,7 @@ install_system_wide() {
     fi
   fi
   echo
+
```
JayjeetAtGithub, line 24: Can we undo this line addition ?

docs/sections/cn_workflows.md
```diff
@@ -440,6 +440,10 @@ All the available configuration options have been described below:
   If this option is not provided, Popper will leave the task of scheduling the pods upon Kubernetes. 
   The exception to this is, when both the `pod_host_node` and `persistent_volume_name` options are not provided, Popper will try to find out a pod and schedule all the pods (init-pods + step-pods) on that node to use the `HostPath` persistent volume of 1GB which will be automatically created.
 
+* `hostpathvol_path`: A HostPath persistent volume is created at the path `hostpathvol_path`. If not provided, the /tmp path will be used.
```
JayjeetAtGithub, line 443: We can just write "The path to use for creating a HostPath volume. If not provided , `/tmp` will be used."

docs/sections/cn_workflows.md
```diff
@@ -440,6 +440,10 @@ All the available configuration options have been described below:
...
+
+* `hostpathvol_size`: A HostPath persistent volume is created of `hostpathvol_size`. If not provided, the default size of 1GB will be used.
```
JayjeetAtGithub, line 445: Same here. We can write "The size of the HostPath volume to create. ......."

JayjeetAtGithub (changes requested):
Thanks a lot for working on this. Really nice work. Can you also please squash the commits into a single one?

vivekr123:
Hi @JayjeetAtGithub. Thanks! I'm glad they were ok. I incorporated the changes you mentioned in a new pull request. I haven't squashed commits before, but I attempted to do so. Did it work?

vivekr123:
@JayjeetAtGithub I think it should all be cleaned up now with only 1 commit. Please let me know what you think :)

src/popper/runner_kubernetes.py
```diff
@@ -263,16 +263,22 @@ def _init_pod_delete(self):
     def _vol_create(self, volume_name):
         """Create a default PersistentVolume of hostPath type.
         """
+        hostpathvol_path = "/tmp"
+        hostpathvol_size = "1Gi"
+        if self._config.resman_opts.get("hostpathvol_path", None):
+            hospathvol_path = self._config.resman_opts.hostpathvol_path
```
JayjeetAtGithub, line 269: There is a typo in the variable name. It is `hospathvol_path`. Can you please fix this ?

src/popper/runner_kubernetes.py
```diff
@@ -263,16 +263,22 @@ def _init_pod_delete(self):
...
+        if self._config.resman_opts.get("hostpathvol_size", None):
+            hospathvol_size = self._config.resman_opts.hostpathvol_size
```
JayjeetAtGithub, line 271: Same here.

src/popper/runner_kubernetes.py
```diff
@@ -263,16 +263,22 @@ def _init_pod_delete(self):
...
         vol_conf = {
             "kind": "PersistentVolume",
             "apiVersion": "v1",
             "metadata": {"name": volume_name, "labels": {"type": "host"},},
             "spec": {
                 "persistentVolumeReclaimPolicy": "Recycle",
                 "storageClassName": "manual",
-                "capacity": {"storage": "1Gi",},
+                "capacity": {"storage": hostpathvol_size,},
```
JayjeetAtGithub, line 279: We can remove the extra `,`

docs/sections/cn_workflows.md
```diff
@@ -440,6 +440,10 @@ All the available configuration options have been described below:
   If this option is not provided, Popper will leave the task of scheduling the pods upon Kubernetes. 
   The exception to this is, when both the `pod_host_node` and `persistent_volume_name` options are not provided, Popper will try to find out a pod and schedule all the pods (init-pods + step-pods) on that node to use the `HostPath` persistent volume of 1GB which will be automatically created.
 
+* `hostpathvol_path`: The path to use for creating a HostPath volume. If not provided, /tmp will be used.
+
+* `hostpathvol_size`: The size to use for creating a HostPath volume. If not provided, 1GB will be used.
```
JayjeetAtGithub, line 445: We can just say `The size of the HostPath volume. ...`

JayjeetAtGithub (changes requested):
Thanks @vivekr123 . Just a few comments. I think then it will be good to merge.

src/popper/runner_kubernetes.py
JayjeetAtGithub: To catch typos like this, it will be a good idea to test it locally on a minikube cluster maybe. If its too much work, its fine, I can do it on my end.

vivekr123: Thanks! Good to know. I'll look into that.

vivekr123:
The suggested edits should be in the updated files

JayjeetAtGithub:
Thanks a lot @vivekr123  🎉 . Merged it.

PR #940, opened 2020-11-29
Fix #927: Prevent trying to schedule on master nodes first

### LINKED ISSUE #927: cli: k8s: try scheduling the init pod on the master node at the very last
When no shared PersistentVolume is available, we try to schedule the init pod iteratively on all the nodes, and once a node accepts the init pod, we remember it and schedule all the subsequent steps pods there on that node so that all the pods can use a local volume on that node. Currently,  [while visiting each node iteratively](https://github.com/getpopper/popper/blob/01eadd5e54bd686291cef20bcb4e2fd3b493829f/src/popper/runner_kubernetes.py#L117), we start from the first node and most of the time it turns out to be the master node (labeled as master) and results in waste work trying to schedule on the master node. To prevent this wastage of time trying to schedule on master nodes, we can tweak the logic to find out the master nodes from the node list and push them to the end of the list, so that we start with visiting a non-master node and keep the masters as the last resort.

CarterB-UT:
Based off of what the description of the issue said and the generous aid from slack this is my current idea for the solution to this problem. I

Seeing as the base premise involved re-ordering the list of nodes I assumed the order was of little consequence aside from having masters last so the non-master nodes will be ordered differently than before.  Furthermore, I considered having two arrays that just got connected after iterating through the nodes but I was unsure if you guys would want to just keep it one array the whole time for cleanliness.

If I have done anything incorrectly or there are issues of cleanliness or efficiency just let me know.

### DIFF
--- a/src/popper/runner_kubernetes.py
+++ b/src/popper/runner_kubernetes.py
@@ -110,10 +110,19 @@ class KubernetesRunner(StepRunner):
             pod_host_node = self._config.resman_opts.pod_host_node
 
         elif not self._config.resman_opts.get("persistent_volume_name", None):
-            nodes = [
-                node.metadata.labels["kubernetes.io/hostname"]
-                for node in self._kclient.list_node().items
-            ]
+            nodes = []
+
+            for node in self._kclient.list_node().items:
+                node_role = ""
+                if node.spec.taints and len(node.spec.taints) > 0:
+                    node_role = node.spec.taints[0].key
+                if node_role != "node-role.kubernetes.io/master" and node_role != "node-role.kubernetes.io/unreachable":
+                    nodes.insert(0, node.metadata.labels["kubernetes.io/hostname"])
+                else:
+                    nodes.insert(
+                        len(nodes), node.metadata.labels["kubernetes.io/hostname"]
+                    )
+
             for node in nodes:
                 log.debug(f"trying to schedule init pod on {node}")
                 e = self._init_pod_create(node)
--- a/src/popper/runner_kubernetes.py
+++ b/src/popper/runner_kubernetes.py
@@ -116,7 +116,10 @@ class KubernetesRunner(StepRunner):
                 node_role = ""
                 if node.spec.taints and len(node.spec.taints) > 0:
                     node_role = node.spec.taints[0].key
-                if node_role != "node-role.kubernetes.io/master" and node_role != "node-role.kubernetes.io/unreachable":
+                if (
+                    node_role != "node-role.kubernetes.io/master"
+                    and node_role != "node-role.kubernetes.io/unreachable"
+                ):
                     nodes.insert(0, node.metadata.labels["kubernetes.io/hostname"])
                 else:
                     nodes.insert(


JayjeetAtGithub:
Thanks a lot for working on this @CarterB-UT . I will look into it and let you know.

src/popper/runner_kubernetes.py
```diff
@@ -110,10 +110,19 @@ def _init_pod_schedule(self):
...
-            nodes = [
-                node.metadata.labels["kubernetes.io/hostname"]
-                for node in self._kclient.list_node().items
-            ]
+            nodes = []
+
+            for node in self._kclient.list_node().items:
+                node_role = ""
+                if node.spec.taints is not None:
+                    node_role = node.specs.taints.key
```
JayjeetAtGithub, line 118: It will be `node.spec` not `node.specs`.

JayjeetAtGithub: Also, node.spec.taints is a list. We need to just check the first element of the taints list. the first element has the information about if the node whether it is a master, or is unreachable.

src/popper/runner_kubernetes.py
```diff
@@ -110,10 +110,19 @@ def _init_pod_schedule(self):
...
+                if node_role != "node-role.kubernetes.io/master":
```
JayjeetAtGithub, line 119:
```
if node_role != "node-role.kubernetes.io/master" and node_role != "node.kubernetes.io/unreachable":
```


We can also check for the `node.kubernetes.io/unreachable` label to check for unreachable nodes.

JayjeetAtGithub (changes requested):
Thanks a lot @CarterB-UT . The changes look good. Just a few fixes.

src/popper/runner_kubernetes.py
CarterB-UT: Should unreachable nodes be added in the else case below this code or should they be excluded there as well?

src/popper/runner_kubernetes.py
CarterB-UT: Ah. Sorry for missing those typos. 

As for taints, are you saying `node.spec.taints[0]` would be better as opposed to `node.spec.taints.key?` And is the `is not None` line needed? I was uncertain if you meant taints could be null or simply empty on slack. 

Thank you very much for the continued help and time.

JayjeetAtGithub: spec may or may not have the taints key. if the taints key , the value it will have is a list. Now this taints list can either be empty or non-empty. if non-empty, pick the first element and check.

src/popper/runner_kubernetes.py
JayjeetAtGithub: the unreachable label can be excluded in line 119 as well.

JayjeetAtGithub:
Thanks for the updates. I will get back soon.

src/popper/runner_kubernetes.py
```diff
@@ -110,10 +110,19 @@ def _init_pod_schedule(self):
...
-                node.metadata.labels["kubernetes.io/hostname"]
-                for node in self._kclient.list_node().items
-            ]
+            nodes = []
+
+            for node in self._kclient.list_node().items:
+                node_role = ""
+                if node.spec.taints is not None:
+                    node_role = node.spec.taints[0]
+                if node_role != "node-role.kubernetes.io/master" && node_role != "node-role.kubernetes.io/unreachable":
```
JayjeetAtGithub, line 119: `and` is the logical operator for doing and in python.

JayjeetAtGithub (changes requested):
Thanks a lot. Left some comments. Also can you please separate the changes into a self-contained single commit?

CarterB-UT:
Okay, we have tried squashing with some more changes based upon what you said. Let us know if there are any lingering issues @JayjeetAtGithub. Thanks again for the continued help.

src/popper/runner_kubernetes.py
```diff
@@ -110,10 +110,19 @@ def _init_pod_schedule(self):
...
-            ]
+            nodes = []
+
+            for node in self._kclient.list_node().items:
+                node_role = ""
+                if node.spec.taints.key is not None and len(node.spec.taints) > 0:
+                    node_role = node.spec.taints[0]
+                if node_role != "node-role.kubernetes.io/master" and node_role != "node-role.kubernetes.io/unreachable":
+                    nodes.insert(0, node.metadata.labels["kubernetes.io/hostname"])
+                else if node_role != "node-role.kubernetes.io/unreachable":
```
JayjeetAtGithub, line 121: this can be just a simple `else`, we do not need to check the role again here. we are already checking for it in the if part. Also `else if` is expressed as `elif` in python language.

JayjeetAtGithub (changes requested):

src/popper/runner_kubernetes.py
JayjeetAtGithub: You can just add a commit. no need to squash.

src/popper/runner_kubernetes.py
```diff
@@ -110,10 +110,19 @@ def _init_pod_schedule(self):
...
         elif not self._config.resman_opts.get("persistent_volume_name", None):
-            nodes = [
-                node.metadata.labels["kubernetes.io/hostname"]
-                for node in self._kclient.list_node().items
-            ]
+            nodes = []
+
+            for node in self._kclient.list_node().items:
+                node_role = ""
+                if node.spec.taints.key is not None and len(node.spec.taints) > 0:
```
JayjeetAtGithub, line 117: `taints` is a list. We can avoid the first check.

src/popper/runner_kubernetes.py
```diff
@@ -110,10 +110,19 @@ def _init_pod_schedule(self):
...
+                    node_role = node.spec.taints[0]
```
JayjeetAtGithub, line 118: It should be `node_role = node.spec.taints[0].key`

src/popper/runner_kubernetes.py
CarterB-UT: Ah. I must have misinterpreted you in the past. I thought we were attempting to totally exclude unreachable nodes.

CarterB-UT:
Accidentally unsquashed the process. Will resquash.

JayjeetAtGithub (approved):
Thanks a lot @CarterB-UT .

PR #941, opened 2020-11-29
Fix #821: Added life_of_a_workflow.md to describe the life of a workflow

### LINKED ISSUE #821: doc: write "life of a workflow execution" section
add a section that explains and shows images of what popper does when it executes a workflow:
 1. look at `uses` attribute and pull/build image
 2. configure and create container
 3. launch container, wait for it to be done
 4. move on to next step

rishiUT:
Added life_of_a_workflow.md to describe the life of a workflow

fixes #821


docs/sections/life_of_a_workflow.md
```diff
@@ -0,0 +1,111 @@
+# Life of a Workflow
+
+This section explains what popper does when it executes a workflow. We will break down what popper does behind the scenes when executing the sample workflow described [here](https://github.com/getpopper/popper/blob/master/docs/sections/getting_started.md).
+
+Each step of a workflow has the following stages:
+
+## 1. Look at `uses` attribute and pull/build image
+
+Each step of a workflow must specify the `DockerFile` images it requires with a `uses` line. For example, the first step of our example workflow contains the following line:
```
ivotron, line 9: We cannot reference an image as "a `Dockerfile` image", rather an image is built from a `Dockerfile`, and a container is instantiated from an image. The `uses` keyword references either a `Dockerfile` or an image. If it is referencing a `Dockerfile`, Popper builds a container image first, then it instantiates a container from it.

docs/sections/life_of_a_workflow.md
```diff
@@ -0,0 +1,111 @@
...
+```
+uses: docker://byrnedo/alpine-curl:0.1.8
+```
+These statements may refer to a `Dockerfile` inside the same repository as the workflow, or to a `Dockerfile` inside an external, public repository or container registry.
```
ivotron, line 13: "... or an image in a registry"

docs/sections/life_of_a_workflow.md
```diff
@@ -0,0 +1,111 @@
...
+
+The example `uses` line above would result in the following output from Popper:
+```
+[download] docker pull byrnedo/alpine-curl:0.1.8
+```
+This line indicates that the necessary image was successfully pulled and built by docker. The workings and limitations of `uses` and other possible attributes for a workflow are outlined [here](cn_workflows.md).
```
ivotron, line 19: in this case the image is just pulled, not built

docs/sections/life_of_a_workflow.md
```diff
@@ -0,0 +1,111 @@
...
+This section explains what popper does when it executes a workflow. We will break down what popper does behind the scenes when executing the sample workflow described [here](https://github.com/getpopper/popper/blob/master/docs/sections/getting_started.md).
+
+Each step of a workflow has the following stages:
+
+## 1. Look at `uses` attribute and pull/build image
+
+Each step of a workflow must specify the `DockerFile` images it requires with a `uses` line. For example, the first step of our example workflow contains the following line:
+```
+uses: docker://byrnedo/alpine-curl:0.1.8
+```
```
ivotron, line 12: in order to make it self-contained, we can add the entire workflow. And to avoid repeating it, we could add the entire workflow at the beginning and then reference it throughout all subsections.

docs/sections/life_of_a_workflow.md
```diff
@@ -0,0 +1,111 @@
...
+These statements may refer to a `Dockerfile` inside the same repository as the workflow, or to a `Dockerfile` inside an external, public repository or container registry.
+
+The example `uses` line above would result in the following output from Popper:
+```
+[download] docker pull byrnedo/alpine-curl:0.1.8
+```
```
ivotron, line 18: it would be useful if this can show what command would need to be executed if users were doing this manually on a terminal. So for each subsection we would show the corresponding command(s).
We could show how this would be done in Docker and Singularity, as a way of showing how popper takes care of handling the differences in CLI syntax for each distinct tool.

ivotron (changes requested):
thanks a lot @rishiUT, this looks great! Would it be possible to put this as the last subsection of the [Workflow Syntax and Execution Runtime](https://github.com/getpopper/popper/blob/master/docs/sections/cn_workflows.md) section?

docs/sections/life_of_a_workflow.md
rishiUT: We can add the workflow to the beginning of the section! However, it would be easiest to reference it the same way we have been, by quoting the relevant sections. Is this alright?

docs/sections/life_of_a_workflow.md
rishiUT: Ah, thank you for the explanation! We have reworded that line to be more accurate.

docs/sections/life_of_a_workflow.md
ivotron: yes! that would make it even clearer what the text is referencing to

docs/sections/cn_workflows.md
```diff
@@ -528,3 +528,133 @@ This runs the step `one` locally in the host and step `two` through SLURM on any
...
+#### Host
+
+Popper executes the workflows by default using the `host` machine as the resource manager. So, when no resource manager is provided like the example below, the workflow runs on the local machine.
+
+```bash
+popper run -f sample.yml
+```
+
+The above assumes `docker` as the container engine and `host` as the resource manager to be
+used.
```
ivotron, line 541: all this section was removed in a recent commit, got back in this PR. Could you please remove it?

docs/sections/cn_workflows.md
```diff
@@ -528,3 +528,133 @@ This runs the step `one` locally in the host and step `two` through SLURM on any
...
+
+```
+id: download
+```
+```
+args: [-LO, https://github.com/datasets/co2-fossil-global/raw/master/global.csv]
+```
+Using these inputs, Popper executes the following command for a Docker build:
+```
+docker create name=popper_download_f20ab8c9 image=byrnedo/alpine-curl:0.1.8 command=['-LO', 'https://github.com/datasets/co2-fossil-global/raw/master/global.csv']
```
ivotron, line 602: the equivalent command is a bit different:

```bash
docker create --name=popper_download_f20ab8c9  byrnedo/alpine-curl:0.1.8 -LO https://github.com/datasets/co2-fossil-global/raw/master/global.csv
```

docs/sections/life_of_a_workflow.md
```diff
@@ -0,0 +1,113 @@
+# Life of a Workflow
```
ivotron, line 1: this file can be removed

ivotron (changes requested):
thanks @rishiUT, this is almost done. Please take a look at the few remaining comments!

docs/sections/cn_workflows.md
rishiUT: Sorry about that! It's fixed now.

ivotron:
thanks a lot @rishiUT !

PR #942, opened 2020-11-30
Added test to check that workflow using buildkit

VD44:


ivotron:
thanks @VD44 , we reached the limit of travis free credits. So we're transitioning to another CI service, once we do that, then we can run tests for this PR and merge

ivotron:
Hi @VD44, can you please rebase master? We have now moved to github actions and that should trigger builds for this

PR #945, opened 2020-12-02
Added output for docker pull

### LINKED ISSUE #892: cli: show output of docker pull
similar to #785 but for `docker pull`

michaelroyzen:
Resolves #892. @ivotron your feedback would be appreciated.

### DIFF
--- a/src/popper/runner_host.py
+++ b/src/popper/runner_host.py
@@ -193,7 +193,6 @@ class DockerRunner(StepRunner):
         elif not self._config.skip_pull and not step.skip_pull:
             log.info(f"[{step.id}] docker pull {img}:{tag}")
             if not self._config.dry_run:
-                #self._d.images.pull(repository=f"{img}:{tag}")
                 streamer = self._d.api.pull(
                     repository=f"{img}:{tag}", decode=True,
                 )
@@ -202,7 +201,6 @@ class DockerRunner(StepRunner):
                     if self._config.quiet:
                         continue
                     chunk = chunk.strip()
-                    #print("chunk", chunk)
                     if chunk:
                         import json
                         chunk = json.loads(chunk)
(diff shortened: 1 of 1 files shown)


src/popper/runner_host.py
```diff
@@ -193,7 +193,23 @@ def _create_container(self, cid, step):
...
+                
+                for chunk in streamer.splitlines():
+                    if self._config.quiet:
+                        continue
+                    chunk = chunk.strip()
+                    if chunk:
+                        import json
+                        chunk = json.loads(chunk)
+                        if 'id' in chunk:
+                            log.step_info(f"[{step.id}] " + chunk['id'] + ': ' + chunk['status'])
```
ivotron, line 208: to be consistent with the way logging is done in other places, this can ommit the `[{step.id}] ` prefix and log the output directly

src/popper/runner_host.py
```diff
@@ -193,7 +193,23 @@ def _create_container(self, cid, step):
...
+                        else:
+                            log.step_info(f"[{step.id}] " + chunk['status'])
+                log.step_info(f"[{step.id}] {img}:{tag}")
```
ivotron, line 211: this line can be omitted, as the last line from docker pull is something like `Downloaded newer image for <url>:<tag>` so it is already informing about what got downloaded

src/popper/runner_host.py
```diff
@@ -193,7 +193,23 @@ def _create_container(self, cid, step):
...
+                    if self._config.quiet:
+                        continue
+                    chunk = chunk.strip()
+                    if chunk:
+                        import json
+                        chunk = json.loads(chunk)
+                        if 'id' in chunk:
+                            log.step_info(f"[{step.id}] " + chunk['id'] + ': ' + chunk['status'])
+                        else:
+                            log.step_info(f"[{step.id}] " + chunk['status'])
```
ivotron, line 210: similarly, the `[{step.id}] ` prefix can be omitted

ivotron (changes requested):
thanks a lot @michaelroyzen , this looks great. Please take a look at the feedback

michaelroyzen:
Thanks for the feedback. I've made the changes, please let me know if there is anything else I should do.

ivotron:
thanks a lot @michaelroyzen! 🙏

PR #946, opened 2020-12-05
Add colorized output to log.info and log.step_info for bracket prefixes

### LINKED ISSUE #541: cli: colorize output
Add colors to output to improve what's being shown on the terminal. For this, we can color the square-bracketed prefix that we print currently. For example, this:

```
[my action] docker pull foo
[my action] docker create foo...
[my action] docker start.....
```

would result in having `[my action]` in color. Same for the `--dry-run` prefixes being worked as part of #509 .

zacswolf:
Fixes #541. @ivotron your feedback would be appreciated

### DIFF
--- a/src/popper/log.py
+++ b/src/popper/log.py
@@ -36,8 +36,8 @@ class PopperFormatter(logging.Formatter):
 
     log_format = {
         "DEBUG": f"{BOLD_CYAN}%(levelname)s: %(msg)s {RESET}",
-        "STEP_INFO": "%(msg)s",
-        "INFO": "%(msg)s",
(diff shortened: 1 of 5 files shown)


ivotron (commented):
thanks @zacswolf, this looks great! Would it be possible to add coloring for the `msg_prefix` string as well? That'd be really cool! To test you can run with `popper run --dry-run`

ivotron (commented):
All looks good, only the error being thrown by black in CI

zacswolf:
@ivotron How do I fix this formatting issue? I ran the black linter locally on python3.8 and was able to successfully format the file that is causing the CI build issues (runner_slurm.py).

Issue #812, opened 2020-04-22
cli: check for the presence of required executables


JayjeetAtGithub:
The [`assert_executable_exists(cmd)`](https://github.com/systemslab/popper/blob/master/cli/popper/utils.py#L55) function checks for the presence of an executable in the environment where the workflow is executed.

So, we can utilize this function to check for the presence of the required binaries/executables through the `__init__()` method(constructor) of the concerned runner classes, like [`runner_slurm.SlurmRunner`](https://github.com/systemslab/popper/blob/master/cli/popper/runner_slurm.py#L12), etc defined in [`runner_host.py`](https://github.com/systemslab/popper/blob/master/cli/popper/runner_host.py) and [`runner_slurm.py`](https://github.com/systemslab/popper/blob/master/cli/popper/runner_slurm.py). This would block a runner from getting initialized and fail if it does not meet its required software dependencies.

For example, we need to check for `sbatch` executable in `runner_slurm.SlurmRunner`, `singularity` executable in `runner_host.SingularityRunner`,  `podman` in `runner_host.PodmanRunner`, etc.

abbycriswell:
Hello again! My teammate and I also commented on issue #913. We would also like to work on this issue.

JayjeetAtGithub:
Hi @abbycriswell. Thanks a lot for looking into this issue. Please go ahead. Feel free to ask any questions you might have. You can start by looking into the [contribution guidelines](https://github.com/getpopper/popper/blob/master/CONTRIBUTING.md) and the [development environment setup](https://github.com/getpopper/popper/blob/master/docs/installation.md#development-setup) guides. Also, feel free to join our slack channel ([join here](https://join.slack.com/t/getpopper/shared_invite/zt-dtn0se2s-c50myMHNpeoikQXDeNbPew)) for any discussion regarding the issue.

PR #950, opened 2020-12-08
Executable checks for SlurmRunner, DockerRunner, SingularityRunner etc as in #812

maruthgoyal:
The build appears to fail on Travis, as seen here:
[Travis job](https://travis-ci.com/github/maruthgoyal/popper/jobs/451104742)

In particular, the `sbatch` executable appears to be missing.

### DIFF
--- a/src/popper/runner_host.py
+++ b/src/popper/runner_host.py
@@ -115,6 +115,10 @@ class DockerRunner(StepRunner):
         if not init_docker_client:
             return
 
+        docker_executables = ["docker"]
+        for exe in singularity_executables:
+            assert_executable_exists(exe)
+
         try:
             self._d = docker.from_env()
             self._d.version()
@@ -244,6 +248,10 @@ class PodmanRunner(StepRunner):
 
         if not init_podman_client:
             return
+
+        podman_executables = ["podman"]
+        for exe in podman_executables:
+            assert_executable_exists(exe)
 
         try:
             _, _, self._p_info = HostRunner._exec_cmd(["podman", "info"], logging=False)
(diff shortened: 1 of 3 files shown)


src/popper/runner_host.py
```diff
@@ -115,6 +116,10 @@ def __init__(self, init_docker_client=True, **kw):
         if not init_docker_client:
             return
 
+        docker_executables = ["docker"]
```
ivotron, line 119: we can skip checking for the `docker` binary. This is because in the case of Docker, the code doesn't strictly depend on the `docker` client binary. The docker-py library can directly talk to the daemon, which is implicitly done in line 125 (when `self._d.version()` is executed)

src/popper/runner_slurm.py
```diff
@@ -8,13 +8,18 @@
...
 from popper.runner_host import SingularityRunner as HostSingularityRunner
+from popper.utils import assert_executable_exists
 
 
 class SlurmRunner(HostRunner):
     def __init__(self, **kw):
         super(SlurmRunner, self).__init__(**kw)
         self._spawned_jobs = set()
 
+        slurm_executables = ["sbatch", "srun", "scancel", "mpirun"]
```
ivotron, line 19: for CI tests, these won't be available, and thus this fail as you correctly pointed out (on Slack). So one thing that can be done is to mock the `assert_executable_exists()` function in the test (`test_runner_slurm.py`).

ivotron (changes requested):
thanks a lot for working on this @maruthgoyal! Please take a look at the comments and let us know if you have any questions

src/popper/runner_host.py
maruthgoyal: Sounds good, makes sense! I'll make the change

src/popper/runner_slurm.py
maruthgoyal: Gotcha! Thank you, I'm working on the mocking now.

maruthgoyal:
Ok got the mocking working now!

ivotron (approved):

ivotron:
thanks a lot @maruthgoyal ! 🙏

PR #951, opened 2020-12-10
Fix --quiet

### LINKED ISSUE #815: cli: singularity build should not print logs to terminal

ishanashah:
Fixes #815. @ivotron your feedback would be appreciated

### DIFF
--- a/src/popper/runner_host.py
+++ b/src/popper/runner_host.py
@@ -147,7 +147,7 @@ class DockerRunner(StepRunner):
         if not container and not self._config.reuse:
             container = self._create_container(cid, step)
 
-        log.step_info(f"[{step.id}] docker start")
+        log.info(f"[{step.id}] docker start")
 
         if self._config.dry_run:
             return 0
@@ -178,7 +178,7 @@ class DockerRunner(StepRunner):
         build, _, img, tag, build_ctx_path = self._get_build_info(step)
 
         if build:
-            log.step_info(f"[{step.id}] docker build {img}:{tag} {build_ctx_path}")
+            log.info(f"[{step.id}] docker build {img}:{tag} {build_ctx_path}")
             if not self._config.dry_run:
                 streamer = self._d.api.build(
                     decode=True, path=build_ctx_path, tag=f"{img}:{tag}", rm=True,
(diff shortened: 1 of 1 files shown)


ivotron (changes requested):
thanks @ishanashah! The logging level for the lines you modified was already OK. The main problem for the bug you're trying to address is in [this line](https://github.com/getpopper/popper/blob/master/src/popper/runner_host.py#L488) . What needs to be done for this issue is to find a way for that line to _not_ produce output when `--quiet` is given.

ishanashah:
I set the "quiet" parameter in self._s.build to pass information to the build method about whether low priority logs should commit. The Singularity source code allows this parameter to be set, but it seems to not be working on the Singularity side. This fix should solve the issue on the side of Popper, but another fix in the Singularity source code is required.

src/popper/runner_host.py
```diff
@@ -147,7 +147,7 @@ def run(self, step):
         if not container and not self._config.reuse:
             container = self._create_container(cid, step)
 
-        log.info(f"[{step.id}] docker start")
+        log.step_info(f"[{step.id}] docker start")
```
ivotron, line 150: these were OK as `log.info()`, this are supposed to still be shown when `--quiet` is given

src/popper/runner_host.py
```diff
@@ -485,7 +485,13 @@ def _build_from_recipe(self, build_ctx_path, build_dest, cid):
         pwd = os.getcwd()
         os.chdir(build_ctx_path)
         recipefile = SingularityRunner._get_recipe_file(build_ctx_path, cid)
-        self._s.build(recipe=recipefile, image=cid, build_folder=build_dest, force=True)
+        self._s.build(
+            recipe=recipefile,
+            image=cid,
+            build_folder=build_dest,
+            force=True,
+            quiet=self._s.quiet,
```
ivotron, line 493: cool! it's nice that there's already a `quiet` parameter on the singularity python lib side. The only thing to note here is that the value for this parameter should be the one from `self._config.quiet`, which is the one that holds the value provided from the CLI

ivotron (changes requested):

ishanashah:
Thanks for the advice! I fixed the logging level changes, sorry about that. I also changed the parameter value to be associated with the command line argument.

ivotron:
thanks @ishanashah!

PR #954, opened 2020-12-13
 Issue #541 - Added coloring to message string and bracket prefixes

SamarthGoyal007:


ivotron:
This is already being worked on #946, thanks a lot @SamarthGoyal007!

Issue #862, opened 2020-06-19
example: add a CI pipeline for gradle


ivotron:
Add an example workflow in the `examples/ci/` folder that exemplifies how to make use of Popper to create a CI pipeline for java projects that use [`gradle`](https://guides.gradle.org/building-java-libraries/) as their build system.

MJVNOR:
i will take this

PR #963, opened 2021-01-05
example: add a CI pipeline for gradle (#862)

AdrianSot:
Added a java example that uses gradle as the build system.


ivotron:
thanks!

PR #976, opened 2021-07-07
Add `translate` command

shumbo:
## Overview

This PR adds a new `translate` command, which translates between specifications of different workflow runners, to the Popper CLI. At this moment the command only supports the very basic popper workflow to a drone workflow.

## Command

The `translate` command takes four options as follows:

```
> popper translate --help
Usage: popper translate [OPTIONS]

Options:
  --from [popper]  Format of the input file.
  -f, --file TEXT  File containing the definition of the workflow.
  --to [drone]     Format of the output file.
  --out TEXT       Name of the output file.
  --help           Show this message and exit.
```

At this moment, it only supports `popper` for `--from` and `drone` for `--to`, but we intend to support the opposite and possibly more.

The command uses the default filename for an input/output format. For example, it uses `.popper.yml` for popper workflows and `.drone.yml` for drone workflows.

## Translator

To implement the command, this PR adds a new `WorkflowTranslator` base class. The role of a translator instance is to take a popper workflow and translate it to another workflow specification.

There already is `Exporter` that provides similar functionality. However, the output of `Exporter` runs Popper on different workflow engines, while `Translator` translates workflows and runs the same set of operations without using Popper.

`WorkflowTranslator` provides a `get_translator` static method that allows the caller to get an instance of translator for given format.

## Drone Translator

`DroneTranslator` implements `WorkflowTranslator` and translates a Popper workflow into Drone workflow. At this moment, it only supports the following properties but is capable of running the most basic workflows, such as [getting a transpose of CSV from the popper document](https://popper.readthedocs.io/en/latest/sections/getting_started.html#create-your-first-workflow).


src/popper/commands/cmd_translate.py
```diff
@@ -0,0 +1,51 @@
...
+)
+@click.option(
+    "outfmt",
+    "--to",
+    help="Format of the output file.",
+    type=click.Choice(["drone"]),
+    default="drone",
+)
+@click.option(
+    "outfile",
```
JayjeetAtGithub, line 36: Maybe it is a good idea to keep the names of the arguments and the flags the same for the sake of consistency in the codebase.

src/test/test_translator_drone.py
```diff
@@ -0,0 +1,24 @@
+from click.testing import CliRunner
+
+from popper import __version__
+from popper import _version_file
+from popper.commands import cmd_version
+from popper.translators.translator_drone import DroneTranslator
+from .test_common import PopperTest
+
+
+class TestDroneTranslator(PopperTest):
```
JayjeetAtGithub, line 10: Can we also add tests for `translate_step` and `translate` ?

JayjeetAtGithub (changes requested):
Thanks a lot for working on this @shumbo. The changes look great. Can you please also fix the lining errors? because without that the other workflows don't run.

src/popper/commands/cmd_translate.py
shumbo: @JayjeetAtGithub 

I renamed all options because `from` is a reserved word in Python. I personally think `translate --from --to` is natural, but do you think it's better to

- rename the `from` option to something else and use the same name for the argument and flag

or

- only rename the argument for `from`

src/popper/commands/cmd_translate.py
```diff
@@ -0,0 +1,48 @@
...
+    "outfmt",
+    "--to",
+    help="Format of the output file.",
+    type=click.Choice(["drone"]),
+    default="drone",
+)
+@click.option(
+    "outfile", "--out", help="Name of the output file.", required=False,
+)
+@pass_context
```
JayjeetAtGithub, line 38: Remove this `--out` flag for simplicity and impose filename to `drone.yml` for drone and `taskfile.yml` for taskfile

src/popper/commands/cmd_translate.py
```diff
@@ -0,0 +1,48 @@
...
+@click.option(
+    "infile",
+    "--file",
+    "-f",
+    help="File containing the definition of the workflow.",
+    required=False,
+    default=".popper.yml",
+)
+@click.option(
+    "outfmt",
```
JayjeetAtGithub, line 29: maybe `to_fmt` and `from_fmt`

src/popper/commands/cmd_translate.py
```diff
@@ -0,0 +1,48 @@
...
+)
+@click.option(
+    "infmt",
+    "--from",
+    help="Format of the input file.",
+    type=click.Choice(["popper"]),
+    default="popper",
+)
+@click.option(
+    "infile",
```
JayjeetAtGithub, line 21: since we are omitting `outfile`, only file should work

JayjeetAtGithub (changes requested):

src/popper/commands/cmd_translate.py
shumbo: resolved 6c5917fdb882bbf684faa871a38a3fea6d06a9fb

src/popper/commands/cmd_translate.py
shumbo: resolved 5c358a324021e98425e9bd11b35219fdb66d558f

src/popper/commands/cmd_translate.py
shumbo: resolved 5c358a324021e98425e9bd11b35219fdb66d558f

src/test/test_translator_drone.py
shumbo: resolved 

https://github.com/getpopper/popper/pull/976/files#diff-59acabdf0d01eb06086a11bb0f995aab6e2939f239067b1dc9c8358b428ca6cfR12

https://github.com/getpopper/popper/pull/976/files#diff-59acabdf0d01eb06086a11bb0f995aab6e2939f239067b1dc9c8358b428ca6cfR50

JayjeetAtGithub: Thanks

JayjeetAtGithub (approved):
Very nice work! Thanks a lot @shumbo .

PR #993, opened 2021-09-24
Document Popper-to-Task Translator

shumbo:
This PR adds documentation for Popper-to-Task translator. 

It also makes changes to Popper-to-Task translator to raise an error if the source popper workflow contains a step whose id is `default`, which conflicts with the Task's reserved name.


JayjeetAtGithub:
LGTM 🎉

PR #994, opened 2021-09-29
Popper-to-Task Translation

shumbo:
This PR releases the Popper-to-Task translator.

See the updated docs for its usage.


JayjeetAtGithub:
Awesome work @shumbo  🎉

